PR #81, opened 2020-09-09
additional fixes for status update

monotosh-avi:
- don't sync status for dummy vs
- add service metadata in pool for namespace shard
- test cases for duplicate backend in route
- Update status from AviVsCacheAdd only for Post / Delete operation: If we update for put operation, then we may be adding updates which are not required. These should be handled during pool add operations


sudswasavi (approved):

sudswasavi:
build ako

monotosh-avi:
build ako

sudswasavi (approved):

PR #82, opened 2020-09-09
Misc fixes

sudswasavi:
This PR, modifies the helm charts to become more readable.
Plus makes the controller version auto-discoverable.

It also implements a slow retry layer in AKO.
The slow retry layer is used only if there's a problem in contacting the Avi Controller. 
We now call AviRestOperate inside a go routine and timeout post 80 seconds.

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -14,11 +14,13 @@
 package rest
 
 import (
+	"errors"
 	"os"
 	"regexp"
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	avicache "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/cache"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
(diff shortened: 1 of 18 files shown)


tests/bootuptests/stale_obj_delete_test.go
```diff
@@ -116,9 +116,12 @@ func injectMWForObjDeletion() {
 		} else if r.Method == "GET" {
 			integrationtest.FeedMockCollectionData(w, r, mockFilePath)
 
-		} else if strings.Contains(url, "login") {
+		} else if strings.Contains(url, "initial-data") {
 			w.WriteHeader(http.StatusOK)
-			w.Write([]byte(`{"success": "true"}`))
+			w.Write([]byte(`{"version": {"Version": "20.1.2"}}`))
+		} else if strings.Contains(url, "initial-data") {
```
monotosh-avi, line 122: repeated checks ?

sudswasavi: Great catch. Done.

internal/rest/dequeue_nodes.go
```diff
@@ -399,21 +399,23 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
 									utils.AviLog.Infof("key: %s, msg: Error is not of type AviError, err: %v, %T", key, rest_ops[i].Err, rest_ops[i].Err)
 									continue
 								}
-								retryable, fastRetryable := rest.RefreshCacheForRetryLayer(publishKey, aviObjKey, rest_ops[i], aviError, aviclient, avimodel, key)
-								fastRetry = fastRetry || fastRetryable
-								retry = retry || retryable
+								regularRetry, slowRetry = rest.RefreshCacheForRetryLayer(publishKey, aviObjKey, rest_ops[i], aviError, aviclient, avimodel, key)
```
monotosh-avi, line 402: RefreshCacheForRetryLayer would return true always for regular retry, so we would never go to slow retry.

sudswasavi: Retained the logic that was there earlier and handling the slow retry checks separately.

parimanur (approved):

monotosh-avi (approved):

PR #86, opened 2020-09-10
Stop the slow retry queue on sigterm

sudswasavi:
We were missing stopping the slow retry queue on sigterm.
This commit takes care of that.


shubhamavi:
build ako

shubhamavi (approved):

PR #88, opened 2020-09-10
gateway status conditions init, gw-gwclass UTs

shubhamavi:
This PR introduces gateway status condition implementation for both gatewayConditions and listenerConditions.
Cache mapping updates to accommodate identifying portconflicts among multiple services
Added UTs for gateway <-> gatewayclass relationships

adding static IP support via gateway, since PUT calls on vsvip for changing preferred IP is blocked by Avi controller 20.1.2, extending the same behaviour for gateways. Gateways must be recreated with changed preferred IPs in order to change IPs.

also included is fix for VS protocol

Gateway conditions will be more fine tuned via UTs subsequently


internal/k8s/ako_init.go
```diff
@@ -420,6 +421,7 @@ func (c *AviController) FullSyncK8s() {
 			utils.AviLog.Errorf("Unable to retrieve the gateways during full sync: %s", err)
 		} else {
 			for _, gatewayObj := range gatewayObjs {
+				status.InitializeGatewayConditions(gatewayObj)
```
sudswasavi, line 424: Will this put a working `synced` gateway into initialising state?

internal/status/advl4_status.go
```diff
@@ -160,9 +200,40 @@ func updateGatewayStatusObject(gw *advl4v1alpha1pre1.Gateway, updateStatus *advl
...
 		}
-		return updateGatewayStatusObject(updatedGW, updateStatus, retry+1)
+		return UpdateGatewayStatusObject(updatedGW, updateStatus, retry+1)
 	}
 
 	utils.AviLog.Infof("msg: Successfully updated the gateway %s/%s status %+v", gw.Namespace, gw.Name, utils.Stringify(updateStatus))
 	return nil
 }
+
+func InitializeGatewayConditions(gw *advl4v1alpha1pre1.Gateway) error {
```
sudswasavi, line 210: If AKO is rebooted, we would get a state transition from init-->ready--> etc...and the last updated timestamp would be modified. I think that's fine because think in the `pod` case, if the Kubernetes API dies, it would kill the pod and bring it back up when the controller is online?

sudswasavi (dismissed):

internal/status/advl4_status.go
```diff
@@ -160,9 +200,40 @@ func updateGatewayStatusObject(gw *advl4v1alpha1pre1.Gateway, updateStatus *advl
...
+		Message:            "Initializing",
+		LastTransitionTime: metav1.Now(),
+	}}
+
+	var listenerStatuses []advl4v1alpha1pre1.ListenerStatus
+	for _, listener := range gw.Spec.Listeners {
+		listenerStatuses = append(listenerStatuses, advl4v1alpha1pre1.ListenerStatus{
+			Port: strconv.Itoa(int(listener.Port)),
+			Conditions: []advl4v1alpha1pre1.ListenerCondition{{
+				Type:               "Ready",
```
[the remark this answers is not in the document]
shubhamavi: yes, no pending condition for listeners

internal/k8s/ako_init.go
shubhamavi: added a check in the initialisation function, it shouldn't reinitialise things anymore. thanks

shubhamavi:
build ako

shubhamavi:
build ako

shubhamavi:
build ako

sudswasavi (approved):

PR #89, opened 2020-09-10
moving ako from deployment to statefulset

monotosh-avi:
- statefulset ensures that there is only one pod running at a time if replica count is set to one.
- this would resolve problems which we face because there are multiple pods of ako running at a time

- Also increased wait time for controller response to 300 seconds

Note: serviceName is mandatory in statefulset spec

### DIFF
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: ako
   namespace: {{ .Release.Namespace }}
@@ -7,6 +7,7 @@ metadata:
     {{- include "ako.labels" . | nindent 4 }}
 spec:
   replicas: {{ .Values.replicaCount }}
+  serviceName: ako
   selector:
     matchLabels:
       {{- include "ako.selectorLabels" . | nindent 6 }}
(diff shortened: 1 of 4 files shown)


sudswasavi:
As a part of this change can you also add some resource limits:

```		         resources:
           limits:
             cpu: 250m
             memory: 300Mi
           requests:
             cpu: 100m
             memory: 75Mi```

monotosh-avi:
> As a part of this change can you also add some resource limits:
> 
> ```
>            limits:
>              cpu: 250m
>              memory: 300Mi
>            requests:
>              cpu: 100m
>              memory: 75Mi```
> ```

Added @sudswasavi

parimanur (approved):

sudswasavi:
@rathinikunj / @anish-avi this is meant to break your tests in FT and system test but seems like a simple change.

rathinikunj:
@sudswasavi are we going to merge this in release1.2.1 ? 

Do i need to specify serviceName manually or via values.yaml?

monotosh-avi:
ServiceName is just a value we need to set in statefulset .. we need not have to change that value, the service need not exist .. we would need that if we want to send traffic to ako.

sudswasavi:
> @sudswasavi are we going to merge this in release1.2.1 ?
> 
> Do i need to specify serviceName manually or via values.yaml?

Yeah we will merge this to 1.2.1 based off our discussions from yesterday.

PR #90, opened 2020-09-10
Verify server rootCA chain verification in AKO

sudswasavi:
This commit proposes a change that allows AKO to verify the server
side rootca chain if provided by the values.yaml.

This value is rendered in the secret object and subsequently rendered
in the deployment yaml.

NOTE: The rootca must have a proper SAN IP for it to get validated.


sudswasavi:
@hemantavi FYI for AMKO.

pkg/utils/avi_rest_utils.go
```diff
@@ -50,11 +54,30 @@ func SharedAVIClients() *AviRestClientPool {
 func NewAviRestClientPool(num uint32, api_ep string, username string,
 	password string) (*AviRestClientPool, error) {
 	var p AviRestClientPool
+	ctrlRootCa := os.Getenv("CTRL_ROOTCA")
```
parimanur, line 57: we will have to do a b64decode.

sudswasavi: I don't think so, Kubernetes does the de-coding and populates the environment variable accordingly. That is how the password/username is also getting handled.

parimanur: Ok.

parimanur (approved):

chauhanshubham (approved):

shubhamavi (approved):

PR #92, opened 2020-09-11
moving ako from deployment to statefulset

monotosh-avi:
- statefulset ensures that there is only one pod running at a time if replica count is set to one.
- this would resolve problems which we face because there are multiple pods of ako running at a time

- added default resource limit for ako
- Also increased wait time for controller response to 300 seconds

(cherry picked from commit 9f11e02e83a69f27ee0af2fe057c72760bdecc43)

### DIFF
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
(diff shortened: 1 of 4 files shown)


monotosh-avi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

sudswasavi (approved):

PR #93, opened 2020-09-11
changes in helm templates

monotosh-avi:
added ControllerIP  in configs in values.yaml. 
let me know if it's better to keep it in configs, or under any other tag

### DIFF
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -50,7 +50,7 @@ ControllerSettings:
   serviceEngineGroupName: "Default-Group" # Name of the ServiceEngine Group.
   controllerVersion: "18.2.10" # The controller API version
   cloudName: "Default-Cloud" # The configured cloud name on the Avi controller.
-
+  controllerIP: ""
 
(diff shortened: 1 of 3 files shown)


sudswasavi (commented):
@monotosh-avi would you do this change in 1.2.1 as well for the controllerIP?

helm/ako/values.yaml
```diff
@@ -51,6 +51,8 @@ ControllerSettings:
   controllerVersion: "18.2.10" # The controller API version
   cloudName: "Default-Cloud" # The configured cloud name on the Avi controller.
 
+configs:
```
sudswasavi, line 54: If we move this inside the controllerSettings and modify our helm install instructions, do you see an issue? @monotosh-avi

monotosh-avi: moved

sudswasavi (approved):

apalsule:
@monotosh-avi Please make sure FTs dont break if you cherry-pick this change to 1.2.1

PR #96, opened 2020-09-11
skipping retry for dummy VS for stale object deletion

monotosh-avi:


internal/rest/dequeue_nodes.go
```diff
@@ -381,7 +381,17 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
...
+				models.RestStatus.UpdateAviApiRestStatus(utils.AVIAPI_CONNECTED, nil)
+				utils.AviLog.Debugf("key: %s, msg: rest call executed successfully, will update cache", key)
+				// Add to local obj caches
+				for _, rest_op := range rest_ops {
+					rest.PopulateOneCache(rest_op, aviObjKey, key)
+				}
+
+			} else if aviObjKey.Name == lib.DummyVSForStaleData {
+				utils.AviLog.Warnf("key: %s, msg: error in rest request %v, for %s, won't retry", key, err.Error(), lib.DummyVSForStaleData)
+			} else {
```
sudswasavi, line 394: Can you also put a  fix:
					if publishKey == "" {
						// This is a delete case for the virtualservice. Derive the virtualservice from the 'key'
						splitKeys := strings.Split(key, "/")
						if len(splitKeys) == 2 {
							publishKey = splitKeys[1]
						}
					}
around line 423?

monotosh-avi:
@sudswasavi  why was this PR closed ?

internal/rest/dequeue_nodes.go
monotosh-avi: added

sudswasavi: Thanks

sudswasavi (approved):

PR #103, opened 2020-09-14
wrap avi client inits in goroutines for faster bootup

shubhamavi:
avi controller clients, earlier were being initialised sequentially, wrapped the initialisation sequence with goroutines.
Please feel free to merge this after the release/review extra carefully :). This affects both AKO/AMKO.

### DIFF
--- a/internal/cache/avi_ctrl_clients.go
+++ b/internal/cache/avi_ctrl_clients.go
@@ -56,6 +56,7 @@ func SharedAVIClients() *utils.AviRestClientPool {
 				if err != nil {
 					connectionStatus = utils.AVIAPI_DISCONNECTED
 					utils.AviLog.Error("AVI controller initilization failed")
+					return nil
 				}
 				// set the tenant and controller version in avisession obj
 				for _, client := range AviClientInstance.AviClient {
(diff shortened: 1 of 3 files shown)


internal/cache/avi_ctrl_clients.go
```diff
@@ -55,7 +55,7 @@ func SharedAVIClients() *utils.AviRestClientPool {
 				connectionStatus = utils.AVIAPI_CONNECTED
 				if err != nil {
 					connectionStatus = utils.AVIAPI_DISCONNECTED
-					utils.AviLog.Error("AVI controller initilization failed")
+					utils.AviLog.Fatal("AVI controller initilization failed")
```
sudswasavi, line 58: if you do a fatal here, won't it crash the pod?

sudswasavi: instead relying on the liveness probe is a better option?

shubhamavi: thanks, made the changes. It should shutdown the apiserver and the liveness probe will fail.

pkg/utils/avi_rest_utils.go
```diff
@@ -67,34 +70,49 @@ func NewAviRestClientPool(num uint32, api_ep string, username string,
...
+					session.SetVersion(version)
+					CtrlVersion = version
+				}
+			}
+
+			clientPool.AviClient = append(clientPool.AviClient, aviClient)
+		}()
+	}
+
+	wg.Wait()
```
sudswasavi, line 110: if out of the 8 threads 1 thread is hung, would we require a timeout on the `wait` here? @shubhamavi / @monotosh-avi

sudswasavi (approved):

PR #106, opened 2020-09-14
Do not delete vsvip cache for insecure passthrough VS deletion

monotosh-avi:
- vsvip objects should be deleted from cache only if controoler version
is >= 20.1.1 and the VS is not of type insecure passthrough. Else while
deleting the secure passthrough VS, vsvip won;t be pressent in the cache
and vsvip deletion would be skipped

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/Masterminds/semver"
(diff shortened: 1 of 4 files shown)


monotosh-avi:
build ako

shubhamavi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

sudswasavi (approved):

shubhamavi (approved):

PR #109, opened 2020-09-14
service_api fixes, remove invalidAddress condition, sanitised gateway status condition reasons

shubhamavi:


sudswasavi (approved):
I think we do need the InvalidAddress support on the Gateway for a static that falls outside the configured IPAM networks. But let's take it up in the next PR.

shubhamavi:
> I think we do need the InvalidAddress support on the Gateway for a static that falls outside the configured IPAM networks. But let's take it up in the next PR.

i see a controller limitation there, but will try taking it up next

PR #112, opened 2020-09-15
RemoveListeningPortOnVsDown only for L4 VSes

sudswasavi:
RemoveListeningPortOnVsDown was applicable for the shared VSes as well.
Due to the current behavior of the avi controller this flag would bring
down the shared VS with insecure pools going down, despite of having
functional SNIs.

Hence this commit restricts the changes to L4 VSes only.


monotosh-avi (approved):

apalsule:
@sudswasavi How does it work for the shared VIP we create for passthrough, would we still set this option, if so how does it behave ?

hemantavi (approved):

PR #115, opened 2020-09-15
Multiples Fixes:  log file name and sni vs deletion error handling

monotosh-avi:
Adding timestamp to log file name, as changing file permission did not fix the issue as per system test report. 

Pass model to ExecuteRestAndPopulateCache for sni vs deletion  …
- If we encounter an error while deleting sni vs, if the model is nil, we would  not refresh the cache
and go on retrying forever, as we don't be able to get avimodel.GetRetryCounter().
- The same problem can also happen if we face error while bootup if avimodel is nil. Not making any change
for this now. We can potentially skip retry here or do a slow retry here.

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -241,7 +241,7 @@ func (rest *RestOperations) RestOperation(vsName string, namespace string, avimo
 		utils.AviLog.Infof("key: %s, msg: SNI delete candidates are : %s", key, sni_to_delete)
 		var rest_ops []*utils.RestOp
 		for _, del_sni := range sni_to_delete {
-			rest.SNINodeDelete(del_sni, namespace, rest_ops, key)
+			rest.SNINodeDelete(del_sni, namespace, rest_ops, avimodel, key)
 			rest.ExecuteRestAndPopulateCache(rest_ops, vsKey, avimodel, key)
 		}
 	}
@@ -319,7 +319,7 @@ func (rest *RestOperations) deleteVSOper(vsKey avicache.NamespaceName, vs_cache_
 			sniVsKey, ok := rest.cache.VsCacheMeta.AviCacheGetKeyByUuid(sni_uuid)
 			if ok {
 				delSNI := sniVsKey.(avicache.NamespaceName)
-				rest.SNINodeDelete(delSNI, namespace, rest_ops, key)
+				rest.SNINodeDelete(delSNI, namespace, rest_ops, nil, key)
 			}
 		}
 		if !skipVS {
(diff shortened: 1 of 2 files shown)


sudswasavi:
@monotosh-avi on retry - if the operation is DELETE and the response is 404, can we not skip publishing the model for retry?

sudswasavi:
A rough code block could look like this:

image

monotosh-avi:
> @monotosh-avi on retry - if the operation is DELETE and the response is 404, can we not skip publishing the model for retry?

We are facing this error in case of delete

                    returned err null
2020-09-15T18:00:12.379Z    WARN    rest/dequeue_nodes.go:405   key: admin/cluster6--Shared-L7-1, msg: there was an error sending the macro Error during DELETE: Encountered an error on DELETE request to URL https://10.79.204.46//api/virtualservice/virtualservice-e37fac05-622a-4cf3-ad28-99ee5934e51d: HTTP code: 404; error from Avi: map[error:VirtualService object not found!]
rest/dequeue_nodes.go:426   key: admin/cluster6--Shared-L7-1, msg: Avi model not set

Do you think we need too do a cache refresh ?

sudswasavi:
> > @monotosh-avi on retry - if the operation is DELETE and the response is 404, can we not skip publishing the model for retry?
> 
> We are facing this error in case of delete
> 
> ```
> 
> returned err null
> 2020-09-15T18:00:12.379Z    WARN    rest/dequeue_nodes.go:405   key: admin/cluster6--Shared-L7-1, msg: there was an error sending the macro Error during DELETE: Encountered an error on DELETE request to URL https://10.79.204.46//api/virtualservice/virtualservice-e37fac05-622a-4cf3-ad28-99ee5934e51d: HTTP code: 404; error from Avi: map[error:VirtualService object not found!]
> rest/dequeue_nodes.go:426   key: admin/cluster6--Shared-L7-1, msg: Avi model not set
> ```
> 
> Do you think we need too do a cache refresh ? @sudswasavi

Yeah I think based on the above code snippet that I pasted, the only thing I can think of adding to that is the cache delete. What do you think?

monotosh-avi:
build ako

internal/rest/dequeue_nodes.go
```diff
@@ -423,8 +423,18 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
...
+							utils.AviLog.Warnf("key: %s, msg: Avi model not set, possibly a DELETE call", key)
+							aviError, ok := rest_ops[i].Err.(session.AviError)
+							// If it's 404, don't retry
+							if ok {
+								statuscode := aviError.HttpStatusCode
+								if statuscode != 404 {
+									rest.PublishKeyToSlowRetryLayer(publishKey, aviclient, key)
+									return
+								} else {
+									rest.AviVsCacheDel(rest_ops[i], aviObjKey, key)
```
sudswasavi, line 435: Can this come here during a static route deletion? @parimanur / @monotosh-avi

sudswasavi (approved):
@parimanur can you PTAL and merge?

parimanur (approved):

internal/rest/dequeue_nodes.go
parimanur: As discussed we dont set the model to nil for vrfcontext.

PR #117, opened 2020-09-17
Allow AKO to work in psp enabled environments

sudswasavi:
This includes a psp template needed by AKO and a corresponding
flag in the values.yaml that allows the policy to be used.

The clusterrole is modified as per.

### DIFF
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -61,10 +61,13 @@ resources:
     memory: 300Mi
   requests:
     cpu: 100m
-    memory: 75Mi
+    memory: 200Mi
 
 podSecurityContext: {}
-  # fsGroup: 2000
+
+rbac:
+  # Creates the pod security policy if set to true
+  pspEnable: false
 
 
 avicredentials:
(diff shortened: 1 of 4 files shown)


helm/ako/values.yaml
```diff
@@ -60,11 +60,12 @@ resources:
     cpu: 250m
     memory: 300Mi
   requests:
-    cpu: 100m
-    memory: 75Mi
+    cpu: 250m
+    memory: 200Mi
 
-podSecurityContext: {}
```
monotosh-avi, line 66: Can we keep the field podSecurityContext ? In some cases, this may be required.

monotosh-avi (dismissed):

helm/ako/values.yaml
sudswasavi: re-instated. thx.

monotosh-avi (approved):

sudswasavi:
@hemantavi AMKO would need this too.

shubhamavi (approved):

PR #118, opened 2020-09-17
Allow AKO to work in psp enabled environments

sudswasavi:
This includes a psp template needed by AKO and a corresponding
flag in the values.yaml that allows the policy to be used.

The clusterrole is modified as per.

Additional fix for vs port down for passthrough-secure.


monotosh-avi:
build ako

monotosh-avi (approved):

shubhamavi (approved):

PR #123, opened 2020-09-21
static IP on vsvip rest layer changes and gw status updates

shubhamavi:
With this we allow PUT requests to vsvip in case of ip address changes, but in case there are vsvip IPAddress change related errors from the controller, they are reported back in the gateway status with reason `InvalidAddress`.
The two strings we look out for as part of those errors are:
- "Changing an existing VIP's vip_id is not supported"
- "If a new preferred IP is needed, please recreate the VIP"

### DIFF
--- a/internal/rest/avi_obj_vs.go
+++ b/internal/rest/avi_obj_vs.go
@@ -25,11 +25,11 @@ import (
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/nodes"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/status"
-
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
 
 	avimodels "github.com/avinetworks/sdk/go/models"
 	"github.com/davecgh/go-spew/spew"
+	corev1 "k8s.io/api/core/v1"
 )
 
 const VSVIP_NOTFOUND = "VsVip object not found"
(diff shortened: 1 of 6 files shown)


internal/rest/avi_obj_vs.go
```diff
@@ -796,8 +790,35 @@ func (rest *RestOperations) AviVsVipDel(uuid string, tenant string, key string)
...
+				}
+
+				status.UpdateGatewayStatusGWCondition(gw, &status.UpdateGWStatusConditionOptions{
+					Type:    "Pending",
+					Status:  corev1.ConditionTrue,
+					Reason:  "InvalidAddress",
+					Message: rest_op.Message,
+				})
+				status.UpdateGatewayStatusObject(gw, &gw.Status)
+				utils.AviLog.Warnf("key: %s, msg: ", key, rest_op.Message)
```
sudswasavi, line 816: Can you pls remove this message?

internal/rest/dequeue_nodes.go
```diff
@@ -415,7 +416,11 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
 									utils.AviLog.Infof("key: %s, msg: Error is not of type AviError, err: %v, %T", key, rest_ops[i].Err, rest_ops[i].Err)
 									continue
 								}
-								retryable, fastRetryable := rest.RefreshCacheForRetryLayer(publishKey, aviObjKey, rest_ops[i], aviError, aviclient, avimodel, key)
+								retryable, fastRetryable, pushWithMessage := rest.RefreshCacheForRetryLayer(publishKey, aviObjKey, rest_ops[i], aviError, aviclient, avimodel, key)
+								if pushWithMessage {
```
sudswasavi, line 420: Why are we populating the cache on `this` error where the IP address request couldn't be fulfilled?

Isn't it much simpler to just handle this error as a special case outside of the RefreshCacheRetryLayer method? Since technically you aren't going to retry anyway?

@monotosh-avi pls review this code.

monotosh-avi: Yes I think that would be better. 
If the error is of this special type, then we can handle it before line no : 409, and continue to the next error. 
Also, what happens if we need to retry for another error in the same error ? We would again see this special vsvip error  .. but I guess there is no way to avoid that.

sudswasavi (dismissed):

monotosh-avi (approved):

sudswasavi (approved):

PR #125, opened 2020-09-23
Avoid unnecessary PUT calls to the Avi controller for Vsvip updates

sudswasavi:
This commit addresses the issue of unnecessary PUT calls for VS
vip updates during AKO reboot.

Modified the logic of calculating checksum. Thing to note is,
an empty string when done a hash on, results in a non-zero
checksum. This is fixed as a part of the PR.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -108,8 +108,8 @@ func GetL4PoolName(vsName string, port int32) string {
 	return vsName + "--" + strconv.Itoa(int(port))
 }
 
-func GetAdvL4PoolName(svcName, namespace string, port int32) string {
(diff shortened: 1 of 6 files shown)


monotosh-avi (approved):

parimanur (dismissed):

sudswasavi:
@monotosh-avi / @parimanur / @shubhamavi since the vsvip update is an independent call, I have moved it as the first call during the object creation. This would ensure that if the vsvip creation fails, the other stale objects won't get generated.

sudswasavi:
Added fix to handle gateway object transitions from the service object.

monotosh-avi (approved):

hemantavi (approved):

PR #132, opened 2020-09-29
Changing CONTRIBUTING file to point to latest AKO repo.

akshayhavile:
Changing upstream url to point to appropriate AKO Repo.

### DIFF
--- a/Dockerfile.ako
+++ b/Dockerfile.ako
@@ -10,5 +10,6 @@ RUN GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build -o $GOPATH/bin/akc -mod=vendo
 FROM photon:latest
 RUN yum install -y tar.x86_64
 COPY --from=build /go/bin/akc .
+COPY ./scripts/pre_stop_hook.sh /var/pre_stop_hook.sh
 EXPOSE 8080
 ENTRYPOINT ["./akc"]
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -40,6 +40,10 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          lifecycle:
+            preStop:
+              exec:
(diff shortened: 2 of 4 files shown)


shubhamavi (approved):

akshayhavile:
This PR contains two changes.
1. Contirbuting File changes: This points to latest AKO repo.
2. Fix for AV-92959:ako: log files are not rolling over thus filling up the storage with Prestop hook for pod.

Tested on single node K8 Cluster.

akshayhavile:
Updated code as per review comments.

helm/ako/templates/statefulset.yaml
```diff
@@ -40,6 +40,10 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh","-c","if [ $USE_PVC ]; then cd $LOG_FILE_PATH; old_log_tar_file=$LOG_FILE_PATH/old_logs.tar ;old_pods_gz_file=$LOG_FILE_PATH/old_logs.tar.gz; if [ ! -f \"$old_pods_gz_file\" ] ; then tar -cf $old_log_tar_file $POD_NAME*.log* ; else gunzip $old_pods_gz_file;tar rf $old_log_tar_file $POD_NAME*.log*;fi; rm -rf $POD_NAME*.log*; gzip $old_log_tar_file; fi;"]
```
monotosh-avi, line 46: Would it be better if we ad this in a script inside the docker file and call that script ?
The script is bot long, so  it would be easier to understand and later we can add to that script if required. 

@sudswasavi  what do you think ?

akshayhavile: @monotosh-avi: If I understand correctly, you are suggesting: check-in shell script in AKO git repo.. Write copy statement in docker file to cp that file inside container at specific path (say at /var/script.sh) while building ako image and then call that script from pre-stop hook.  Sounds good. 
Only thing is - any changes in code needs to be done by logging into docker instead of 'kubectl edit' command.

monotosh-avi: yes @akshayhavile, we can add the script in the ako container image.
We should be able to catch any issue in our testing and I think customer would't have to modify this script.

sudswasavi: A script:
  - is good for debuggability.
  - is baked into the image so can't be changed by the customer.

Unless we have any side effects of using the scripts, that will check through our course of system test,

Dockerfile.ako
```diff
@@ -10,5 +10,6 @@ RUN GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build -o $GOPATH/bin/akc -mod=vendo
 FROM photon:latest
 RUN yum install -y tar.x86_64
 COPY --from=build /go/bin/akc .
+COPY ./scripts/pre_stop_hook.sh /var/pre_stop_hook.sh
```
akshayhavile, line 13: @sudswasavi : Can we avoid this hardcoding of path and mention relative path? as Dockerfile.ako will be running from AKO main folder so I didn't try to mention relative path

sudswasavi: Yes feel free to change in another PR. I will go ahead and merge this one.

sudswasavi (approved):

monotosh-avi (approved):

PR #133, opened 2020-09-29
Don't reset status if it's already reset.

sudswasavi:
This commit would not reset the LB status, if it's already reset.


monotosh-avi (approved):

shubhamavi (approved):

shubhamavi:
cherry-pick candidate

PR #134, opened 2020-09-29
changes in retry logic

monotosh-avi:
- Do a fast retry on failure due to session expiry
- If any error occurs during execution of a rest call to avi,
then return false to the caller, so that no subsequent requests are made.

Without this change, we may land into a inconsient state. E.g. during clean up
of VSes for passthrough routes, if we get an error in deletion of insecure
passthrough VS, possibly due to session expiry, we would continute with the deletion
of the secure passthrough VS. But deletrion of the vsvip would fail.

On retry, insecure passthrough VS would be deleted, but we would get a 404 as secure
passthrough VS is already deleted, leading to stale pool objects.

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -185,7 +185,9 @@ func (rest *RestOperations) RestOperation(vsName string, namespace string, avimo
 			rest_ops = append(rest_ops, restOp...)
 
 		}
-		rest.ExecuteRestAndPopulateCache(rest_ops, vsKey, avimodel, key)
+		if success := rest.ExecuteRestAndPopulateCache(rest_ops, vsKey, avimodel, key); !success {
+			return
+		}
 	} else {
 		var rest_ops []*utils.RestOp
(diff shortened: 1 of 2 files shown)


monotosh-avi:
build ako

sudswasavi:
@shubhamavi this is a good change for 1.3.1 too?

internal/rest/dequeue_nodes.go
```diff
@@ -450,9 +471,11 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
 				if retry {
 					rest.PublishKeyToRetryLayer(publishKey, aviclient, key)
 				}
+				return false
```
shubhamavi, line 474: can we have any logs while we return false from this function?

shubhamavi: i think the one on L426 should be fine

shubhamavi (approved):

sudswasavi:
@monotosh-avi can you pls cherry-pick this to 1.3.1 as well?

sudswasavi (approved):

PR #137, opened 2020-09-30
changes in retry logic

monotosh-avi:
- Do a fast retry on failure due to session expiry
- If any error occurs during execution of a rest call to avi,
then return false to the caller, so that no subsequent requests are made.

Without this change, we may land into a inconsient state. E.g. during clean up
of VSes for passthrough routes, if we get an error in deletion of insecure
passthrough VS, possibly due to session expiry, we would continute with the deletion
of the secure passthrough VS. But deletrion of the vsvip would fail.

On retry, insecure passthrough VS would be deleted, but we would get a 404 as secure
passthrough VS is already deleted, leading to stale pool objects.

(cherry picked from commit 88e7bc383a0c6c502a2993c430856dd8f2a367cf)


monotosh-avi:
build ako

PR #139, opened 2020-09-30
Use ldflags to print AKO version

sudswasavi:
This commit introduces a way by which one should be able to pass
a ldflag argument to AKO and AKO on bootup would print the version.

This will help us recognize the running AKO version.


shubhamavi:
build ako

shubhamavi (approved):

parimanur (approved):

PR #140, opened 2020-09-30
Disable full sync in advanced l4.

sudswasavi:
Advanced L4 does not support DNS. Hence the full sync thread is
disabled because it's mainly used for fetching DNS info from the
controller.


monotosh-avi:
build ako

hemantavi (approved):

monotosh-avi (approved):

PR #143, opened 2020-10-01
bulk gateway status updates, listener condition type fixes

shubhamavi:


internal/status/advl4_status.go
```diff
@@ -221,38 +248,104 @@ func UpdateGatewayStatusObject(gw *advl4v1alpha1pre1.Gateway, updateStatus *advl
...
+		gw, err := lib.GetAdvL4Clientset().NetworkingV1alpha1pre1().Gateways(nsNameSplit[0]).Get(nsNameSplit[1], metav1.GetOptions{})
+		if err != nil {
+			utils.AviLog.Warnf("msg: Could not get the gateway object for UpdateStatus: %s", err)
+			// retry get if request timeout
+			if strings.Contains(err.Error(), utils.K8S_ETIMEDOUT) {
+				return getGateways(gwNSNames, bulk, retry+1)
+			}
+		}
+
+		gwMap[gw.Namespace+"/"+gw.Name] = gw
```
sudswasavi, line 347: seems like if we get an error, then also we will hit this code block and nil pointer would be seen.

internal/status/advl4_status.go
```diff
@@ -221,38 +248,104 @@ func UpdateGatewayStatusObject(gw *advl4v1alpha1pre1.Gateway, updateStatus *advl
...
+	if bulk {
+		gwList, err := lib.GetAdvL4Clientset().NetworkingV1alpha1pre1().Gateways("").List(metav1.ListOptions{})
+		if err != nil {
+			utils.AviLog.Warnf("Could not get the gateway object for UpdateStatus: %s", err)
+			// retry get if request timeout
+			if strings.Contains(err.Error(), utils.K8S_ETIMEDOUT) {
+				return getGateways(gwNSNames, bulk, retry+1)
+			}
+		}
+		for i := range gwList.Items {
```
sudswasavi, line 329: Seems like if there's an error we would still try to iterate the list.

sudswasavi (commented):
Due to urgency of this fix, I will go ahead and make amends to the branch with the fixes.

internal/status/advl4_status.go
```diff
@@ -40,48 +40,63 @@ type UpdateGWStatusConditionOptions struct {
...
+			// assuming 1 IP per gateway
+			gwStatus := gw.Status.DeepCopy()
+			gwStatus.Addresses = []advl4v1alpha1pre1.GatewayAddress{{
+				Value: option.Vip,
+				Type:  advl4v1alpha1pre1.IPAddressType,
+			}}
+
+			// when statuses are synced during bootup
+			InitializeGatewayConditions(gwStatus, &gw.Spec, true)
+			UpdateGatewayStatusGWCondition(gwStatus, &UpdateGWStatusConditionOptions{
```
sudswasavi, line 71: if you update this to the Gateway and then call the UpdateGatewayStatusObject, it would always result into a diff and hence `DeepEqual` inside the UpdateGatewayStatusObject method will not be useful anymore. Hence I see that we are trying to update the Gateway status on every AKO boot up.

shubhamavi:
build ako

shubhamavi:
build ako

PR #147, opened 2020-10-06
Do fast retry only for 401 error

monotosh-avi:
Currently we get same response from sdk for both session expiry (401) and 500 error.
    Modified sdk code in vendor to get response for 401 error so that it can be handled separately.

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -410,12 +410,19 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
 						publishKey = splitKeys[1]
 					}
 				}
-				_, validAviError := err.(session.AviError)
-				if !validAviError && avimodel != nil && avimodel.GetRetryCounter() != 0 {
-					// We may face this error if the connection is idle for some time
-					utils.AviLog.Errorf("key: %s, error in parsing the error to avi error: %v, adding to fast retry queue", key, err.Error())
-					rest.PublishKeyToRetryLayer(publishKey, aviclient, key)
-					return false
+
+				if webSyncErr, ok := err.(*utils.WebSyncError); ok {
+					aviError, ok := webSyncErr.GetWebAPIError().(session.AviError)
+					if ok && aviError.HttpStatusCode == 401 {
+						if avimodel != nil && avimodel.GetRetryCounter() != 0 {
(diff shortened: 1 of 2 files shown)


internal/rest/dequeue_nodes.go
```diff
@@ -410,13 +410,7 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
...
 					}
 				}
-				_, validAviError := err.(session.AviError)
-				if !validAviError && avimodel != nil && avimodel.GetRetryCounter() != 0 {
-					// We may face this error if the connection is idle for some time
-					utils.AviLog.Errorf("key: %s, error in parsing the error to avi error: %v, adding to fast retry queue", key, err.Error())
-					rest.PublishKeyToRetryLayer(publishKey, aviclient, key)
-					return false
-				}
+
```
sudswasavi, line 413: Let's say AKO is idle, someone creates and ingress. The session would have expired. We would wait for 2 mins before the ingress syncs.

monotosh-avi:
build ako

monotosh-avi:
build ako

vendor/github.com/avinetworks/sdk/go/session/avisession.go
```diff
@@ -546,13 +546,17 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
...
 
 		if resp.StatusCode == 401 && len(avisess.sessionid) != 0 && uri != "login" {
-			resp.Body.Close()
 			glog.Infof("Retrying url %s; retry %d due to Status Code %d", url, retry, resp.StatusCode)
 			err := avisess.initiateSession()
 			if err != nil {
+				resp.Body.Close()
 				return nil, err
 			}
+			if avisess.ctrlStatusCheckDisabled {
```
hemantavi, line 555: Is the intention of not putting a `resp.Body.Close()` in this check is to check the response's body in the AKO's rest layer caller function?

monotosh-avi: The response has to be read again. This is usually done in the function fetchBody(), which calls resp.Body.Close(), and fails if resp.Body is already closed.

hemantavi: Got it. Thanks!

sudswasavi (approved):

PR #150, opened 2020-10-08
use patch call to update status for ingress and route

monotosh-avi:
patch call would make sure that update doesn't fail if the object version has changed


monotosh-avi:
build ako

monotosh-avi:
build ako

sudswasavi (approved):

PR #151, opened 2020-10-08
AKO Stability Fix for invalid clustername

akshayhavile:
This Commit addresses: 
1. IF clustername is invalid, do not delete stale VS and objects.
2. Adding header to pre_stop_hook
3. Adding env variable for running ako locally

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -149,13 +149,14 @@ AKO can be tested from your laptop. Please follow the instructions:
         export CTRL_PASSWORD=<password>
         export CTRL_IPADDRESS=<controller-ip>
         export CTRL_VERSION=<controller-api-version>
-        export SHARD_VS_SIZE=<LARGE.MEDIUM,SMALL>
+        export SHARD_VS_SIZE=<LARGE,MEDIUM,SMALL>
         export FULL_SYNC_INTERVAL=1800
         export CLOUD_NAME=<Avi-cloud-name>
         export CLUSTER_NAME=<your-unique-cluster-name>
         export NODE_NETWORK_LIST='[{"networkName":"Nw1","cidrs":["10.79.168.0/22"]}]'
-        export SEG_NAME="Default-Group"
+        export SEG_NAME=<se-group-name>
         
+     NODE_NETWORK_LIST is a managment network name and a CIDR of that network. SEG_NAME can be empty for advance L4 deployment.
      You can control additional settings by exporting respective variables from inside the deployment [file](https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/helm/ako/templates/statefulset.yaml).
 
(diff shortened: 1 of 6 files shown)


go.sum
```diff
@@ -35,6 +35,7 @@ github.com/avinetworks/sdk v0.0.0-20200910040148-2ed48a016241/go.mod h1:BcllDeAF
 github.com/avinetworks/sdk v0.0.0-20200910070359-d9ffda19a7dd h1:W+4xqQczmklVpG+JQlY9nBQaSbhbv3MvGaFLX6UvNrY=
 github.com/avinetworks/sdk v0.0.0-20200910070359-d9ffda19a7dd/go.mod h1:BcllDeAFx8PtaMrPxvvuCUo7NRS2x6w+3W17WFDu0sk=
 github.com/avinetworks/sdk v0.0.0-20200910092938-580bed7585cb h1:i6jbYz4bcAN6ropnUUmfG0RCQv+w/SC/MkCXCIUUaAs=
+github.com/avinetworks/sdk v0.0.0-20200921035123-787ceaed92c5 h1:yTqiP3xkjLNhfC7NOkWYz7YxaFhWpGLlSxBJbluJofg=
```
sudswasavi, line 38: Is this bringing in a new SDK version? You don't seem to have any SDK related changes.

CONTRIBUTING.md
```diff
@@ -154,6 +154,7 @@ AKO can be tested from your laptop. Please follow the instructions:
         export CLOUD_NAME=<Avi-cloud-name>
         export CLUSTER_NAME=<your-unique-cluster-name>
         export NODE_NETWORK_LIST='[{"networkName":"Nw1","cidrs":["10.79.168.0/22"]}]'
+        export SEG_NAME="Default-Group"
```
sudswasavi, line 157: Maybe we want to specify in what conditions the "Default-Group" has to be specified?
For advancedL4 if this value isn't set, then the code would assume "Default-Group".
If you specify the "Default-Group" here, it seems like the default use case, but users can select a non-default SE group as well. So maybe I will just alter it and say:
export SEG_NAME=<SE Group name>

go.sum
sudswasavi: If not, then can we checkout go.sum and reset it to what it was before this commit and do a push?

sudswasavi (commented):
Looks good overall. Couple of minor nits.

go.sum
akshayhavile: Sure .. I will revert it back to original.

akshayhavile:
Changed code as per review request.

sudswasavi (approved):

sudswasavi:
Thanks @akshayhavile . Can you pls cherry-pick this to the release-1.3.1 branch as well?

sudswasavi:
I think you can just selectively make the fix for invalid cluster name.

PR #153, opened 2020-10-09
adding retry for vsvip build failure

monotosh-avi:
For vsvip build, we first do a get in layer3. If this fails, we should do a retry.


sudswasavi (approved):

sudswasavi:
LGTM

PR #155, opened 2020-10-09
AKO-Stability-issue:

akshayhavile:
This commit cherry-picks from master:
1. IF clustername is invalid, do not delete stale VS and objects.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -59,8 +59,8 @@ func PopulateCache() error {
 	aviclient := avicache.SharedAVIClients()
 	restlayer := rest.NewRestOperations(avi_obj_cache, aviclient)
 	staleVSKey := utils.ADMIN_NS + "/" + lib.DummyVSForStaleData
-	utils.AviLog.Infof("Starting clean up of stale objects")
-	if aviclient != nil && len(aviclient.AviClient) > 0 {
+	if lib.IsClusterNameValid() && aviclient != nil && len(aviclient.AviClient) > 0 {
+		utils.AviLog.Infof("Starting clean up of stale objects")
 		restlayer.CleanupVS(staleVSKey, true)
 		staleCacheKey := avicache.NamespaceName{
 			Name:      lib.DummyVSForStaleData,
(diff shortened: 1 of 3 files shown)


akshayhavile:
build ako

monotosh-avi (approved):

internal/k8s/ako_init.go
```diff
@@ -59,8 +59,8 @@ func PopulateCache() error {
 	aviclient := avicache.SharedAVIClients()
 	restlayer := rest.NewRestOperations(avi_obj_cache, aviclient)
 	staleVSKey := utils.ADMIN_NS + "/" + lib.DummyVSForStaleData
-	utils.AviLog.Infof("Starting clean up of stale objects")
-	if aviclient != nil && len(aviclient.AviClient) > 0 {
+	if lib.IsClusterNameValid() && aviclient != nil && len(aviclient.AviClient) > 0 {
```
hemantavi, line 62: I am assuming that this function `PopulateCache()` during bootup is being called by `InitializeAKC()`. With that assumption, is it right to return `nil` if the cluster name is invalid? Would it better if we return an error if cluster name was invalid, so that, we disable sync in the caller function? @akshayhavile @monotosh-avi

monotosh-avi: Ideally we shouldn't reach populate cache, because we call HandleConfigMap() before this, which would call ValidateUserInput() to validate the clusterName. So we are safe. 
But this we can make this change, it's better to return a non-nil value are

akshayhavile: @hemantavi We are disabling sync by this flow: HandleConfigMap()-->ValidateUserInput()-->ValidateClustName .. This flow has been called before PopulateCache() in InitializeAKC() function... So we are taking care of disabling sync.

hemantavi: Makes sense, but to be safe, since this function is also called from multiple places, we should return a non-nil error from here. You can take care of this in a subsequent PR @akshayhavile .

hemantavi (approved):

internal/k8s/ako_init.go
akshayhavile: sure

PR #161, opened 2020-10-14
gateway finalizers implementation

shubhamavi:
if an IP address is present on the gateway object, it is fair to assume that the gateway corresponds to a VS in avi
in case the IP is not present, the gateway can be deleted freely since deleting that would be a NOOP for AKO
so we add finalizer when an IP is updated to the gateway, and remove it when we delete the IP address.
The finalizer we attach to the gateway object is `gateway.ako.vmware.com`

This approach takes care of reboot scenarios, updates from the gateway AND service event handler, and of-course layer3 workflows.

### DIFF
--- a/internal/lib/advl4.go
+++ b/internal/lib/advl4.go
@@ -15,8 +15,13 @@
 package lib
 
 import (
+	"encoding/json"
+
+	advl4v1alpha1pre1 "github.com/vmware-tanzu/service-apis/apis/v1alpha1pre1"
 	advl4crd "github.com/vmware-tanzu/service-apis/pkg/client/clientset/versioned"
 	advl4informer "github.com/vmware-tanzu/service-apis/pkg/client/informers/externalversions/apis/v1alpha1pre1"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
+	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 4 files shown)


internal/lib/advl4.go
```diff
@@ -44,3 +49,35 @@ func SetAdvL4Informers(c *AdvL4Informers) {
...
+	UpdateGatewayFinalizer(gw)
+}
+
+func CheckAndSetGatewayFinalizer(gw *advl4v1alpha1pre1.Gateway) {
+	finalizers := gw.GetFinalizers()
+	if !ContainsFinalizer(gw, GatewayFinalizer) {
+		finalizers = append(finalizers, GatewayFinalizer)
+	}
+
+	gw.SetFinalizers(finalizers)
```
sudswasavi, line 65: Looks like this function can be optimised. If you discover that the finalisers are already present on the gateway, there's no need to re-update it right?

shubhamavi: yes! thanks

sudswasavi:
Thanks @shubhamavi

sudswasavi:
Can you pls cherry-pick to master?

apalsule:
@shubhamavi Can you update the PR with all the testing done for this one ? As this is super sensitive, want to make sure we have all scenarios covered.

PR #162, opened 2020-10-14
Shutdown API Server if Avi Controller password is changed

monotosh-avi:
If controller pasword is changed while AKO is running, we can check
the error msg of the HTTP response. We would get this after session gets timed out
with HTTP Response code 401. The error msg would contain the string - Invalid credentials.

Then we would shutdown API server and AKO pod would get rebooted.

(cherry picked from commit d78a9fb229324a15aa2e86efd2debcea3c9d2089)

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -165,7 +165,10 @@ func (rest *RestOperations) CheckAndPublishForRetry(err error, publishKey, key s
 	if webSyncErr, ok := err.(*utils.WebSyncError); ok {
 		aviError, ok := webSyncErr.GetWebAPIError().(session.AviError)
 		if ok && aviError.HttpStatusCode == 401 {
-			if avimodel != nil && avimodel.GetRetryCounter() != 0 {
+			if strings.Contains(*aviError.Message, "Invalid credentials") {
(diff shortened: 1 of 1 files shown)


internal/rest/dequeue_nodes.go
```diff
@@ -165,7 +165,10 @@ func (rest *RestOperations) CheckAndPublishForRetry(err error, publishKey, key s
 	if webSyncErr, ok := err.(*utils.WebSyncError); ok {
 		aviError, ok := webSyncErr.GetWebAPIError().(session.AviError)
 		if ok && aviError.HttpStatusCode == 401 {
-			if avimodel != nil && avimodel.GetRetryCounter() != 0 {
+			if strings.Contains(*aviError.Message, "Invalid credentials") {
```
apalsule, line 168: we have to rely on strings ? is that what we get even if session expires ? Is there something similar even if certificate is changed ?

sudswasavi: We had this discussion and I was in the middle drafting a mail to the controller folks. We are relying on such strings at multiple places. There's no good way to use constants here at the moment. 401 is used for multiple things, one more example:
```401; error from Avi: map[detail:Authentication credentials were not provided.]```

PR #166, opened 2020-10-19
Increase informer re-sync value and reduction of configmap informer cache

sudswasavi:
Thi commit increases the value of informer re-sync period to 12 hours
This value was currently set to 30 seconds, which would mean that the
kubernetes API server would be polled too frequently for full sync.

This commit sets the value as used in the kube api server defaults:
https://github.com/kubernetes/kubernetes/blob/release-1.17/pkg/controller/apis/config/v1alpha1/defaults.go#L120

In addition to this, we also restrict the configmap informers to listen on specific namespaces for both advancedL4 and non-advancedL4. This brings down the indexer cache consumption significantly for systems that have huge number of configmaps.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -147,17 +147,27 @@ func RandomSeq(n int) string {
 var informer sync.Once
 var informerInstance *Informers
 
-const informerDefaultResync = 12 * time.Hour
-
-func instantiateInformers(kubeClient KubeClientIntf, registeredInformers []string, ocs oshiftclientset.Interface, namespace string) *Informers {
+func instantiateInformers(kubeClient KubeClientIntf, registeredInformers []string, ocs oshiftclientset.Interface, namespace string, adv_l4 bool) *Informers {
 	cs := kubeClient.ClientSet
-	var kubeInformerFactory kubeinformers.SharedInformerFactory
+	var kubeInformerFactory, configMapInformerFactory kubeinformers.SharedInformerFactory
 	if namespace == "" {
-		kubeInformerFactory = kubeinformers.NewSharedInformerFactoryWithOptions(cs, informerDefaultResync)
+		kubeInformerFactory = kubeinformers.NewSharedInformerFactoryWithOptions(cs, InformerDefaultResync)
 	} else {
 		// The informer factory only allows to initialize 1 namespace filter. Not a set of namespaces.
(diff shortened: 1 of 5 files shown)


monotosh-avi (approved):

monotosh-avi:
build ako

pkg/utils/utils.go
```diff
@@ -147,14 +147,16 @@ func RandomSeq(n int) string {
...
 var informerInstance *Informers
 
+const informerDefaultResync = 12 * time.Hour
+
 func instantiateInformers(kubeClient KubeClientIntf, registeredInformers []string, ocs oshiftclientset.Interface, namespace string) *Informers {
 	cs := kubeClient.ClientSet
 	var kubeInformerFactory kubeinformers.SharedInformerFactory
 	if namespace == "" {
-		kubeInformerFactory = kubeinformers.NewSharedInformerFactoryWithOptions(cs, time.Second*30)
+		kubeInformerFactory = kubeinformers.NewSharedInformerFactoryWithOptions(cs, informerDefaultResync)
```
hemantavi, line 156: Can we please take the resync time as an argument to `instantiateInformers`? If no argument is provided, then we can take the default value of `informerDefaultResync`?

sudswasavi: Done.

akshayhavile:
@sudswasavi : Out of context ask: can we take out this line https://github.com/sudswasavi/load-balancer-and-ingress-services-for-kubernetes-1/blob/sudswas-informer-sync/cmd/ako-main/main.go#L114? 

We are calling same function on line #116

sudswasavi:
> @sudswasavi : Out of context ask: can we take out this line https://github.com/sudswasavi/load-balancer-and-ingress-services-for-kubernetes-1/blob/sudswas-informer-sync/cmd/ako-main/main.go#L114?
> 
> We are calling same function on line #116



> @sudswasavi : Out of context ask: can we take out this line https://github.com/sudswasavi/load-balancer-and-ingress-services-for-kubernetes-1/blob/sudswas-informer-sync/cmd/ako-main/main.go#L114?
> 
> We are calling same function on line #116

Nice catch. Done.

hemantavi:
build ako

hemantavi (approved):

monotosh-avi (approved):

hemantavi (approved):

PR #168, opened 2020-10-22
persist VS in avi for broken gw-svc relationship if gateway is valid

shubhamavi:


internal/nodes/dequeue_ingestion.go
```diff
@@ -165,32 +165,11 @@ func isGatewayDelete(gatewayKey string, key string) bool {
 		return true
 	}
 
-	found, gwListeners := objects.ServiceGWLister().GetGWListeners(namespace + "/" + gwName)
+	found, _ := objects.ServiceGWLister().GetGWListeners(namespace + "/" + gwName)
```
sudswasavi, line 168: This would just persist the VS but how about init conditions?
Let's say you had a gateway but no service of type LBs mapped to them. What's going to happen then? Would we create a VS or we will skip?

shubhamavi: we create a VS only @sudswasavi, intended functionality

sudswasavi: Sounds good.

monotosh-avi (approved):

sudswasavi (approved):

PR #176, opened 2020-10-30
Allow System-TCP-Fast-Path only as the network profile for adv l4

sudswasavi:
Since advancedl4 is meant to run with the ESSENTIALS Avi license,
changing the network profile to always use System-TCP-Fast-Path
instead of proxy.


hemantavi (approved):

monotosh-avi (approved):

monotosh-avi:
build ako

PR #178, opened 2020-10-30
Enable RHI as a AKO wide setting

sudswasavi:
This commit introduces a `enableRHI` flag that allows the operator
to setup BGP peering for the virtual services created by AKO as a
cluster wide setting.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -258,6 +258,15 @@ func GetNamespaceToSync() string {
 	return ""
 }
 
+func GetEnableRHI() bool {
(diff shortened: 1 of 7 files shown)


internal/lib/lib.go
```diff
@@ -258,6 +258,13 @@ func GetNamespaceToSync() string {
 	return ""
 }
 
+func GetEnableRHI() bool {
```
parimanur, line 261: log the error to let the user know that we are defaulting this to false?

sudswasavi: Debug logged it.

parimanur (approved):

monotosh-avi (approved):

PR #183, opened 2020-11-08
For Ingress: Namespace wise migration changes

akshayhavile:
Addresses: namespace wise migration changes for L7.
---
TODO: Integration test case. 
             Updating helm-chart values.md.
------
Tested :
1. Valid/Invalid namespace add
2. Valid/Invalid namespace deletion
3. Valid/Invalid ingress add and deletion

Above test case with NSmigration flag enabled and disabled--> tested.

### DIFF
--- a/helm/ako/templates/configmap.yaml
+++ b/helm/ako/templates/configmap.yaml
@@ -26,6 +26,9 @@ data:
   {{ if .Values.AKOSettings.syncNamespace  }}
   syncNamespace: {{ .Values.AKOSettings.syncNamespace | quote }}
   {{ end }}
+  {{ if .Values.AKOSettings.namespaceSyncLabel  }}
+  namespaceSyncLabel: {{ .Values.AKOSettings.namespaceSyncLabel | quote }}
+  {{ end }}
   serviceType:  {{ .Values.L7Settings.serviceType | quote }}
   {{ if eq .Values.L7Settings.serviceType "NodePort" }}
   nodeKey: {{ .Values.nodePortSelector.key | quote }}
--- a/helm/ako/templates/configmap.yaml
+++ b/helm/ako/templates/configmap.yaml
@@ -26,9 +26,8 @@ data:
   {{ if .Values.AKOSettings.syncNamespace  }}
   syncNamespace: {{ .Values.AKOSettings.syncNamespace | quote }}
   {{ end }}
-  {{ if .Values.AKOSettings.namespaceSyncLabel  }}
-  namespaceSyncLabel: {{ .Values.AKOSettings.namespaceSyncLabel | quote }}
-  {{ end }}
+  nsSyncLabelKey: {{ .Values.AKOSettings.nameSpaceSelector.labelKey | quote }}
+  nsSyncLabelValue: {{ .Values.AKOSettings.nameSpaceSelector.labelValue | quote }}
   serviceType:  {{ .Values.L7Settings.serviceType | quote }}
   {{ if eq .Values.L7Settings.serviceType "NodePort" }}
   nodeKey: {{ .Values.nodePortSelector.key | quote }}
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -17,6 +17,10 @@ AKOSettings:
   disableStaticRouteSync: "false" # If the POD networks are reachable from the Avi SE, set this knob to true.
   clusterName: "my-cluster" # A unique identifier for the kubernetes cluster, that helps distinguish the objects for this cluster in the avi controller. // MUST-EDIT
   cniPlugin: "" # Set the string if your CNI is calico or openshift. enum: calico|canal|flannel|openshift 
+  #Label for namespacemigration in format key=value 
+  #eg namespaceSyncLabel: "app=migrate"
+  #Same label has to be present on namespace/s which needs migration/sync to AKO
+  namespaceSyncLabel: "app=migrate"
 
 ### This section outlines the network settings for virtualservices. 
 NetworkSettings:
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -17,10 +17,11 @@ AKOSettings:
   disableStaticRouteSync: "false" # If the POD networks are reachable from the Avi SE, set this knob to true.
   clusterName: "my-cluster" # A unique identifier for the kubernetes cluster, that helps distinguish the objects for this cluster in the avi controller. // MUST-EDIT
   cniPlugin: "" # Set the string if your CNI is calico or openshift. enum: calico|canal|flannel|openshift 
-  #Label for namespacemigration in format key=value 
-  #eg namespaceSyncLabel: "app=migrate"
+  #NamespaceSelector contains label key and value used for namespacemigration
   #Same label has to be present on namespace/s which needs migration/sync to AKO
-  namespaceSyncLabel: "app=migrate"
+  nameSpaceSelector:
+    labelKey: ""
+    labelValue: ""
 
 ### This section outlines the network settings for virtualservices. 
 NetworkSettings:
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -423,7 +423,7 @@ func (c *AviController) FullSyncK8s() error {
 				for _, ingObj := range ingObjs {
 					ingLabel := utils.ObjKey(ingObj)
 					ns := strings.Split(ingLabel, "/")
-					if utils.NSFilterFunction(ns[0], utils.GetGlobalK8NSObj(), nil, true) {
+					if utils.CheckIfNamespaceAccepted(ns[0], utils.GetGlobalNSFilter(), nil, true) {
 						key := utils.Ingress + "/" + ingLabel
 						utils.AviLog.Debugf("Dequeue for ingress key: %v", key)
 						nodes.DequeueIngestion(key, true)
@@ -591,7 +591,7 @@ func (c *AviController) InitializeNameSpaceSync() {
 		utils.AviLog.Debugf("Initializing NameSpace Sync. Received namespace label: %s = %s", nsLabelToSyncKey, nsLabelToSyncVal)
 		utils.InitializeNSSync(nsLabelToSyncKey, nsLabelToSyncVal)
 	}
-	nsFilterObj := utils.GetGlobalK8NSObj()
+	nsFilterObj := utils.GetGlobalNSFilter()
 	if !nsFilterObj.EnableMigration {
 		utils.AviLog.Info("Namespace Sync is disabled.")
 		return
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -29,11 +29,12 @@ import (
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/objects"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/rest"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/retry"
-
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+
+	ingruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -370,6 +371,9 @@ func (c *AviController) FullSyncK8s() error {
 			utils.AviLog.Warnf("Timed out while waiting for rest layer to respond, moving on with bootup")
 		}
 	}
+	//nsList := utils.GetAllNamespacesToSync(c.nsFilterObj)
+	nsList := utils.GetAllNamespacesToSync(utils.GetGlobalK8NSObj())
+	utils.AviLog.Debugf("In FullSynck8... Namespace list obtained is: [%v]", nsList)
 
 	svcObjs, err := utils.GetInformers().ServiceInformer.Lister().Services("").List(labels.Set(nil).AsSelector())
 	if err != nil {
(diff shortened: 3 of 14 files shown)


internal/k8s/ako_init.go
```diff
@@ -29,11 +29,12 @@ import (
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/objects"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/rest"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/retry"
-
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+
+	ingruntime "k8s.io/apimachinery/pkg/runtime"
```
sudswasavi, line 37: Why do you need a duplicate import? runtime is getting imported below?

internal/k8s/controller.go
```diff
@@ -153,9 +154,105 @@ func isRouteUpdated(oldRoute, newRoute *routev1.Route) bool {
 	return false
 }
 
+func isNamespaceUpdated(oldNS, newNS *corev1.Namespace) bool {
+	if oldNS.ResourceVersion == newNS.ResourceVersion {
+		return false
+	}
+	oldLabelHash := utils.Hash(utils.Stringify(oldNS.Labels))
```
sudswasavi, line 161: if I have labels like this:
 app:prod
and 
 prod:app

I presume, they will be two different hashes?

internal/k8s/controller.go
```diff
@@ -153,9 +154,105 @@ func isRouteUpdated(oldRoute, newRoute *routev1.Route) bool {
...
+					//Call ingress add
+					utils.AviLog.Debugf("Adding ingresses for namespaces:[%s]", nsCur.GetName())
+					utils.AddNamespace(nsCur.GetName(), nsFilterObj)
+					AddIngressFromNSToIngestionQueue(numWorkers, c, nsCur.GetName(), "ADD")
+				} else if nsOldFlag && !nsNewFlag {
+					//Case 2: Old valid namespace updated with invalid labels
+					//Call ingress delete
+					utils.AviLog.Debugf("Deleting ingresses for namespaces:[%s]", nsCur.GetName())
+					utils.DeleteNamespace(nsCur.GetName(), nsFilterObj)
+					AddIngressFromNSToIngestionQueue(numWorkers, c, nsCur.GetName(), "DELETE")
```
sudswasavi, line 240: Can we use constants for "ADD", "DELETE" wherever applicable?

sudswasavi (commented):
@monotosh-avi / @shubhamavi / @hemantavi / @parimanur  pls review.

internal/k8s/ako_init.go
```diff
@@ -570,3 +595,16 @@ func SyncFromNodesLayer(key string, wg *sync.WaitGroup) error {
...
 	return nil
 }
+
+//Controller Specific method
+func (c *AviController) InitializeNameSpaceSync() {
+	nsLabelToSync := lib.GetLabelToSyncNameSpace()
+	utils.AviLog.Debugf("Initializing NameSpace Sync. Received namespace label: [%s]", nsLabelToSync)
+	utils.InitializeNSSync(nsLabelToSync)
+	nsFilterObj := utils.GetGlobalK8NSObj()
+	if !nsFilterObj.EnableMigration {
```
monotosh-avi, line 605: Should we check if EnableMigration is true and then call InitializeNSSync () ?

pkg/utils/utils.go
```diff
@@ -311,3 +311,101 @@ func Remove(arr []string, item string) []string {
...
+	AviLog.Debugf("Is Namespace [%s] valid? -->[%v]", namespace, flag)
+	return flag
+}
+
+func InitializeNSSync(label string) {
+	AviLog.Debugf("In InitializeNSSync function, namespace label to check is: %s\n", label)
+	if label != "" {
+		segments := strings.Split(label, "=")
+		seg0 := strings.Trim(segments[0], " ")
+		seg1 := strings.Trim(segments[1], " ")
```
monotosh-avi, line 334: We should check the length of segments before accessing it's index

pkg/utils/utils.go
```diff
@@ -311,3 +311,101 @@ func Remove(arr []string, item string) []string {
...
+		key = obj.nsFilter.key
+	}
+	if obj.nsFilter.value != "" {
+		value = obj.nsFilter.value
+	}
+	return key, value
+}
+
+func AddNamespace(namespace string, obj *K8ValidNamespaces) {
+	obj.validNSList.lock.Lock()
```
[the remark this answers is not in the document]
akshayhavile: Over nsList ... reading and writing is going on from different points. like add, update namespace event handler, reading from BuildL7VSGraph etc... I think this will be in two different threads.. so added lock before writing.

helm/ako/values.yaml
```diff
@@ -17,6 +17,10 @@ AKOSettings:
   disableStaticRouteSync: "false" # If the POD networks are reachable from the Avi SE, set this knob to true.
   clusterName: "my-cluster" # A unique identifier for the kubernetes cluster, that helps distinguish the objects for this cluster in the avi controller. // MUST-EDIT
   cniPlugin: "" # Set the string if your CNI is calico or openshift. enum: calico|canal|flannel|openshift 
+  #Label for namespacemigration in format key=value 
+  #eg namespaceSyncLabel: "app=migrate"
+  #Same label has to be present on namespace/s which needs migration/sync to AKO
+  namespaceSyncLabel: "app=migrate"
```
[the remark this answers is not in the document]
hemantavi: May be we can break this into two keys, `namespaceSelectorKey: app` and `namespaceSelectorValue: value`. Or, these are some of the better ways to do this:
```
namespaceSelector:
    app: migrate
```
or,
```
namespaceSelector: |
    label.key: app
    label.value: migrate
```
Please check this: https://kubernetes.io/docs/concepts/configuration/configmap/#configmaps-and-pods

@akshayhavile @sudswasavi

internal/k8s/controller.go
```diff
@@ -153,9 +154,105 @@ func isRouteUpdated(oldRoute, newRoute *routev1.Route) bool {
...
+	newLabelHash := utils.Hash(utils.Stringify(newNS.Labels))
+	if oldLabelHash != newLabelHash {
+		return true
+	}
+	return false
+}
+func AddIngressFromNSToIngestionQueue(numWorkers uint32, c *AviController, namespace string, msg string) {
+	ingObjs, err := utils.GetInformers().IngressInformer.Lister().ByNamespace(namespace).List(labels.Set(nil).AsSelector())
+	if err != nil {
+		utils.AviLog.Errorf("NS to ingress queue add: Error occured while retrieving ingresss for namespace:[%s]", namespace)
```
hemantavi, line 171: If we return from here, then we can remove the `else` below.

hemantavi (changes requested):

helm/ako/templates/configmap.yaml
```diff
@@ -25,6 +25,9 @@ data:
   {{ if .Values.AKOSettings.syncNamespace  }}
   syncNamespace: {{ .Values.AKOSettings.syncNamespace | quote }}
   {{ end }}
+  {{ if .Values.AKOSettings.namespaceSyncLabel  }}
+  namespaceSyncLabel: {{ .Values.AKOSettings.namespaceSyncLabel | quote }}
```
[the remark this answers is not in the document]
akshayhavile: @hemantavi : I will updated above label as namespaceSelector. Regarding format of value in values.yaml.. I feel we should use 
namespaceSelector:
label-key: ""
label-value: ""

Above same format, we have used for nodeport selector.. we should be uniform in our approach.  Only thing with this is we need to make documentation bit mode explicit about how these values should be reflected while doing labeling to namespace, otherwise customer will be confused.

@sudswasavi  what is your thoughts on this?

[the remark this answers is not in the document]
sudswasavi: yeah makes sense.

pkg/utils/utils.go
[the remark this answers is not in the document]
sudswasavi: Okay if you are just adding things from one thread, maybe you just need a READ lock. You currently have a R/W lock.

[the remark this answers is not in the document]
akshayhavile: As we are updating nsList here, we added write lock here.

internal/k8s/controller.go
akshayhavile: Done

internal/k8s/controller.go
akshayhavile: it wll be different

internal/k8s/ako_init.go
akshayhavile: Taken care. We are now checking labels are there or not. If present, only then calling intializeNSSync to initialise global structure.

akshayhavile:
Addressed review comments.

helm/ako/templates/configmap.yaml
```diff
@@ -26,6 +26,8 @@ data:
   {{ if .Values.AKOSettings.syncNamespace  }}
   syncNamespace: {{ .Values.AKOSettings.syncNamespace | quote }}
   {{ end }}
+  nsSyncLabelKey: {{ .Values.AKOSettings.nameSpaceSelector.labelKey | quote }}
+  nsSyncLabelValue: {{ .Values.AKOSettings.nameSpaceSelector.labelValue | quote }}
```
hemantavi, line 30: May be I wasn't clear about the label configuration. But can we make it a bit structured here, like:
nsSyncLabel: |
   key:
   value:

pkg/utils/utils.go
```diff
@@ -311,3 +311,81 @@ func Remove(arr []string, item string) []string {
...
+	AviLog.Debugf("Namespace [%s] is valid [true/false]: [%v]", namespace, flag)
+	return flag
+}
+
+func InitializeNSSync(label string) {
+	segments := strings.Split(label, "=")
+
+	GlobalK8NSObj.EnableMigration = true
+	GlobalK8NSObj.nsFilter.key = segments[0]
+	GlobalK8NSObj.nsFilter.value = segments[1]
```
monotosh-avi, line 334: This would fail if length of segments is less than 1

pkg/utils/utils.go
```diff
@@ -311,3 +311,81 @@ func Remove(arr []string, item string) []string {
...
+func IsNSPresent(namespace string, obj *K8ValidNamespaces) bool {
+	obj.validNSList.lock.RLock()
+	defer obj.validNSList.lock.RUnlock()
+	_, flag := obj.validNSList.nsList[namespace]
+	AviLog.Debugf("Namespace [%s] is valid [true/false]: [%v]", namespace, flag)
+	return flag
+}
+
+func InitializeNSSync(label string) {
+	segments := strings.Split(label, "=")
```
hemantavi, line 330: We should do a length check here.

akshayhavile:
Addressed review comments.

internal/k8s/controller.go
```diff
@@ -153,9 +154,105 @@ func isRouteUpdated(oldRoute, newRoute *routev1.Route) bool {
...
+func AddNameSpaceEventHandler(numWorkers uint32, c *AviController) cache.ResourceEventHandler {
+	namespaceEventHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if c.DisableSync {
+				return
+			}
+			ns := obj.(*corev1.Namespace)
+			nsLabels := ns.GetLabels()
+			nsFilterObj := utils.GetGlobalK8NSObj()
+			if utils.NSFilterFunction(ns.GetName(), nsFilterObj, nsLabels, false) {
```
hemantavi, line 205: Can we also rename this function `NSFilterFunction` to possibly `ApplyNSFilter` or `CheckIfNamespaceAccepted`?

internal/k8s/controller.go
```diff
@@ -153,9 +154,105 @@ func isRouteUpdated(oldRoute, newRoute *routev1.Route) bool {
...
+				utils.AddNamespace(ns.GetName(), nsFilterObj)
+				utils.AviLog.Debugf("NS Add event: Namespace passed filter: %s", ns.GetName())
+			} else {
+				//Case: previoulsy deleted valid NS, added back with no labels or invalid labels but nsList contain that ns
+				utils.AviLog.Debugf("NS Add event: Namespace did not pass filter: %s", ns.GetName())
+				if utils.NSFilterFunction(ns.GetName(), nsFilterObj, nil, true) {
+					utils.AviLog.Debugf("Ns Add event: Deleting previous valid namespace: %s from valid NS List", ns.GetName())
+					utils.DeleteNamespace(ns.GetName(), nsFilterObj)
```
hemantavi, line 213: If this deletes the namespace from the filter, then let's rename it to `DeleteNamespaceFromFilter`. What do you think?

internal/k8s/controller.go
```diff
@@ -153,9 +154,105 @@ func isRouteUpdated(oldRoute, newRoute *routev1.Route) bool {
...
+				}
+			}
+
+		},
+		UpdateFunc: func(old, cur interface{}) {
+			if c.DisableSync {
+				return
+			}
+			nsOld := old.(*corev1.Namespace)
+			nsCur := cur.(*corev1.Namespace)
```
hemantavi, line 223: Let's add some checks about the object casting here?
```
nsOld, ok := old.(*corev1.Namespace)
if !ok { // handle error }
```

pkg/utils/utils.go
```diff
@@ -311,3 +311,79 @@ func Remove(arr []string, item string) []string {
 	}
 	return arr
 }
+
+var GlobalK8NSObj *K8ValidNamespaces = &K8ValidNamespaces{}
```
hemantavi, line 315: We are anyway using a getter for this variable. So, maybe we should make this variable private, if you don't want it to be manipulated directly?

internal/k8s/controller.go
```diff
@@ -153,9 +154,105 @@ func isRouteUpdated(oldRoute, newRoute *routev1.Route) bool {
...
+			nsFilterObj := utils.GetGlobalK8NSObj()
+			if isNamespaceUpdated(nsOld, nsCur) {
+				nsOldFlag := utils.NSFilterFunction(nsOld.GetName(), nsFilterObj, nsOld.Labels, false)
+				nsNewFlag := utils.NSFilterFunction(nsCur.GetName(), nsFilterObj, nsCur.Labels, false)
```
hemantavi, line 227: Maybe change the above variables to `oldNSAccepted` and `newNSAccepted`?

internal/lib/constants.go
```diff
@@ -95,6 +95,10 @@ const (
 	DefaultSEGroup                             = "Default-Group"
 	GatewayFinalizer                           = "gateway.ako.vmware.com"
 	ClusterStatusCacheKey                      = "cluster-runtime"
+
+	//Specifies command used in namespace event handler
+	Add    = "ADD"
+	Delete = "DELETE"
```
hemantavi, line 101: Can we rename the above constants to something which tells us what it does? For e.g.,
```
NsFilterAdd = "ADD"
NsFilterDelete = "DELETE"
```

hemantavi (commented):
LGTM @akshayhavile ! Just a few minor nits and comments.

akshayhavile:
Revision 2: Addressed review comments.

hemantavi (approved):

monotosh-avi (approved):

PR #186, opened 2020-11-09
Be tolerant to ingresses without backends

sudswasavi:
This commit makes AKO tolerant to ingresses without backends.
We simply return and not process.

Added a test to verify this.

### DIFF
--- a/internal/nodes/validator.go
+++ b/internal/nodes/validator.go
@@ -56,18 +56,23 @@ func (v *Validator) IsValiddHostName(hostname string) bool {
 	return false
 }
 
-func validateSpecFromHostnameCache(key, ns, ingName string, ingSpec v1beta1.IngressSpec) {
+func validateSpecFromHostnameCache(key, ns, ingName string, ingSpec v1beta1.IngressSpec) bool {
 	nsIngress := ns + "/" + ingName
 	for _, rule := range ingSpec.Rules {
-		for _, svcPath := range rule.IngressRuleValue.HTTP.Paths {
-			found, val := SharedHostNameLister().GetHostPathStoreIngresses(rule.Host, svcPath.Path)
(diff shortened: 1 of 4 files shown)


internal/nodes/validator.go
```diff
@@ -56,18 +56,23 @@ func (v *Validator) IsValiddHostName(hostname string) bool {
...
-func validateSpecFromHostnameCache(key, ns, ingName string, ingSpec v1beta1.IngressSpec) {
+func validateSpecFromHostnameCache(key, ns, ingName string, ingSpec v1beta1.IngressSpec) bool {
 	nsIngress := ns + "/" + ingName
 	for _, rule := range ingSpec.Rules {
-		for _, svcPath := range rule.IngressRuleValue.HTTP.Paths {
-			found, val := SharedHostNameLister().GetHostPathStoreIngresses(rule.Host, svcPath.Path)
-			if found && len(val) > 0 && utils.HasElem(val, nsIngress) && len(val) > 1 {
-				// TODO: push in ako apiserver
-				utils.AviLog.Warnf("key: %s, msg: Duplicate entries found for hostpath %s%s: %s in ingresses: %+v", key, nsIngress, rule.Host, svcPath.Path, utils.Stringify(val))
+		if rule.IngressRuleValue.HTTP != nil {
```
parimanur, line 62: Nit: return false if rule.IngressRuleValue.HTTP == nil ? else can be avoided.

parimanur (approved):

internal/nodes/validator.go
sudswasavi: Thought I am doing that in line 71?

monotosh-avi (approved):

shubhamavi (approved):

PR #190, opened 2020-11-11
Set status in AKO statefulset on all avi object deletion

monotosh-avi:
Here are the set of actions takes by ako when deleteConfig is set to true:
- When deleteConfig is set to true, a condition in AKO status is set to - {“message":"objDeletionStarted",}
- After all avi objects are deleted, a condition in AKO status is set to - {“message":"objDeletionDone"}


internal/rest/dequeue_nodes.go
```diff
@@ -407,8 +407,29 @@ func (rest *RestOperations) deleteVSOper(vsKey avicache.NamespaceName, vs_cache_
 		rest_ops = rest.L4PolicyDelete(vs_cache_obj.L4PolicyCollection, namespace, rest_ops, key)
 		rest_ops = rest.PoolGroupDelete(vs_cache_obj.PGKeyCollection, namespace, rest_ops, key)
 		rest_ops = rest.PoolDelete(vs_cache_obj.PoolKeyCollection, namespace, rest_ops, key)
-		return rest.ExecuteRestAndPopulateCache(rest_ops, vsKey, nil, key)
+		retVal := rest.ExecuteRestAndPopulateCache(rest_ops, vsKey, nil, key)
```
parimanur, line 410: nit: rename retVal -> success?

internal/status/configmap_finaliser.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	currConfig.SetFinalizers([]string{})
+	UpdateConfigmapFinalizer(currConfig, []string{})
+	utils.AviLog.Infof("Removed the finalizer %s from avi CM", lib.ConfigmapFinalizer)
+}
+
+// SetConfigmapFinalizer : update from configmap with ako finaliser.
+// After this the configmap cannot be deleted by the user without clearing the finaliser
+func AddConfigmapFinalizer() {
+	currConfig, err := utils.GetInformers().ConfigMapInformer.Lister().ConfigMaps("avi-system").Get("avi-k8s-config")
```
parimanur, line 28: use AviNS constant.

internal/rest/dequeue_nodes.go
```diff
@@ -407,8 +407,29 @@ func (rest *RestOperations) deleteVSOper(vsKey avicache.NamespaceName, vs_cache_
...
+		if retVal == true {
+			vsKeysPending := rest.cache.VsCacheMeta.AviGetAllKeys()
+			utils.AviLog.Infof("key: %s, msg: Number of VS deletion pending: %d", key, len(vsKeysPending))
+			if len(vsKeysPending) == 0 {
+				// All VSes got deleted, done with deleteConfig operation. Now notify the user
+				if lib.ConfigDeleteSyncChan != nil {
+					utils.AviLog.Debugf("key: %s, msg: sending signal for vs deletion notification", key)
+					close(lib.ConfigDeleteSyncChan)
+					lib.ConfigDeleteSyncChan = nil
```
parimanur, line 419: Can we close the chan at single place? Because it can effectively enter into L#: 427 and close already closed chan.

monotosh-avi: We would not reach that, we are returning at line 423

internal/rest/dequeue_nodes.go
monotosh-avi: I'll update

internal/status/configmap_finaliser.go
monotosh-avi: Makes sense, I'll use a constant

sudswasavi (commented):
Broadly looks good.
One problem in a TKGm flow with just finalizer without ownerref. They might issue a delete call for all kubernetes resources. The statefulset would get deleted, but the configmap won't. It can then get into an infinite wait because the finalizer string from the configmap had to be removed by the SS and it's no longer present.

We could have two options: 
 - Make the SS a child of the CM.
 - Always create CM before SS.

go.sum
```diff
@@ -2,6 +2,14 @@ cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMT
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.38.0 h1:ROfEUZz+Gh5pa62DJWXSaonyu3StP6EA6lPEXPI6mCo=
 cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSRU=
+github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78/go.mod h1:LmzpDX56iTiv29bbRTIsUNlaFfuhWRQBWjQdVyAevI8=
```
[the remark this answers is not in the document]
monotosh-avi: These got added somehow automatically. I'll remove them.

internal/k8s/ako_init.go
```diff
@@ -542,6 +548,21 @@ func (c *AviController) DeleteModels() {
 		utils.AviLog.Infof("Deleting objects for model: %s", modelName)
 		sharedQueue.Workqueue[bkt].AddRateLimited(modelName)
 	}
+
+	// Wait for maximum 30 minutes for the sync to get completed
+	timeout := make(chan bool, 1)
+	go func() {
+		time.Sleep(lib.AviObjDeletionTime * time.Minute)
```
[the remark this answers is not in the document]
monotosh-avi: I had checked with system test team. It takes around 23 minutes with 3k ingresses (secure + insecure)

sudswasavi (commented):
One minor nit.

internal/rest/dequeue_nodes.go
```diff
@@ -407,8 +407,29 @@ func (rest *RestOperations) deleteVSOper(vsKey avicache.NamespaceName, vs_cache_
 		rest_ops = rest.L4PolicyDelete(vs_cache_obj.L4PolicyCollection, namespace, rest_ops, key)
 		rest_ops = rest.PoolGroupDelete(vs_cache_obj.PGKeyCollection, namespace, rest_ops, key)
 		rest_ops = rest.PoolDelete(vs_cache_obj.PoolKeyCollection, namespace, rest_ops, key)
-		return rest.ExecuteRestAndPopulateCache(rest_ops, vsKey, nil, key)
+		success := rest.ExecuteRestAndPopulateCache(rest_ops, vsKey, nil, key)
+		if success == true {
```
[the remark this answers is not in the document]
monotosh-avi: Didn't get this .. do you think we need to check something more ?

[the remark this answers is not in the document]
sudswasavi: "If success" is equivalent to "if success == true"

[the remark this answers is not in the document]
monotosh-avi: Oh yes, that would work. Updated.

parimanur (approved):

internal/status/statefulset_status.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		utils.AviLog.Warnf("Error in updating ako statefulset: %v", err)
+		return
+	}
+	utils.AviLog.Debugf("Successfully reset ako statefulset: %v", u)
+}
+
+// AddStatefulSetStatus sets a condition in status of AKO statefulset to the desired value
+func AddStatefulSetStatus(msg string) {
+	ss, err := utils.GetInformers().ClientSet.AppsV1().StatefulSets(lib.AviNS).Get(lib.AKOStatefulSet, metav1.GetOptions{})
```
hemantavi, line 64: @sudswasavi @shubhamavi does AKO run in `avi-system` always? Or, in wcp, it can run in a different namespace too?

internal/k8s/ako_init.go
```diff
@@ -554,6 +562,21 @@ func (c *AviController) DeleteModels() {
...
+		timeout <- true
+	}()
+	lib.SetConfigDeleteSyncChan()
+	select {
+	case <-lib.ConfigDeleteSyncChan:
+		utils.AviLog.Infof("Processing done for deleteConfig, user would be notified through statefulset update")
+	case <-timeout:
+		utils.AviLog.Warnf("Timed out while waiting for rest layer to respond for delete config")
+	}
+	status.AddStatefulSetStatus(lib.ObjectDeletionDoneStatus)
```
hemantavi, line 579: Can we pls check for the timeout here and then update the status as `objDeletionTimeout`?

hemantavi (commented):
LGTM with a couple of comments.

sudswasavi (approved):

PR #192, opened 2020-11-12
static IP support for service LB using spec.loadbalancerIP

shubhamavi:
This introduces staticIP support for k8s services objects using the `spec.loadBalancerIP` field
The functionality is similar to how AKO handles static IPs in gateways, wherein updates on the corresponding vsvip objects is not allowed. In order to update the static IP, the Service/Gateway object must be re-created using the new preferred IP.
Errors associated to preferred IP updates are reported to the gateway status conditions, but is not currently possible in case of service objects, although AKO logs the Avi controller API response error.


sudswasavi (approved):
LGTM. @parimanur / @monotosh-avi can you guys pls take a look?

monotosh-avi (approved):

PR #193, opened 2020-11-12
Support Tenancy in AKO.

parimanur:
In AKO custom role,  below permissions are needed for admin tenant.
image

image

image

Custom role needs to be assigned for each cluster tenant.
image


sudswasavi:
@parimanur is it possible to not allow `admin` access to the individual AKOs? What permissions are granted if we add the `Tenant-admin` role. Does it only allow access to VRFs or more than that? cc @apalsule

parimanur:
> @parimanur is it possible to not allow `admin` access to the individual AKOs? What permissions are granted if we add the `Tenant-admin` role. Does it only allow access to VRFs or more than that? cc @apalsule

@sudswasavi Tenant-Admin has privilege only in his tenant. We can in-fact have another custom role ako-tenant which has specific Read/Write privilege for that particular tenant.

sudswasavi (commented):
Looks good to me overall.
@parimanur  looks like the tests only try the `admin` tenant. Can we try a few UTs for this change in the next PR?

sudswasavi:
@shubhamavi / @monotosh-avi PTAL and merge as appropriate.

sudswasavi (approved):

sudswasavi:
@parimanur what happens if a customer had VSes in the `admin` tenant and then configured AKO with a `foo` tenant?

parimanur:
> @parimanur what happens if a customer had VSes in the `admin` tenant and then configured AKO with a `foo` tenant?

The upgrade/switch will be disruptive, the flow would be delete all the config from older tenant and then to switch the tenant in AKO

sudswasavi:
> > @parimanur what happens if a customer had VSes in the `admin` tenant and then configured AKO with a `foo` tenant?
> 
> The upgrade/switch will be disruptive, the flow would be delete all the config from older tenant and then to switch the tenant in AKO

Can we handle this automatically in AKO though? Maybe in the next PR? What are your thoughts? @apalsule

monotosh-avi (approved):

apalsule:
> > > @parimanur what happens if a customer had VSes in the `admin` tenant and then configured AKO with a `foo` tenant?
> > 
> > 
> > The upgrade/switch will be disruptive, the flow would be delete all the config from older tenant and then to switch the tenant in AKO
> 
> Can we handle this automatically in AKO though? Maybe in the next PR? What are your thoughts? @apalsule

@sudswasavi Seems difficult to handle automatically, will tenant change involve a reboot of AKO ? We may not even have the old tenant context and permissions to delete the older VSes.

PR #195, opened 2020-11-14
shutdown ako on akoNS/avi-secret update or delete

shubhamavi:


shubhamavi:
failed due to docker rate limit, will trigger the pipeline later

shubhamavi:
build ako

monotosh-avi (commented):
@shubhamavi  We don't want secret informers for open shift for now. Can we retain that ? 
Openshift has too many secrets by default, secret informers would take too much resources in open shift. 

Looks good to me otherwise

internal/k8s/controller.go
```diff
@@ -475,54 +465,68 @@ func (c *AviController) SetupEventHandlers(k8sinfo K8sinformers) {
...
+					utils.AviLog.Errorf("Tombstone contained object that is not a Secret: %#v", obj)
 					return
 				}
 			}
-			namespace, _, _ := cache.SplitMetaNamespaceKey(utils.ObjKey(ingress))
-			key := utils.Ingress + "/" + utils.ObjKey(ingress)
-			bkt := utils.Bkt(namespace, numWorkers)
-			c.workqueue[bkt].AddRateLimited(key)
-			utils.AviLog.Debugf("key: %s, msg: DELETE", key)
+			if validateAviSecret(secret) {
```
monotosh-avi, line 481: Why if we delete an secret referred by an ingress, which is not in avi NS ?

monotosh-avi (commented):
I think we need  to check if we can only sync secret from avi NS for openshift and from all Namespaces in non Openshift environment.

shubhamavi:
@sudswasavi the code to initialise the informer for AKO specific ns is utils.go. checkout the akoNSInformerFactory code in utils.go

sudswasavi:
> @sudswasavi the code to initialise the informer for AKO specific ns is utils.go. checkout the akoNSInformerFactory code in utils.go

Thanks. I missed seeing that. This will address @monotosh-avi 's concern as well I think.

monotosh-avi:
@shubhamavi do you think this condition should be extended to open shift also, or is it fine ?
`
			if adv_l4 {
 				informers.SecretInformer = akoNSInformerFactory.Core().V1().Secrets()
 			} else {
 				informers.SecretInformer = kubeInformerFactory.Core().V1().Secrets()
 			}
`
@sudswasavi

sudswasavi:
> @shubhamavi do you think this condition should be extended to open shift also, or is it fine ?
> `if adv_l4 { informers.SecretInformer = akoNSInformerFactory.Core().V1().Secrets() } else { informers.SecretInformer = kubeInformerFactory.Core().V1().Secrets() }`
> @sudswasavi

I think we should do it for openshift as well in this PR for completeness sake. Let's not init the secrets for openshift.

monotosh-avi (approved):

sudswasavi (approved):

sudswasavi (approved):

PR #196, opened 2020-11-16
Changes to dockerfile and makefile

sudswasavi:
These changes are being done due to the new docker rate limit
imposition on docker pulls. For internal CI frameworks, we would
like to use internal registry of base images for builds.


monotosh-avi (approved):

shubhamavi (approved):
needs cherry-pick

PR #179, opened 2020-11-02
Support for networking/v1 Ingress 

shubhamavi:
This PR:

 - Updates the k8s.io dependencies: api, apimachinery, client-go to 0.19.3
This introduces syntactical changes in the way kubernetes Client calls are made. All RESTClient calls (Create, Update, Delete, Patch, List...) now take additional `context` and query options. The updates corresponding to this have been done in AKO and the UTs
 - Uses the networking/v1beta1/Ingress consistently across AKO. Since kubernetes 1.19 still supports the networking/v1beta Ingress, and the 0.19.3 clients have the same struct in networking/v1 and v1beta1, AKO now ingests to networking/v1 Ingress from kubernetes 1.19 as well. The UTs have been moved from extensions to networking/v1beta1
 - Deprecates Extensions/v1beta1 Ingress
 - Introduces support for networking/v1/IngressClass. adds cache mappings for Ingress and IngressClass (just like the existing Gateway/GatewayClass). AKO retains the class functionality (via AKO) for kubernetes clusters that do not support IngressClass yet. For the ones that do, AKO helm installation/upgrade flow would automatically add an AKO specific IngressClass written here `helm/ako/templates/ingressclass.yaml`
 - Support for ingress paths' pathType `Exact` and `Prefix`. `ImplementationSpecific` pathType functionality for AKO would be like that of pathType `Prefix`
<br/>

- Moves away from client-sets/listers/informers generated in vmware-tanzu/service-apis and adds scripts to generate the same from AKO itself. The generated clients are kept in `third_party/service-apis` directory. The script used to generate these client sets from types in vmware-tanzu resides in `hack/update-codegen-service-apis.sh`
- Updates generated code for hostrule/httprule clients to be compatible with k8s.io 0.19.3 dependencies. The script to do this will be added in a subsequent PR (with v1alpha2 support)
<br/>

- fixes issues in Makefile to run the UTs inside golang based docker containers

PR #197, opened 2020-11-16
Add an operator for deploying and managing AKO

hemantavi:
This operator takes the AKO installation/deployment configurations from
 a new CRD called `AKOConfig`.

Create/Update of an `AKOConfig`:
    - We look for the `AKOConfig` object. Once we get an object, the
      reconcile method is called. In this method, we first generate all the
      artifacts viz. Configmap, Statefulset, PodSecurityPolicy,
      ServiceAccount, Clusterrole and a Clusterrolebinding. And then, we
      create them on the cluster.
    - An update event on the `AKOConfig` object triggers the same reconcile
      method. This flow re-writes all the existing artifacts. Note that some
      changes to the configmap require a reboot. So, we do a checksum
      calculation of the configmap data fields and compare them with the
      existing one. If there's a checksum change, we delete the AKO
      statefulset if it exists and then create a new one.
    - For delete events, we use a finalizer in the `AKOConfig` object. If
      deleted, the reconcile method is first called, where we delete all the
      artifactories and then remove the finalizer from the `AKOConfig` object.


hemantavi:
Based off of the client-go changes https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/179 .

TODOs:
- [ ] SecurityContext in Statefulset.
- [ ] SyncNamespace in Configmap.
- [ ] Watch on Configmap resource to update it back to the desired state.
- [ ] Update current state of deployments in the status of the `AKOConfig` object.

sudswasavi (commented):
I was in the middle of the review but looks like there was force push made. Hence publishing what I already had.

ako-operator/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Build the manager binary
```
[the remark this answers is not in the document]
hemantavi: Not sure if we need copyright on the Dockerfile as well. The existing one for `Dockerfile.ako` doesn't have it. Have added it to the ".go" files. Let me know if we need it, and I will add it.

ako-operator/README.md
```diff
@@ -0,0 +1,41 @@
+## An operator for AKO
+
+This operator takes care of deploying, managing and removing AKO from openshift/kubernetes clusters. It takes the AKO installation/deployment configuration from a CRD called `AKOConfig`.
+
+### Pre-reqs before deploying the operator
+- CRD for `AKOConfig` must be installed.
+- Asecret called `avi-secret` must be installed. This contains the username
```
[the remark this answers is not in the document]
hemantavi: Yep, makes sense. Will add a helm chart in the next PR. With that, I will also add instructions on how to do helm install.

ako-operator/README.md
```diff
@@ -0,0 +1,41 @@
...
+  and password for the Avi controller (Currently, not enforced).
+- And, CRD definitions for `HostRule` and `HttpRule` must be installed
+  (Currently, not enforced).
+
+Run `make install` to install the above definitions.
+
+### Installing the operator
+#### Out of cluster execution
+To run the operator outside of a cluster, firs build the binary:
```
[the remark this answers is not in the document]
hemantavi: Fixed.

ako-operator/controllers/akoconfig_controller.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (r *AKOConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&akov1alpha1.AKOConfig{}).
+		Owns(&corev1.ConfigMap{}).
+		Owns(&appsv1.StatefulSet{}).
+		Complete(r)
+}
+
+func (r *AKOConfigReconciler) ReconcileConfigmap(ctx context.Context, ako akov1alpha1.AKOConfig, log logr.Logger) error {
```
monotosh-avi, line 195: Instead of using this function, can we call createOrUpdateConfigMap directly ? ReconcileConfigmap doesn't seem to be doing anything else. 
Similar for other functions bellow.

ako-operator/controllers/crb.go
```diff
@@ -0,0 +1,91 @@
...
+				Kind:      "ServiceAccount",
+				Name:      AKOServiceAccount,
+				Namespace: AviSystemNS,
+			},
+		},
+	}
+
+	err := ctrl.SetControllerReference(&ako, &crb, r.Scheme)
+	if err != nil {
+		log.Error(err, "error in setting controller reference, clusterrolebinding changes would be ignored")
```
monotosh-avi, line 88: Is this long msg correct ? In createOrUpdateClusterroleBinding() we would always update the crb even if ctrl.SetControllerReference fails.

ako-operator/controllers/crb.go
```diff
@@ -0,0 +1,91 @@
...
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      AKOServiceAccount,
+				Namespace: AviSystemNS,
+			},
+		},
+	}
+
+	err := ctrl.SetControllerReference(&ako, &crb, r.Scheme)
```
monotosh-avi, line 86: Could not find the definition of SetControllerReference() .. probably I missed something.

monotosh-avi (commented):
Looks good to me. Few minor comments.

ako-operator/controllers/akoconfig_controller.go
```diff
@@ -0,0 +1,236 @@
...
+func (r *AKOConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	log := r.Log.WithValues("ako-operator", req.NamespacedName)
+
+	var ako akov1alpha1.AKOConfig
+	err := r.Client.Get(ctx, req.NamespacedName, &ako)
+	if err != nil && errors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	} else if err != nil {
+		log.Info("unable to fetch AKOConfig object", "err", err)
```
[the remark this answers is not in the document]
hemantavi: The reason I set it to info is because, error prints a traceback. And this condition might happen intermittently, so it leaves unnecessary tracebacks in the logs.

ako-operator/controllers/akoconfig_controller.go
hemantavi: Sure, yeah. Makes sense. Will do that.

ako-operator/controllers/configmap.go
```diff
@@ -0,0 +1,196 @@
...
+	log.V(1).Info("building a new configmap for AKO")
+
+	var oldCM corev1.ConfigMap
+
+	if err := r.Get(ctx, getConfigMapName(), &oldCM); err != nil {
+		log.V(0).Info("error getting a configmap with name", "name", ConfigMapName, "err", err)
+	} else {
+		log.V(1).Info("old configmap", "old cm", oldCM)
+		if oldCM.ObjectMeta.GetName() != "" {
+			log.V(0).Info("a configmap with the same name already exists, it will be updated", "name",
```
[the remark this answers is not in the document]
hemantavi: It is, the plan is to convert all log.Info()s to log.V(x).Info(). Will do that.

ako-operator/controllers/crb.go
hemantavi: Yes, but we aren't watching for the clusterrolebinding changes. Meaning, if someone updates the clusterrolebinding out-of-band, then we aren't watching for those changes.

ako-operator/controllers/akoconfig_controller.go
```diff
@@ -0,0 +1,236 @@
...
+// +kubebuilder:rbac:groups="rbac",resources=clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="policy",resources=podsecuritypolicies,verbs=get;list;watch;create;update;patch;delete
+
+func (r *AKOConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	log := r.Log.WithValues("ako-operator", req.NamespacedName)
+
+	var ako akov1alpha1.AKOConfig
+	err := r.Client.Get(ctx, req.NamespacedName, &ako)
+	if err != nil && errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
hemantavi: We handle object deletion via the finalizers. If the root object is deleted without a finalizer, we won't handle the dependent objects. A finalizer is mandatory for taking care of the dependent resources.

ako-operator/controllers/crb.go
hemantavi: This is defined in the runtime package: sigs.k8s.io/controller-runtime/pkg/controller/controllerutil

shubhamavi:
build ako

ako-operator/controllers/crb.go
monotosh-avi: Got it, I thought you have defined this.

hemantavi:
build ako

monotosh-avi (dismissed):

ako-operator/controllers/akoconfig_controller.go
```diff
@@ -0,0 +1,193 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	akov1alpha1 "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/ako-operator/api/v1alpha1"
+)
+
+const (
+	CleanupFinalizer = "ako.vmware.com/finalizer"
```
maplain, line 37: should we be explicit about its role her? like having `cleanup` in the string.

hemantavi: Sure, if you mean something like: `ako.vmware.com/cleanup`.

Makefile
```diff
@@ -52,8 +53,44 @@ ifdef PHOTON_SRC_REPO
...
 endif
+
+ifdef UBI_SRC_REPO
+	$(eval BUILD_ARG_UBI=--build-arg ubi_src_repo=$(UBI_SRC_REPO))
+else
+	$(eval BUILD_ARG_UBI=)
+endif
 	sudo docker build -t $(BINARY_NAME_AKO):latest --label "BUILD_TAG=$(BUILD_TAG)" --label "BUILD_TIME=$(BUILD_TIME)" $(BUILD_ARG_GOLANG) $(BUILD_ARG_PHOTON) -f Dockerfile.ako .
 
+.PHONY: ako-operator-docker
```
shubhamavi, line 64: can we initialise these vars outside the target, globally, since they seem to be getting used for both ako-operator-docker and build.

hemantavi:
build ako

shubhamavi (approved):
minor comments. also please remove the ako-operator/main binary. I wasn't able to  comment on that inline. Thanks

shubhamavi (commented):
I think that going into ObjectMeta to access the functions might not be required. Could you check if we can directly call those functions.
for instance: `cm.GetName()` instead of `cm.ObjectMeta.GetName()`  etc.

shubhamavi (approved):

hemantavi:
build ako

hemantavi:
build ako

monotosh-avi (approved):

ako-operator/controllers/role.go
```diff
@@ -0,0 +1,129 @@
+package controllers
```
sudswasavi, line 1: this should be below the copyright

ako-operator/controllers/statefulset.go
```diff
@@ -0,0 +1,253 @@
...
+				Ports: []v1.ContainerPort{
+					{
+						Name:          "http",
+						ContainerPort: 80,
+						Protocol:      "TCP",
+					},
+				},
+				Resources: resources,
+				LivenessProbe: &v1.Probe{
+					InitialDelaySeconds: 5,
```
sudswasavi, line 236: How does someone change these values with the operator mode?

ako-operator/controllers/sa.go
```diff
@@ -0,0 +1,50 @@
+package controllers
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
```
sudswasavi, line 6: copyright missing.

sudswasavi (commented):
Can you pls ensure that all files have copyrights? Few minor nits and queries inline.

ako-operator/controllers/role.go
hemantavi: Moved, thanks.

ako-operator/controllers/sa.go
hemantavi: Added.

ako-operator/controllers/statefulset.go
hemantavi: As discussed, they can change it in the statefulset directly and the values will be retained till a user makes a change in the `AKOConfig` properties forcing the operator to restart the statefulset.

monotosh-avi (dismissed):

monotosh-avi (approved):

hemantavi:
build ako

hemantavi:
build ako

hemantavi:
build ako

shubhamavi (approved):

PR #200, opened 2020-11-18
Integration Test Cases for syncing k8s namespaced ingresses

akshayhavile:
Integration tests for Namespace wise migration feature

--- Testing---
Ran successfully:
make int_test

### DIFF
--- a/tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
+++ b/tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
@@ -1,0 +1,368 @@
+/*
+ * Copyright 2019-2020 VMware, Inc.
+ * All Rights Reserved.
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*   http://www.apache.org/licenses/LICENSE-2.0
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package namespacemigrationtests
+
+import (
+	"fmt"
+	"os"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/tests/integrationtest"
+
+	"github.com/onsi/gomega"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/cache"
+	crdfake "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/client/clientset/versioned/fake"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/k8s"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
+	avinodes "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/nodes"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/objects"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+)
+
+var KubeClient *k8sfake.Clientset
+var CRDClient *crdfake.Clientset
+var ctrl *k8s.AviController
+
+func TestMain(m *testing.M) {
+	os.Setenv("INGRESS_API", "extensionv1")
+	os.Setenv("NETWORK_NAME", "net123")
+	os.Setenv("CLUSTER_NAME", "cluster")
+	os.Setenv("CLOUD_NAME", "CLOUD_VCENTER")
+	os.Setenv("SEG_NAME", "Default-Group")
+	os.Setenv("NODE_NETWORK_LIST", `[{"networkName":"net123","cidrs":["10.79.168.0/22"]}]`)
+	os.Setenv("SERVICE_TYPE", "ClusterIP")
+
+	KubeClient = k8sfake.NewSimpleClientset()
+	CRDClient = crdfake.NewSimpleClientset()
+	lib.SetCRDClientset(CRDClient)
+
+	registeredInformers := []string{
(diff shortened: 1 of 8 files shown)


akshayhavile:
build ako

akshayhavile:
build ako

tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
```diff
@@ -0,0 +1,368 @@
...
+	os.Setenv("INGRESS_API", "extensionv1")
+	os.Setenv("NETWORK_NAME", "net123")
+	os.Setenv("CLUSTER_NAME", "cluster")
+	os.Setenv("CLOUD_NAME", "CLOUD_VCENTER")
+	os.Setenv("SEG_NAME", "Default-Group")
+	os.Setenv("NODE_NETWORK_LIST", `[{"networkName":"net123","cidrs":["10.79.168.0/22"]}]`)
+	os.Setenv("SERVICE_TYPE", "ClusterIP")
+
+	KubeClient = k8sfake.NewSimpleClientset()
+	CRDClient = crdfake.NewSimpleClientset()
```
hemantavi, line 53: Do we need a `CRDClient`?

tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
```diff
@@ -0,0 +1,368 @@
...
+	wgIngestion := &sync.WaitGroup{}
+	waitGroupMap["ingestion"] = wgIngestion
+	wgFastRetry := &sync.WaitGroup{}
+	waitGroupMap["fastretry"] = wgFastRetry
+	wgSlowRetry := &sync.WaitGroup{}
+	waitGroupMap["slowretry"] = wgSlowRetry
+	wgGraph := &sync.WaitGroup{}
+	waitGroupMap["graph"] = wgGraph
+	AddConfigMap()
+	ctrl.HandleConfigMap(informers, ctrlCh, stopCh, quickSyncCh)
```
hemantavi, line 94: `HandleConfigMap` returns an error, do you think we should have a fatalf on that error?

tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
```diff
@@ -0,0 +1,368 @@
...
+
+	SetUpNameSpaceMigration("app", "migrate", "namespace")
```
hemantavi, line 96: These helpers are more suited for namespace syncing rather than namespace migration. Can we change the function names accordingly?

tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
```diff
@@ -0,0 +1,368 @@
...
+		_, aviModel := objects.SharedAviGraphLister().Get(modelName)
+		if aviModel != nil {
+			nodes := aviModel.(*avinodes.AviObjectGraph).GetAviVS()
+			return nodes[0].PoolRefs
+		}
+		return aviModel
+	}, 30*time.Second).Should(gomega.BeNil())
+}
+
+func TestNameSyncFeatureWithCorrectEnvParameters(t *testing.T) {
```
hemantavi, line 198: Should it be better to rename this to `TestNamespaceSyncWithCorrectEnvs`?

hemantavi (commented):
Looks good overall (have some minor comments), I guess a summary of tests would be nice to have in the commit message:
- Valid namespace selector
- Transition from valid to invalid namespace
- Transition from invalid to valid namespace
- Transition from invalid to invalid namespace
- Transition from valid to invalid namespace with hostname

Some doubts:
- Regarding tests cases related to changing the namespace selector in the configmap itself, would we add those in the FTs?
- These test cases are not related to migration, so would it better to rename them to namespace sync test cases?
- In one of the transition case from valid to invalid namespace, do we also check deletion of ingresses in layer 2?

Also, we should probably have one consitent naming scheme regarding `Namespace`.

akshayhavile:
@hemantavi : I will take care of comments provided.

Regarding doubts:
----
1. Regarding tests cases related to changing the namespace selector in the configmap itself, would we add those in the FTs?
>> Yes, we will add those test cases in FT

2. These test cases are not related to migration, so would it better to rename them to namespace sync test cases?
>> I will remain file and wherever I am mentioning migration to sync.

3. In one of the transition case from valid to invalid namespace, do we also check deletion of ingresses in layer 2?
>> I did not get this. When NS is transition from valid to invalid, ingress object is not deleted from k8. To remove that ingress from AVI side, we just delete pool associated with that Ingress. As that is shared ingress, that object will be still valid(not deleted) in layer 2. 

Also, we should probably have one consitent naming scheme regarding Namespace.
>> Can you please elaborate on this? I did not get this.

akshayhavile:
Addressed review comments.

tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
akshayhavile: @hemantavi .. Yes.. It is required ..if we take out, crash is observed at controller.go:819 .. while starting informers.

tests/namespacemigrationtests/l7_ingress_namespace_migration_test.go
akshayhavile: @hemantavi I think we can ignore it.. as it is not hampering our testing flow.

akshayhavile (commented):
Addressed all review comments.

monotosh-avi (commented):
Can we add the transition test cases for hostname shard also ?

akshayhavile:
@monotosh-avi : Added one transition test case(Valid to Invalid) for hostname based sharding scheme at end, to check pool deletion logic is kicked in or not. Rest of behaviour is same as that of namespace based sharding test cases. That was the reason of not adding remaining test cases (Invalid to Valid/ Invalid to Invalid).

hemantavi:
@akshayhavile 
> I did not get this. When NS is transition from valid to invalid, ingress object is not deleted from k8. To remove that ingress from AVI side, we just delete pool associated with that Ingress. As that is shared ingress, that object will be still valid(not deleted) in layer 2.

Yeah, I meant pool deletion from the layer 2 models. If we verify those, we should be fine.

> Can you please elaborate on this? I did not get this.

Regarding function and struct names which deal with namespaces, some places I see, `FakeNameSpace` and in other places, I see `AddNamespace()`. `Namespace` is not consistent. Can we make this consistent to just `Namespace`?

akshayhavile:
Addressed review comments

akshayhavile:
build ako

hemantavi (approved):

sudswasavi (approved):

sudswasavi:
Thanks @akshayhavile

PR #205, opened 2020-11-24
Adding UT for tenancy change in AKO.

parimanur:


monotosh-avi (approved):

tests/integrationtest/l4_l7_namespace_shard_nodeport_test.go
```diff
@@ -353,3 +353,63 @@ func TestNoHostIngressInNodePort(t *testing.T) {
 
 	TearDownTestForIngressInNodePortMode(t, model_Name)
 }
+
+func TestNoHostIngressInNodePortWithMultiTenantEnabled(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+
+	// enable tenants per cluster
+	SetAkoTenant()
+	defer ResetAkoTenant()
```
sudswasavi, line 362: Is the resetTenant really required? in a running AKO the tenant would be always set to the Ako tenant?

sudswasavi (approved):

tests/integrationtest/l4_l7_namespace_shard_nodeport_test.go
parimanur: Yes, Since other tests are relying on the env variable this ensures that other tests run with admin tenent only.

sudswasavi: Okay makes sense.

PR #208, opened 2020-11-25
Fix: <oc describe route routename > crash

akshayhavile:
Fixes "oc describe route routename" crash.

Added LastTransitionTime test condition to verify, time is getting published or not.

### DIFF
--- a/internal/status/route_status.go
+++ b/internal/status/route_status.go
@@ -311,7 +311,7 @@ func updateRouteObject(mRoute *routev1.Route, updateOption UpdateStatusOptions, 
 
 func compareRouteStatus(oldStatus, newStatus []routev1.RouteIngress) bool {
 
-	if len(oldStatus) == 0 || len(newStatus) == 0 {
+	if len(oldStatus) != len(newStatus) {
 		return false
 	}
 	exists := []string{}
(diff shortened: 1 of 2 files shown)


akshayhavile:
build ako

akshayhavile:
build ako

internal/status/route_status.go
```diff
@@ -306,15 +310,19 @@ func updateRouteObject(mRoute *routev1.Route, updateOption UpdateStatusOptions,
 }
 
 func compareRouteStatus(oldStatus, newStatus []routev1.RouteIngress) bool {
-	if len(oldStatus) != len(newStatus) {
+
+	if len(oldStatus) == 0 || len(newStatus) == 0 {
```
monotosh-avi, line 314: In openshift, route may already have a status set by the default ingress controller, empty length check may not work. 
Is there a problem if we retain the old condition check ?

akshayhavile: @monotosh-avi  With addition of timestamp into status message, each time length differ even if there is no ip update (in old status message retrieved .. time comes with some modification.. due to which length differ). That leads to unnecessary updates and patching of routes. In my testing, I have seen len(newStatus) == 0 condition. len(oldStatus) condition is added for same reason (never encountered this during testing.. but added to take into account extreme case) . IF we are sure that there will be no case with len(oldStatus) == 0, we can take out len(oldStatus) == 0 condition.

akshayhavile:
build ako

monotosh-avi (approved):

sudswasavi (approved):
I tried to verify the code to see if the conditions will lead to any unnecessary updates to the status (let's say when AKO is rebooted) but it didn't appear to be the case. Hence approving. Few checklists for this fix would be:
 - Conditions updated with last transition while AKO is running.
 - Conditions not updated if AKO is rebooted and the syncIngressstatus runs during initial boot up.

PR #214, opened 2020-12-02
move lock to StatusModel from nested AviApiRestStatus object

shubhamavi:
https://eli.thegreenplace.net/2018/beware-of-copying-mutexes-in-go/

moved the lock from a nested non-pointered object to the root struct.
Fixes intermittently occurring UT failures http://jenkins_url/job/ako_OS-pr-builder/563/console 
and possible deadlocks

`go vet` run on the file is clear post this, showed warnings earlier


sudswasavi (approved):

shubhamavi:
build ako

PR #216, opened 2020-12-04
Support for using default secret for TLS route if cert is not specified on route

monotosh-avi:
- If key/cert is not specified on TLS route, then a secret named router-certs-default would be used.
- This secret should be presesnt in avi-system namesapce.
- So support this, routes are now stored as namespace/routename per secret.
- Similarly secret per route is being stored as namesapce/secretname.

### DIFF
--- a/internal/objects/ingress_to_service.go
+++ b/internal/objects/ingress_to_service.go
@@ -361,20 +361,27 @@ func (v *SvcNSCache) UpdateIngressMappings(ingName string, svcName string) {
 	}
 }
 
-func (v *SvcNSCache) UpdateIngressSecretsMappings(ingName string, secret string) {
+func (v *SvcNSCache) AddIngressToSecretsMappings(ingressNS, ingName, secretName string) {
 	v.IngressLock.Lock()
-	defer v.IngressLock.Unlock()
-	_, ingresses := v.GetSecretToIng(secret)
-	if !utils.HasElem(ingresses, ingName) {
-		ingresses = append(ingresses, ingName)
-		v.UpdateSecretToIngMapping(secret, ingresses)
+	nsIngress := ingressNS + "/" + ingName
+	_, ingresses := v.GetSecretToIng(secretName)
+	if !utils.HasElem(ingresses, nsIngress) {
+		ingresses = append(ingresses, nsIngress)
+		v.UpdateSecretToIngMapping(secretName, ingresses)
 	}
+	v.IngressLock.Unlock()
+}
+
+func (v *SvcNSCache) AddSecretsToIngressMappings(secretNS, ingName, secretName string) {
+	v.IngressLock.Lock()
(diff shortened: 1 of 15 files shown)


monotosh-avi:
build ako

monotosh-avi:
build ako

internal/objects/ingress_to_service.go
```diff
@@ -361,20 +361,27 @@ func (v *SvcNSCache) UpdateIngressMappings(ingName string, svcName string) {
 	}
 }
 
-func (v *SvcNSCache) UpdateIngressSecretsMappings(ingName string, secret string) {
+func (v *SvcNSCache) AddIngressToSecretsMappings(ingressNS, ingName, secretName string) {
```
hemantavi, line 364: Since this function adds secret to mapping, probably we should change the name to: `AddSecretToIngressMappings`.

internal/objects/ingress_to_service.go
```diff
@@ -361,20 +361,27 @@ func (v *SvcNSCache) UpdateIngressMappings(ingName string, svcName string) {
...
-	if !utils.HasElem(ingresses, ingName) {
-		ingresses = append(ingresses, ingName)
-		v.UpdateSecretToIngMapping(secret, ingresses)
+	nsIngress := ingressNS + "/" + ingName
+	_, ingresses := v.GetSecretToIng(secretName)
+	if !utils.HasElem(ingresses, nsIngress) {
+		ingresses = append(ingresses, nsIngress)
+		v.UpdateSecretToIngMapping(secretName, ingresses)
 	}
+	v.IngressLock.Unlock()
```
hemantavi, line 372: We should keep the `defer v.IngressLock.Unlock()`.

hemantavi (commented):
LGTM, minor comments.

monotosh-avi:
build ako

pkg/utils/utils.go
```diff
@@ -388,3 +385,24 @@ func retrieveNSList(nsList map[string]bool) []string {
 	}
 	return namespaces
 }
+
+// This utility returns a true/false depending on whether
+// the user requires advanced L4 functionality
+func GetAdvancedL4() bool {
+	advanceL4 := os.Getenv(ADVANCED_L4)
+	if advanceL4 == "true" {
```
sudswasavi, line 393: A better way of expressing this is:
	if ok, _ := strconv.ParseBool(os.Getenv(ADVANCED_L4)); ok {
		return true
	}

sudswasavi (commented):
A minor nit. Otherwise looks good.

pkg/utils/utils.go
monotosh-avi: Updated.

sudswasavi (approved):

hemantavi (approved):

PR #217, opened 2020-12-06
AKO CRD v1alpha1 HostRule/HTTPRule new fields

shubhamavi:
- Introduces new fields in HostRule/HTTPRule CRDs
HostRule: ErrorPageProfile, AnalyticsProfile, SSLProfile, Datascripts, EnabledVirtualHost
HTTPRule: HealthMonitors, DestinationCA
- UTs for the additional fields are also added
<br/>
We are not upgrading the CRD version to v1alpha2. The existing fields work as is and the new ones donot break any existing functionality. For more details about upgrading CRD versions refer https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility
<br/>
<br/>

- Adds conditional ingress class clusterrole
- Addition/updates to the update-codegen scripts and README for usage guidelines.
- This also fixes some issues with the PKIProfile related Rest layer workflow, wherein the pkiprofile was not being deleted, but just getting detached from the pool object.
- Adds a `build-local` target in Makefile for running the go build locally on the host, and not inside a docker container. (It's faster and the build would be host arch specific)


internal/apis/ako/v1alpha2/hostrule.go
```diff
@@ -0,0 +1,100 @@
...
+	ErrorPageProfile
+	Datascripts
+	EnableVirtualService
+
+*/
+type HostRuleVirtualHost struct {
+	AnalyticsProfile     string             `json:"analyticsProfile,omitempty"`
+	ApplicationProfile   string             `json:"applicationProfile,omitempty"`
+	Datascripts          []string           `json:"datascripts,omitempty"`
+	EnableVirtualService bool               `json:"enableVirtualService,omitempty"`
```
shubhamavi, line 52: suggest variable names

shubhamavi: EnableVirtualHost

internal/apis/ako/v1alpha2/httprule.go
```diff
@@ -0,0 +1,81 @@
...
+
+// HTTPRuleTLS holds secure path/pool specific properties
+/*
+	new refs
+	PKIProfile
+*/
+type HTTPRuleTLS struct {
+	Type       string `json:"type,omitempty"`
+	SSLProfile string `json:"sslProfile,omitempty"`
+	PKIProfile string `json:"pkiProfile,omitempty`
```
shubhamavi, line 63: destinationCA

shubhamavi:
build ako

internal/nodes/avi_model_nodes.go
```diff
@@ -523,11 +528,17 @@ func (v *AviVsNode) CalculateCheckSum() {
...
-	})
-	vsRefs := v.WafPolicyRef + v.AppProfileRef + utils.Stringify(policies)
+	scripts := v.VsDatascriptRefs
+
+	vsRefs := v.WafPolicyRef +
+		v.AppProfileRef +
+		utils.Stringify(policies) +
+		v.AnalyticsProfileRef +
+		v.ErrorPageProfileRef +
+		v.SSLProfileRef +
```
sudswasavi, line 540: Just would like to be sure that this does not affect the checksum of an existing virtualservice. If you do a stringily on an empty list, it would still add a checksum.

internal/nodes/validator.go
```diff
@@ -115,6 +115,42 @@ func sslKeyCertHostRulePresent(key, host string) (bool, string) {
...
+	rule, isValid := pathRules[path]
+	if !isValid {
+		utils.AviLog.Debugf("key: %s, msg: Couldn't find path %s to httprule mapping in cache", key, path)
+		return false, ""
+	}
+
+	ruleNSName := strings.Split(rule, "/")
+	// from hostrule check if hostrule.TLS.SSLKeyCertificate is not null
+	httpRuleObj, err := lib.GetCRDInformers().HTTPRuleInformer.Lister().HTTPRules(ruleNSName[0]).Get(ruleNSName[1])
+	if err != nil {
```
sudswasavi, line 135: Can these two blocks if and else if becoming 1 block of if?

internal/rest/avi_obj_pool.go
```diff
@@ -122,13 +121,18 @@ func (rest *RestOperations) AviPoolBuild(pool_meta *nodes.AviPoolNode, cache_obj
 		pool.Servers = append(pool.Servers, &s)
 	}
 
-	var hm string
-	if pool_meta.Protocol == utils.UDP {
-		hm = fmt.Sprintf("/api/healthmonitor/?name=%s", utils.AVI_DEFAULT_UDP_HM)
+	// overwrite with healthmonitors provided by CRD
+	if len(pool_meta.HealthMonitors) > 0 {
```
sudswasavi, line 125: does it make sense to move this entire piece of code into layer 2 and ensure all transition cases are gracefully handled?

internal/rest/avi_obj_vs.go
```diff
@@ -230,7 +234,16 @@ func (rest *RestOperations) AviVsSniBuild(vs_meta *nodes.AviVsNode, rest_method
 		pool_ref := "/api/pool/?name=" + vs_meta.DefaultPool
 		sniChild.PoolRef = &pool_ref
 	}
-	var rest_ops []*utils.RestOp
+
+	var datascriptCollection []*avimodels.VSDataScripts
+	for i, script := range vs_meta.VsDatascriptRefs {
+		j := int32(i)
```
sudswasavi, line 240: isn't the loop already initializing the value of 'I' and 'script' every time you iterate through it? Then why again re-init them into j and datascript?

sudswasavi (commented):
@hemantavi can you pls review it as well?

internal/nodes/validator.go
shubhamavi: This is  just for different logs, i believe the two logs need to be called out separately.

internal/rest/avi_obj_vs.go
shubhamavi: the `range` functionality does not re-initialize these variables. So if we take the address of this, it'll always end up  being the same.  https://stackoverflow.com/questions/48826460/using-pointers-in-a-for-loop

internal/nodes/crd_translator.go
```diff
@@ -92,15 +132,21 @@ func BuildL7HostRule(host, namespace, ingName, key string, vsNode *AviVsNode) {
...
 
 	vsNode.SSLKeyCertAviRef = vsSslKeyCertificate
 	vsNode.WafPolicyRef = vsWafPolicy
 	vsNode.HttpPolicySetRefs = vsHTTPPolicySets
 	vsNode.AppProfileRef = vsAppProfile
+	vsNode.AnalyticsProfileRef = vsAnalyticsProfile
+	vsNode.ErrorPageProfileRef = vsErrorPageProfile
+	vsNode.SSLProfileRef = vsSslProfile
+	vsNode.VsDatascriptRefs = vsDatascripts
+	vsNode.Enabled = hostrule.Spec.VirtualHost.EnableVirtualHost
```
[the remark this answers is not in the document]
shubhamavi: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/217/files/2ab42350e9b8f263b34d93a9006cbf9d807d1539#diff-69ab2338bb5b9be3fd86d13a9e41ec013d5f93ab75dc1a6f169ab5dbff885c35R552

`utils.Hash(utils.Stringify(v.Enabled)) +`

It's here, i've checked the  functionality manually as well.

internal/nodes/crd_translator.go
```diff
@@ -31,6 +31,39 @@ import (
...
 )
 
+func removeVSNodeRefs(crd string, vsNode *AviVsNode) {
+	// Fields based off Hostrule CRD
+	if crd == lib.HostRule {
+		vsNode.Enabled = true
+		vsNode.SSLKeyCertAviRef = ""
+		vsNode.WafPolicyRef = ""
+		vsNode.HttpPolicySetRefs = []string{}
+		vsNode.AppProfileRef = ""
```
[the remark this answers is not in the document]
shubhamavi: yes, this is for cases when someone deletes the hostrule/httprule. In that case hostrule/httprule would be nil and all crd functionalities will need to be revoked from the VS/Pool structs

internal/rest/avi_obj_pool.go
shubhamavi: This'll have to then go in several other model generation areas/translators.

internal/k8s/crdcontroller.go
```diff
@@ -41,6 +41,21 @@ func NewCRDInformers(cs akocrd.Interface) {
 	})
 }
 
+func isHTTPRuleUpdated(oldHTTPRule, newHTTPRule *akov1alpha1.HTTPRule) bool {
+	if oldHTTPRule.ResourceVersion == newHTTPRule.ResourceVersion {
+		return false
+	}
+
+	oldSpecHash := utils.Hash(utils.Stringify(oldHTTPRule.Spec))
+	newSpecHash := utils.Hash(utils.Stringify(newHTTPRule.Spec))
```
hemantavi, line 50: What if someone changes the order of two properties? That will give a different hash, right?

internal/nodes/crd_translator.go
```diff
@@ -31,6 +31,37 @@ import (
...
+		vsNode.ErrorPageProfileRef = ""
+		vsNode.SSLProfileRef = ""
+		vsNode.VsDatascriptRefs = []string{}
+		if vsNode.ServiceMetadata.CRDStatus.Value != "" {
+			vsNode.ServiceMetadata.CRDStatus.Status = "INACTIVE"
+		}
+	}
+
+	// fields based off HTTPRule CRD
+	if crd == lib.HTTPRule {
```
hemantavi, line 52: We should either return from the previous if, or make this an `else if` condition.

internal/nodes/validator.go
```diff
@@ -115,6 +115,42 @@ func sslKeyCertHostRulePresent(key, host string) (bool, string) {
...
+		utils.AviLog.Warnf("key: %s, msg: Couldn't find httprule %s", key, rule)
```
hemantavi, line 136: Can't there be any other errors? If yes, can we also log the error here?

sudswasavi (approved):

hemantavi (approved):

shubhamavi:
build ako

shubhamavi:
build ako

shubhamavi:
build ako

PR #220, opened 2020-12-08
Fixing the usable_network_uuids deprecation for GCP Cloud.

parimanur:
Fixing the error in the VsVip creation.
`ipam_network_subnet not set in vip, no usable_networks in ipam_profile`

### DIFF
--- a/internal/rest/avi_obj_vs.go
+++ b/internal/rest/avi_obj_vs.go
@@ -641,7 +641,27 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, cache_
 		vip := avimodels.Vip{AutoAllocateIP: &auto_alloc}
 		networkRef := lib.GetNetworkName()
 		if lib.IsPublicCloud() && lib.GetNetworkName() != "" {
-			vip.SubnetUUID = &networkRef
+
+			if lib.GetCloudType() == lib.CLOUD_GCP {
+				// add the IPAMNetworkSubnet
+				intCidr, err := strconv.ParseInt(lib.GetSubnetPrefix(), 10, 32)
+				if err != nil {
+					utils.AviLog.Warnf("The value of CIDR couldn't be converted to int32. Defaulting to /24")
+					intCidr = 24
+				}
+				subnetMask := int32(intCidr)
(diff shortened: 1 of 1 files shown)


internal/rest/avi_obj_vs.go
```diff
@@ -641,7 +641,27 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, cache_
...
 		networkRef := lib.GetNetworkName()
 		if lib.IsPublicCloud() && lib.GetNetworkName() != "" {
-			vip.SubnetUUID = &networkRef
+
+			if lib.GetCloudType() == lib.CLOUD_GCP {
+				// add the IPAMNetworkSubnet
+				intCidr, err := strconv.ParseInt(lib.GetSubnetPrefix(), 10, 32)
+				if err != nil {
+					utils.AviLog.Warnf("The value of CIDR couldn't be converted to int32. Defaulting to /24")
+					intCidr = 24
```
sudswasavi, line 650: why are we defaulting to /24? Is this something documented in GCP? (if so can we cite it as code comment). Else it leads to some kind of an unknown behaviour.

sudswasavi (commented):
@parimanur an additional request as a part of this PR, can we also include a case where the network-name is not specified for Vcenter cloud but we can work with an auto-allocate only in that case, like how we do for advancedl4 currently.

internal/rest/avi_obj_vs.go
shubhamavi: This seems to be the case in normal AKO workflow as well, where we default to /24 in case a subnetIP is provided but subnetPrefix is not provided or is invalid

parimanur:
@sudswasavi Sure will take this up in the next PR.

monotosh-avi (approved):

shubhamavi (approved):

internal/rest/avi_obj_vs.go
sudswasavi: Is this documented anywhere? Maybe we should if not.

parimanur: I have taken note of this. Will add it in the helm docs repo.

PR #221, opened 2020-12-08
Added Rebootability & Modified testcases.

RiyaGaniga-avi:
- Added Rebootability. 
- Modified testcases.
- Hybrid execution is enabled.
- Modified testbed format.
- Updation of ingresses added.


tests/scaletest/scale_test.go
```diff
@@ -161,7 +277,26 @@ func PoolVerification(t *testing.T) bool {
 	return false
 }
 
+/* Verifies if all requires DNS A records are created in the DNS VS or not */
+func DNSARecordsVerification(t *testing.T, hostNames []string) bool {
```
monotosh-avi, line 281: The logic of this function may has to be tweaked a bit. E.g. if there were 2 shard VSes initially, then another 2 shard VSes gets created during test run, then len(diffString) would be 4, and len(diffString)-val) would be 0. The following condition would not be true.
    if (len(diffString)-val) == initialNumOfFQDN || len(diffString) == initialNumOfFQDN

tests/scaletest/scale_test.go
```diff
@@ -391,78 +574,285 @@ func DeleteIngressesSerial(t *testing.T, numOfIng int, initialNumOfPools int, Av
...
+			updatedIngresses, err := lib.UpdateIngress(namespace, toUpdateIngresses)
+			if err != nil {
+				t.Fatalf("Error updating ingresses as : %v ", err)
+				return
+			}
+			t.Logf("Updated ingresses %s", updatedIngresses)
+			ingressesUpdated = append(ingressesUpdated, updatedIngresses...)
+		}
+		time.Sleep(time.Second * 1)
+		mutex.Unlock()
```
monotosh-avi, line 626: We can use defer for the unlock

tests/scaletest/scale_test.go
```diff
@@ -391,78 +574,285 @@ func DeleteIngressesSerial(t *testing.T, numOfIng int, initialNumOfPools int, Av
...
+		if len(toUpdateIngresses) > 0 {
+			updatedIngresses, err := lib.UpdateIngress(namespace, toUpdateIngresses)
+			if err != nil {
+				t.Fatalf("Error updating ingresses as : %v ", err)
+				return
+			}
+			t.Logf("Updated ingresses %s", updatedIngresses)
+			ingressesUpdated = append(ingressesUpdated, updatedIngresses...)
+		}
+		time.Sleep(time.Second * 1)
```
monotosh-avi, line 625: Check if we can use a periodic timer and a channel, e.g. 
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L420

monotosh-avi (commented):
Looks great. Now we have an exhaustive set of test cases. 
Few minor comments.

tests/scaletest/scale_test.go
```diff
@@ -391,78 +574,285 @@ func DeleteIngressesSerial(t *testing.T, numOfIng int, initialNumOfPools int, Av
...
+	REBOOTNODE = true
+	UpdateIngressesParallel(t, numOfIng)
+	REBOOTNODE = false
 }
 
 func TestMain(t *testing.M) {
 	Setup()
-	t.Run()
-	Cleanup()
+	// t.Run()
```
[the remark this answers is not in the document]
RiyaGaniga-avi: My bad, missed it. Will uncomment.

tests/scaletest/scale_test.go
RiyaGaniga-avi: Yup, that's a better way, noted!

tests/scaletest/scale_test.go
RiyaGaniga-avi: Nice catch! Making changes to the logic

tests/scaletest/scale_test.go
RiyaGaniga-avi: Realised 'defer' isn't the right way here. As the mutex is being locked and unlocked for each iteration of the loop.

tests/scaletest/scale_test.go
RiyaGaniga-avi: Eliminating the wait part altogether

tests/scaletest/scale_test.go
```diff
@@ -391,78 +574,285 @@ func DeleteIngressesSerial(t *testing.T, numOfIng int, initialNumOfPools int, Av
...
+	// Cleanup()
 }
 
-func TestParallel200InsecureIngresses(t *testing.T) {
+func TestInsecureParallelCreationUpdationDeletionWithoutReboot(t *testing.T) {
+	SetupForTesting(t)
 	ingressType = INSECURE
-	ParallelIngressHelper(t, 200)
+	CreateIngressesParallel(t, numOfIng, initialNumOfPools)
```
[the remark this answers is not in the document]
RiyaGaniga-avi: The variable ingressType is used by 14 functions in total. So I think it would be better to keep it as a global variable. Should I still send it as an argument to functions?

tests/scaletest/lib/ingress.go
```diff
@@ -333,15 +349,49 @@ func DeleteIngress(namespace string, listOfIngressToDelete []string) ([]string,
 	return listOfDeletedIngresses, nil
 }
 
-func ListIngress(t *testing.T, namespace string) error {
-	t.Logf("Listing ingress in namespace %q:\n", namespace)
+func UpdateIngress(namespace string, listOfIngressToUpdate []string) ([]string, error) {
+	for i := 0; i < len(listOfIngressToUpdate); i++ {
```
hemantavi, line 353: It will do a length calculation everytime, can we do it before hand and store it in a variable and use that variable here instead?

tests/scaletest/lib/ingress.go
```diff
@@ -333,15 +349,49 @@ func DeleteIngress(namespace string, listOfIngressToDelete []string) ([]string,
...
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			result, getErr := kubeClient.Resource(ingressResource).Namespace(namespace).Get(ctx, listOfIngressToUpdate[i], metaV1.GetOptions{})
```
hemantavi, line 355: Since you are referencing the `listOfIngressToUpdate[i]` anyway, can we just change the for loop above to:
```
for _, ing := range listOfIngressToUpdate {
```

tests/scaletest/lib/ingress.go
```diff
@@ -333,15 +349,49 @@ func DeleteIngress(namespace string, listOfIngressToDelete []string) ([]string,
...
+		})
+		if retryErr != nil {
+			return listOfIngressToUpdate, retryErr
+		}
+	}
+	return listOfIngressToUpdate, nil
+}
+
+func ListIngress(t *testing.T, namespace string) ([]string, error) {
+	var listOfIngress []string
```
hemantavi, line 380: Can we change the variable to `ingressList`?

tests/scaletest/lib/parser.go
```diff
@@ -25,6 +25,13 @@ type Platform struct {
 	VCenterConfiguration VCenterConfiguration `json:"vcenter_configuration"`
 }
 
+type Nodes struct {
+	Name     string `json:"name"`
+	Ip       string `json:"ip"`
```
hemantavi, line 30: nit: Ip -> IP

tests/scaletest/scale_test.go
```diff
@@ -319,41 +499,56 @@ func DeleteIngressesParallel(t *testing.T, numOfIng int, initialNumOfPools int,
...
+			go parallelIngressUpdation(t, &wg, namespace, ingressesCreated[nextStartInd:nextStartInd+blockSize+1])
+			nextStartInd = nextStartInd + blockSize + 1
+		} else {
+			go parallelIngressUpdation(t, &wg, namespace, ingressesCreated[nextStartInd:nextStartInd+blockSize])
+			nextStartInd = nextStartInd + blockSize
+		}
+	}
+	wg.Wait()
+	g.Expect(ingressesUpdated).To(gomega.HaveLen(numOfIng))
+	t.Logf("*** Updated %d Ingresses ***\n", numOfIng)
```
hemantavi, line 529: Please remove `***` from the log.

tests/scaletest/lib/utils.go
```diff
@@ -213,3 +213,45 @@ func FetchPools(t *testing.T, AviClient *clients.AviClient) []models.Pool {
...
+	}
+	elems := make([]json.RawMessage, result.Count)
+	err = json.Unmarshal(result.Results, &elems)
+	if err != nil {
+		t.Logf("Failed to unmarshal VS data, err: %v", err)
+	}
+	virtualservice := models.VirtualServiceRuntime{}
+	for i := 0; i < len(elems); i++ {
+		vs := models.VirtualServiceRuntime{}
+		err = json.Unmarshal(elems[i], &vs)
```
hemantavi, line 231: Error check?

tests/scaletest/lib/utils.go
```diff
@@ -213,3 +213,45 @@ func FetchPools(t *testing.T, AviClient *clients.AviClient) []models.Pool {
...
+		dnsRecord = *virtualservice.IPAMDNSRecords[i]
+		ipamDNSRecords = append(ipamDNSRecords, dnsRecord)
+	}
+	return ipamDNSRecords
+}
+
+func FetchDNSARecordsFQDN(t *testing.T, dnsVsUuid string, AviClient *clients.AviClient) []string {
+	ipamDNSRecords := FetchDnsVsFqdn(t, dnsVsUuid, AviClient)
+	var FQDNList []string
+	for i := 0; i < len(ipamDNSRecords); i++ {
```
hemantavi, line 249: Can we please use `for ... range` here and everywhere else?

tests/scaletest/lib/utils.go
```diff
@@ -213,3 +213,45 @@ func FetchPools(t *testing.T, AviClient *clients.AviClient) []models.Pool {
...
+		if *ipamDNSRecords[i].Type == "DNS_RECORD_A" {
+			for j := 0; j < len(ipamDNSRecords[i].Fqdn); j++ {
+				FQDNList = append(FQDNList, ipamDNSRecords[i].Fqdn[j])
```
hemantavi, line 252: If you are just trying to copy the fqdn list, you can just use copy().

tests/scaletest/scale_test.go
```diff
@@ -29,57 +31,83 @@ import (
...
-	numGoRoutines, err = strconv.Atoi(os.Args[4])
+	numGoRoutines, err = strconv.Atoi(os.Args[6])
 	if err != nil {
 		numGoRoutines = 5
 	}
 	if numGoRoutines <= 0 {
 		fmt.Println("ERROR : Number of Go Routines cannot be zero or negative.")
 		os.Exit(0)
 	}
+	numOfIng, err = strconv.Atoi(os.Args[7])
```
hemantavi, line 92: Error check.

tests/scaletest/scale_test.go
```diff
@@ -29,57 +31,83 @@ import (
...
+	testbed, er := os.Open(testbedFileName)
+	if er != nil {
+		fmt.Println("ERROR : Error opening testbed file ", testbedFileName, " with error : ", err)
+		os.Exit(0)
+	}
+	defer testbed.Close()
+	byteValue, _ := ioutil.ReadAll(testbed)
```
hemantavi, line 99: Can we get the error here and do an error check for this?

tests/scaletest/scale_test.go
```diff
@@ -29,57 +31,83 @@ import (
...
+	json.Unmarshal(byteValue, &testbedParams)
```
hemantavi, line 100: Error check for this too.

tests/scaletest/scale_test.go
```diff
@@ -109,6 +138,77 @@ func Cleanup() {
...
+}
+
+/* Used for Controller and Node reboot */
+func Reboot(t *testing.T, wg *sync.WaitGroup, nodeType string, controllerIP string, username string, password string) {
+	t.Logf("Rebooting %s ... ", nodeType)
+	loginID := username + "@" + controllerIP
+	cmd := exec.Command("sshpass", "-p", password, "ssh", "-t", loginID, " `echo ", password, " |  sudo -S shutdown --reboot 0 && exit `")
+	_, err := cmd.Output()
+	if err != nil {
+		fmt.Println(err.Error())
```
hemantavi, line 168: Use `t.Log()` instead of `fmt.Println()`.

tests/scaletest/scale_test.go
```diff
@@ -109,6 +138,77 @@ func Cleanup() {
...
+	}
+	t.Logf("%s Rebooted", nodeType)
```
hemantavi, line 170: Please put this inside else, if no error.

tests/scaletest/scale_test.go
```diff
@@ -109,6 +138,77 @@ func Cleanup() {
...
+	defer wg.Done()
+}
+
+/* Reboots AKO pod */
+func RebootAko(t *testing.T, wg *sync.WaitGroup) {
+	akoNamespace := "avi-system"
```
hemantavi, line 176: Use a constant to refer to the `avi-system` namespace.

tests/scaletest/scale_test.go
```diff
@@ -109,6 +138,77 @@ func Cleanup() {
...
+		wg.Add(1)
+		go Reboot(t, wg, CONTROLLER, os.Getenv("CTRL_IPADDRESS"), os.Getenv("CTRL_USERNAME"), os.Getenv("CTRL_PASSWORD"))
+	}
+	if REBOOTNODE == true {
+		wg.Add(1)
+		var testbedParams lib.TestbedFields
+		testbed, err := os.Open(testbedFileName)
+		if err != nil {
+			fmt.Println("ERROR : Error opening testbed file ", testbedFileName, " with error : ", err)
+			os.Exit(0)
```
hemantavi, line 202: Use `t.Fatalf()` instead of `Println()` and `os.Exit()`.

tests/scaletest/scale_test.go
```diff
@@ -109,6 +138,77 @@ func Cleanup() {
...
+		}
+		defer testbed.Close()
+		byteValue, _ := ioutil.ReadAll(testbed)
```
hemantavi, line 205: Error check for ReadAll.

tests/scaletest/scale_test.go
```diff
@@ -161,7 +282,30 @@ func PoolVerification(t *testing.T) bool {
...
+/* Verifies if all requires DNS A records are created in the DNS VS or not */
+func DNSARecordsVerification(t *testing.T, hostNames []string) bool {
+	t.Logf("Verifying DNS A Records...")
+	FQDNList := lib.FetchDNSARecordsFQDN(t, dnsVSUUID, AviClients[0])
+	diffString := DiffOfLists(FQDNList, hostNames)
+	if len(diffString) == initialNumOfFQDN {
+		return true
+	}
+	newSharedVSFQDN := DiffOfLists(diffString, initialFQDNList)
+	var val int = 0
```
hemantavi, line 294: Default for an int is 0 anyway, so need to assign.

tests/scaletest/scale_test.go
```diff
@@ -161,7 +282,30 @@ func PoolVerification(t *testing.T) bool {
 	return false
 }
 
+/* Verifies if all requires DNS A records are created in the DNS VS or not */
+func DNSARecordsVerification(t *testing.T, hostNames []string) bool {
+	t.Logf("Verifying DNS A Records...")
+	FQDNList := lib.FetchDNSARecordsFQDN(t, dnsVSUUID, AviClients[0])
+	diffString := DiffOfLists(FQDNList, hostNames)
+	if len(diffString) == initialNumOfFQDN {
+		return true
```
hemantavi, line 291: Should we also check the contents of FQDNList? As in, the difference might be fine, but the contents might be different? @monotosh-avi @RiyaGaniga-avi

monotosh-avi: @hemantavi yeah I also felt the same, but couldn't find a condition where this won't work :-)

sudswasavi (commented):
@hemantavi / @monotosh-avi / @RiyaGaniga-avi I guess we need a README update as well in terms of how this is run? 
Do you see value in adding this to our nightly runs?

hemantavi:
Yeah, @RiyaGaniga-avi let's update the README too.

@sudswasavi yeah, we can make it part of our nightly runs.

tests/scaletest/lib/ingress.go
RiyaGaniga-avi: Done!

tests/scaletest/lib/ingress.go
RiyaGaniga-avi: Done!

RiyaGaniga-avi:
build ako

tests/scaletest/lib/utils.go
```diff
@@ -213,3 +213,45 @@ func FetchPools(t *testing.T, AviClient *clients.AviClient) []models.Pool {
 	}
 	return pools
 }
+
+func FetchDnsVsFqdn(t *testing.T, dnsVsUuid string, AviClient *clients.AviClient) []models.DNSRecord {
+	uri := "api/virtualservice"
+	result, err := AviClient.AviSession.GetCollectionRaw(uri)
+	if err != nil {
+		t.Logf("Get uri %v returned err for pool %v", uri, err)
```
hemantavi, line 221: Please change this to Fatalf or return after this line.

tests/scaletest/lib/utils.go
```diff
@@ -213,3 +213,45 @@ func FetchPools(t *testing.T, AviClient *clients.AviClient) []models.Pool {
...
+	}
+	elems := make([]json.RawMessage, result.Count)
+	err = json.Unmarshal(result.Results, &elems)
+	if err != nil {
+		t.Logf("Failed to unmarshal VS data, err: %v", err)
```
hemantavi, line 226: Same, Fatalf or return from here.

hemantavi (commented):
LGTM, a couple of minor comments.

tests/scaletest/lib/utils.go
RiyaGaniga-avi: Done

tests/scaletest/lib/utils.go
RiyaGaniga-avi: Done

shubhamavi:
build ako

hemantavi:
build ako

hemantavi (approved):

shubhamavi:
build ako

monotosh-avi:
build ako

sudswasavi (approved):

PR #223, opened 2020-12-11
svc/gw staticIP fix - provide IPAMNetworkSubnet.NetworkRef for vsvip if networkName is provided

shubhamavi:


sudswasavi (commented):
I talked to @parimanur about this but can we also include an additional feature in this PR.
If the user does not specify a networkName, can we default to auto-allocate from the IPAM? This is a behaviour present in advanced-l4 but would be useful for generic AKO as well for folks who want to configure all the usable networks in the IPAM but don't want AKO to use any particular usable network out of that. In which case, AKO can auto-allocate and get the IP from the IPAM directly.

shubhamavi:
As the subject mentions, this solves a bug in non-advancedL4 cases, wherein the networkName is not specified by the AKO consumer (AKO  functional test suite in this case). It would be better to address this as a separate feature request ticket? This is specifically a resolution for avi bug AV-102053

sudswasavi:
> AV-102053

Sure. I was just hoping that the other feature could also get into the build in 1.3.1 before we release it because a customer is waiting for it. I can do another PR for it. Thx.

sudswasavi (approved):

monotosh-avi (approved):

PR #226, opened 2020-12-15
Supporting multiple usable networks as a part of the IPAM by skipping validation for network_name.

parimanur:
- AKO will auto-allocate the ip from the list of usable networks.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -2654,7 +2654,8 @@ func checkPublicCloud(client *clients.AviClient) bool {
 	if lib.IsPublicCloud() {
 		// Handle all public cloud validations here
 		networkName := lib.GetNetworkName()
-		if networkName == "" {
+		if networkName == "" && lib.GetCloudType() != lib.CLOUD_GCP {
+			// networkName is required param for AWS and Azure Clouds
 			utils.AviLog.Errorf("Required param networkName not specified, syncing will be disabled.")
 			return false
 		}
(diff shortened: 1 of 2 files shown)


sudswasavi (commented):
Hi @parimanur  does it work without the adv-l4 with this change? 
@shubhamavi can you also pls review?

shubhamavi:
@sudswasavi  It should work as it is, it was just that the bootup check was not allowing AKO to come up since the networkName was a *required* field for non-advl4. For advl4 we anyways don't do that check.

Would it be a good idea to check that in case `networkName` is not specified in AKO, Avi is configured with atleast 1 usable network, during bootup. Although that would require a version specific API call from Avi since the ipamdnsprofile schema was modified between 20.1.2 and 20.1.3
cc: @apalsule

sudswasavi:
> > Hi @parimanur does it work without the adv-l4 with this change?
> 
> @sudswasavi It should work as it is, it was just that the bootup check was not allowing AKO to come up since the networkName was a _required_ field for non-advl4. For advl4 we anyways don't do that check.
> 
> @parimanur @sudswasavi Would it be a good idea to verify that in case `networkName` is not specified in AKO, Avi is configured with atleast 1 usable network, during bootup, in order to minimise vsvip failures. Although that would require a version specific API call from Avi since the ipamdnsprofile schema was modified between 20.1.2 and 20.1.3. I think this also entails some documentation updates as we need to mention how to set usable networks, if we don't do that already.
> cc: @apalsule

Can we get around this for now by documenting the usable networks being present if NetworkName is not provided and then take it up in 1.4.1? We have to clearly document what errors can be expected if such a scenario is hit.

shubhamavi (approved):

sudswasavi (approved):

PR #227, opened 2020-12-15
Change in AKO statefulset status condition message

monotosh-avi:
The type of the condition is now ako.vmware.com/ObjectDeletionInProgress, which signifies if AKO is deleting all Avi Objects. The value of status would be True only when the delete operation is in progress, else it would be set False. In case of an error like timeout, it would be set to unknown. 

- The status of this condition would be set to true after setting deleteConfig to true, before starting all object deletion.
- The reason would be set to ObjectDeletionStarted.
- The status would be set to false when all objects gets deleted.
- The reason would be set to ObjectDeletionDone.

For example:
Status condition before deletion:
  - lastTransitionTime: "2020-12-15T10:10:45Z"
    message: Started deleting objects
    reason: ObjectDeletionStarted
    status: "True"
    type: ako.vmware.com/ObjectDeletionInProgress

Status condition after deletion:
  - lastTransitionTime: "2020-12-15T10:10:48Z"
    message: Successfully deleted all objects
    reason: ObjectDeletionDone
    status: "False"
    type: ako.vmware.com/ObjectDeletionInProgress

### DIFF
--- a/internal/lib/constants.go
+++ b/internal/lib/constants.go
@@ -101,9 +101,9 @@ const (
 	AviObjDeletionTime                         = 30 // Minutes
 	AKOConditionType                           = "ako.vmware.com/ObjectDeletionInProgress"
 	AKOStatefulSet                             = "ako"
-	ObjectDeletionStartStatus                  = "ObjectDeletionStarted"
-	ObjectDeletionDoneStatus                   = "ObjectDeletionDone"
-	ObjectDeletionTimeoutStatus                = "ObjectDeletionTimeout"
+	ObjectDeletionStartStatus                  = "Started"
+	ObjectDeletionDoneStatus                   = "Done"
+	ObjectDeletionTimeoutStatus                = "Timeout"
 	DefaultIngressClassAnnotation              = "ingressclass.kubernetes.io/is-default-class"
 	DefaultRouteCert                           = "router-certs-default"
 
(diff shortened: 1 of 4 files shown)


sudswasavi (dismissed):

vijay-avi:
build ako

vijay-avi:
build ako

internal/lib/constants.go
```diff
@@ -99,11 +99,11 @@ const (
...
 	AviObjDeletionTime                         = 30 // Minutes
-	AKOConditionType                           = "akoStatus"
+	AKOConditionType                           = "ako.vmware.com/ObjectDeletionInProgress"
 	AKOStatefulSet                             = "ako"
-	ObjectDeletionStartStatus                  = "objDeletionStarted"
-	ObjectDeletionDoneStatus                   = "objDeletionDone"
-	ObjectDeletionTimeoutStatus                = "objDeletionTimeout"
+	ObjectDeletionStartStatus                  = "ObjectDeletionStarted"
+	ObjectDeletionDoneStatus                   = "ObjectDeletionDone"
+	ObjectDeletionTimeoutStatus                = "ObjectDeletionTimeout"
```
shubhamavi, line 106: since ObjectDeletion is common in all enums, can we chuck it out to just be `Started/Done/Timedout`?

shubhamavi (commented):
minor nit.

internal/lib/constants.go
monotosh-avi: Sounds fine to me. @sudswasavi ?

sudswasavi: yeah just on that thought, I wonder how we signify what objects these are. In this case, there are LB Objects I think. Thoughts?

sudswasavi (approved):

shubhamavi (approved):

PR #228, opened 2020-12-16
Renaming controllerIP to ControllerHost in helm values.yaml

monotosh-avi:
We can use both controller IP and hostname to connect to Avi Controller.
Renaming the field in values.yaml to reflect this.
Internal keynames are being kept unchanged.


akshayhavile (approved):
LGTM. Are we also planning to change variable name used inside AKO to make it consistent with field it represent?

monotosh-avi:
> LGTM. Are we also planning to change variable name used inside AKO to make it consistent with field it represent?

@akshayhavile Yes, we can do that after this release. At this point I wanted to make minimum changes.

shubhamavi (approved):

PR #241, opened 2021-01-12
Only set RHI flag if it's set to true

sudswasavi:
This commit will unblock usage of Avi essentials license where RHI
is not supported in the VS payload.

If RHI is set to true only then, we set it in the VS object,
for all other cases RHI is not set in the payload but defaulted
to false by the go sdk.


monotosh-avi (approved):

monotosh-avi:
build ako

shubhamavi (approved):

PR #246, opened 2021-01-18
Update statefulset status even if no VS has to be deleted

monotosh-avi:
After setting deleteConfig to true, update statefulset status even if no VS object has to be deleted.
Make sure this works, by checking if the model is empty and updating status after VRF update is done.


sudswasavi (approved):
Can you pls cherry-pick this to 1.3.2?

shubhamavi (approved):

PR #249, opened 2021-01-19
Update statefulset status even if no VS has to be deleted (#246)

monotosh-avi:
After setting deleteConfig to true, update statefulset status even if no VS object has to be deleted.
Make sure this works, by checking if the model is empty and updating status after VRF update is done.

(cherry picked from commit a89c73b34cb7988a956406d9f7ed45a3f7984fe1)


monotosh-avi:
build ako

PR #256, opened 2021-01-29
passing cache layer keys to status update calls for logging

shubhamavi:
This commit sanitises some of the status package logs and makes sure that we pass and log the REST layer keys to the logs.

### DIFF
--- a/internal/status/services_api_status.go
+++ b/internal/status/services_api_status.go
@@ -17,7 +17,6 @@ package status
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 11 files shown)


shubhamavi:
build ako

shubhamavi:
build ako

shubhamavi:
build ako

shubhamavi:
build ako

internal/status/services_api_status.go
```diff
@@ -302,13 +302,13 @@ func UpdateSvcApiGatewayStatusObject(gw *svcapiv1alpha1.Gateway, updateStatus *s
 		updatedGW, err := lib.GetServicesAPIClientset().NetworkingV1alpha1().Gateways(gw.Namespace).Get(context.TODO(), gw.Name, metav1.GetOptions{})
 		if err != nil {
 			utils.AviLog.Warnf("gateway not found %v", err)
-			return err
```
monotosh-avi, line 305: Don't we need the error anymore ?

shubhamavi: We've not been using those errors anywhere. We also don't have any logic to react to any these errors. In case of any error in such status update functions, we're just retrying the function. After a certain number of retries we eventually give that up as well.

monotosh-avi: Makes sense. No need to return unnecessary errors.

monotosh-avi (approved):

sudswasavi (approved):

PR #257, opened 2021-01-29
Introducing NodePortLocal in AKO

monotosh-avi:
In this first set of changes, addition of pool servers from NPL annotation
of Pod is handled.

- A new service type NodePortLocal is introduced. When this is set,
  servers would be populated from NPL annotation of POD.

- New PodInformers would be started only when this flag is set.

- New store objects are created to store NPL annotation per pod, and
  mapping between Pods and Services.

Not handled in this set of changes:
- Automatic annotation of services from ingresses for NPL
- Mapping between gateway and POD
- CNI verification

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -342,6 +342,72 @@ func AddRouteEventHandler(numWorkers uint32, c *AviController) cache.ResourceEve
 	return routeEventHandler
 }
 
+func AddPodEventHandler(numWorkers uint32, c *AviController) cache.ResourceEventHandler {
+	podEventHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if c.DisableSync {
+				return
+			}
+			if lib.IsNodePortMode() {
+				utils.AviLog.Debugf("skipping Pod for nodeport mode")
+				return
+			}
+			pod := obj.(*corev1.Pod)
+			namespace, _, _ := cache.SplitMetaNamespaceKey(utils.ObjKey(pod))
+			key := utils.Pod + "/" + utils.ObjKey(pod)
+			bkt := utils.Bkt(namespace, numWorkers)
+			c.workqueue[bkt].AddRateLimited(key)
+			utils.AviLog.Debugf("key: %s, msg: ADD\n", key)
+		},
+		DeleteFunc: func(obj interface{}) {
+			if c.DisableSync {
+				return
+			}
+			if lib.IsNodePortMode() {
+				utils.AviLog.Debugf("skipping Pod for nodeport mode")
+				return
+			}
+			pod, ok := obj.(*corev1.Pod)
+			if !ok {
+				tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+				if !ok {
+					utils.AviLog.Errorf("couldn't get object from tombstone %#v", obj)
+					return
+				}
+				pod, ok = tombstone.Obj.(*corev1.Pod)
+				if !ok {
+					utils.AviLog.Errorf("Tombstone contained object that is not an Pod: %#v", obj)
+					return
+				}
+			}
+			namespace, _, _ := cache.SplitMetaNamespaceKey(utils.ObjKey(pod))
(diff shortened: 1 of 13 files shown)


shubhamavi:
build ako

monotosh-avi:
build ako

internal/lib/utils.go
```diff
@@ -79,3 +89,68 @@ func GetSvcKeysForNodeCRUD() (svcl4Keys []string, svcl7Keys []string) {
 	return svcl4Keys, svcl7Keys
 
 }
+
+func GetPodsFromService(namespace, serviceName string) []utils.NamespaceName {
+	var pods []utils.NamespaceName
+	svcKey := namespace + "/" + serviceName
+	svc, err := utils.GetInformers().ClientSet.CoreV1().Services(namespace).Get(context.TODO(), serviceName, metav1.GetOptions{})
```
shubhamavi, line 96: ServiceInformer rather than ClientSet?

internal/nodes/dequeue_ingestion.go
```diff
@@ -153,6 +161,34 @@ func DequeueIngestion(key string, fullsync bool) {
 	}
 }
 
+// handlePod populates NPL annotations for a pod in store.
+// It also stores a mapping of Pod to Services for future use
+func handlePod(key, namespace, podName string) {
+	podKey := namespace + "/" + podName
+	pod, err := utils.GetInformers().ClientSet.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
```
shubhamavi, line 168: same here, can't we use PodInformer?

internal/k8s/controller.go
```diff
@@ -342,6 +342,72 @@ func AddRouteEventHandler(numWorkers uint32, c *AviController) cache.ResourceEve
 	return routeEventHandler
 }
 
+func AddPodEventHandler(numWorkers uint32, c *AviController) cache.ResourceEventHandler {
+	podEventHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if c.DisableSync {
+				return
+			}
+			if lib.IsNodePortMode() {
```
sudswasavi, line 351: Is this required? Since you are only calling this method on serviceType equals `NodePortLocal'?

internal/lib/lib.go
```diff
@@ -488,6 +488,7 @@ func InformersToRegister(oclient *oshiftclient.Clientset, kclient *kubernetes.Cl
 		utils.EndpointInformer,
 		utils.SecretInformer,
 		utils.ConfigMapInformer,
+		utils.PodInformer,
```
sudswasavi, line 491: Should this be added only if serviceType is NodePortLocal?

internal/nodes/ingress_model_rel.go
```diff
@@ -215,6 +221,27 @@ func SvcToGateway(svcName string, namespace string, key string) ([]string, bool)
 	return allGateways, true
 }
 
+// PodToRoute fetches the list of impacted Routes from Pod update.
+// First fetch list of Services for the Pod.
+// Then get list of Routes for the services.
+func PodToRoute(podName string, namespace string, key string) ([]string, bool) {
```
sudswasavi, line 227: NodePortLocal isn't available with OpenShift currently, should we hold off from introducing this code till that happens?

internal/k8s/controller.go
monotosh-avi: Removed.

internal/lib/lib.go
monotosh-avi: Yeah, added a check

internal/lib/utils.go
monotosh-avi: Good catch, updated.

internal/nodes/ingress_model_rel.go
monotosh-avi: I wanted to keep it just to be compatible. But I guess we can add it later when Antrea is used in Openshift. Removed for now.

sudswasavi (approved):
LGTM

shubhamavi (approved):

PR #258, opened 2021-02-01
Fix typo in sample AKOConfig

andrew-su:
The typo causes the controller settings to be set to empty, which in turn makes the ako pod fail to detect the controller.


andrew-su:
Closing, to fix CLA.

PR #261, opened 2021-02-03
Granular RBAC support with AKO

akshayhavile:
Adding GRBAC support for AKO.  Enabling default support for controller version 20.1.2 and above. GRBAC will come into picture if user is associated with labelled role.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -1425,19 +1425,6 @@ func (c *AviObjCache) PopulateSSLKeyToCache(client *clients.AviClient, cloud str
 				utils.AviLog.Warnf("Wrong data type for ssl key: %s in cache", k)
 			}
 		}
-		var cacert string
-		// Find CA Cert name from the cache for checksum calculation.
-		if SslKeyData[i].HasCARef {
-			ca, found := c.SSLKeyCache.AviCacheGetNameByUuid(SslKeyCacheObj.CACertUUID)
-			if !found {
-				utils.AviLog.Warnf("cacertUUID %s for keycert %s not found in cache", SslKeyCacheObj.CACertUUID, SslKeyCacheObj.Name)
-			} else {
-				cacert = ca.(string)
-			}
-		}
-		//TODO: Check with team : What is reason of recomputing checksum here again.
-		SslKeyData[i].CloudConfigCksum = lib.SSLKeyCertChecksum(SslKeyCacheObj.Name, SslKeyCacheObj.Cert, cacert)
-		SslKeyData[i].CloudConfigCksum += lib.GetClusterLabelChecksum()
 		utils.AviLog.Debugf("Adding key to sslkey cache :%s", k)
 		c.SSLKeyCache.AviCacheAdd(k, &SslKeyData[i])
 		delete(sslCacheData, k)
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -461,12 +461,15 @@ func (c *AviObjCache) AviPopulateAllPkiPRofiles(client *clients.AviClient, pkiDa
 			utils.AviLog.Warnf("Incomplete pki data unmarshalled, %s", utils.Stringify(pki))
 			continue
 		}
-
+		checksum := lib.SSLKeyCertChecksum(*pki.Name, string(*pki.CaCerts[0].Certificate), "")
+		if lib.GetEnableGRBAC() && pki.Labels != nil {
+			checksum += lib.ObjectLabelChecksum(pki.Labels)
+		}
 		pkiCacheObj := AviPkiProfileCache{
 			Name:             *pki.Name,
 			Uuid:             *pki.UUID,
 			Tenant:           lib.GetTenant(),
-			CloudConfigCksum: lib.SSLKeyCertChecksum(*pki.Name, string(*pki.CaCerts[0].Certificate), ""),
+			CloudConfigCksum: checksum,
 		}
 		*pkiData = append(*pkiData, pkiCacheObj)
 
@@ -655,6 +658,7 @@ func (c *AviObjCache) AviPopulateAllVSVips(client *clients.AviClient, cloud stri
 		}
 
 		var fqdns []string
+		var checksum string
 		for _, dnsinfo := range vsvip.DNSInfo {
 			fqdns = append(fqdns, *dnsinfo.Fqdn)
 		}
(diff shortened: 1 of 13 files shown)


sudswasavi (commented):
Couple of inline comments. Also we are directly making the label changes in the layer 3 code. This means that if the user's labels were changed due to some reason, the checksums of the existing objects won't change and the code won't be hit in Layer 3. Is this intended?

internal/lib/constants.go
```diff
@@ -74,6 +74,7 @@ const (
 	CertTypeCA                                 = "SSL_CERTIFICATE_TYPE_CA"
 	VSVIPDELCTRLVER                            = "20.1.1"
 	Advl4ControllerVersion                     = "20.1.2"
+	GRBACControllerVersion                     = "20.1.2"
```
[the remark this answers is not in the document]
akshayhavile: Intention behind using different variable is as these two are representing different feature, changes in base version of one should not affect other.  For GRBAC ,  min version should be 20.1.4 due to "delete issue" in GRBAC framework. I will take care of this.

internal/lib/lib.go
```diff
@@ -67,14 +67,29 @@ func SetDisableSync(state bool) {
 var AKOUser string
 
 func SetAKOUser() {
-	AKOUser = "ako-" + GetClusterName()
+	//Set AKO user same as login user if GRBAC enabled
+	if GetEnableGRBAC() {
+		AKOUser = os.Getenv("CTRL_USERNAME")
```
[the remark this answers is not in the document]
akshayhavile: By GRBAC with AKO design, customer has to make sure to use per cluster unique user. Objects from each cluster will be tagged with different labels. If user is duplicated across two cluster, until and unless that user is associated with labeled role which corresponds to clusters, it will not able to view or do REST operations on those objects at AVI Portal side.

akshayhavile (commented):
Here we are labelling objects at REST layer with "clustername:<K8Clustername>" ... so if user changes k8 clustername, then I think AKO will require to reboot to work properly and then we will pick up new labels as hash will change and will reflected at AVI side. Is there any other use case which can cause k8-clustername to change which I did not take into account and will not require AKO reboot.

sudswasavi:
> Here we are labelling objects at REST layer with "clustername:" ... so if user changes k8 clustername, then I think AKO will require to reboot to work properly and then we will pick up new labels as hash will change and will reflected at AVI side. Is there any other use case which can cause k8-clustername to change which I did not take into account and will not require AKO reboot.

This is not true @akshayhavile . The hash will not change if you directly make changes in the layer 3 node. cc @monotosh-avi

internal/lib/lib.go
```diff
@@ -386,8 +401,8 @@ func UseServicesAPI() bool {
 	return false
 }
 
-func CheckControllerVersionCompatibility(version, maxVersion string) bool {
-	if c, err := semver.NewConstraint(fmt.Sprintf("> %s", maxVersion)); err == nil {
+func CheckControllerVersionCompatibility(version, maxVersion, cmpSign string) bool {
```
shubhamavi, line 404: we can change this function to take arguments like
`CheckControllerVersionCompatibility(v1, compare, v2 string)` 
removing the `"max"Version`, and putting the  compare Sign in the middle so that the implementations are eassier to read like
`CheckControllerVersionCompatibility(currentVersion, ">=",  GRBACControllerVersion)`

monotosh-avi:
In case we upgrade the controller to 20.1.2, the objects won't be rebuilt. This is fine if we are only targeting new installations.

internal/cache/controller_obj_cache.go
```diff
@@ -1382,7 +1417,9 @@ func (c *AviObjCache) PopulateSSLKeyToCache(client *clients.AviClient, cloud str
 				cacert = ca.(string)
 			}
 		}
+		//TODO: Check with team : What is reason of recomputing checksum here again.
 		SslKeyData[i].CloudConfigCksum = lib.SSLKeyCertChecksum(SslKeyCacheObj.Name, SslKeyCacheObj.Cert, cacert)
```
akshayhavile, line 1421: @sudswasavi @monotosh-avi : What is reason of computing sslkeydata checksum here again.  We are doing it at AviPopulateAllSSLKeys() call? Shallow-cpy is reason?

akshayhavile:
Introduced checksum logic at layer 2 and addressed review comments

internal/lib/lib.go
```diff
@@ -481,6 +492,32 @@ func ShutdownApi() {
 	akoApi.ShutDown()
 }
 
+var clusterLabelChecksum uint32
+var clusterKey string
+var clusterValue string
+
+func SetClusterLabelChecksum() {
+	if GetEnableGRBAC() {
+		clusterKey = *GetLabels()[0].Key
```
monotosh-avi, line 501: key and values can be set directly or we can make one function call instead of two.

internal/rest/avi_obj_vs.go
```diff
@@ -991,6 +1007,10 @@ func (rest *RestOperations) AviVsVipCacheAdd(rest_op *utils.RestOp, vsKey avicac
 			vsvip_cache_obj.InvalidData = true
 		}
 
+		if resp["VsvipCloudConfigCksum"] != nil {
```
monotosh-avi, line 1010: @shubhamavi we were discussing about this. Probably it is okay to use VsvipCloudConfigCksum. But this would mean  one time PUT call. cc @sudswasavi

internal/cache/controller_obj_cache.go
monotosh-avi: To me, looks like we don't need this.

internal/rest/avi_obj_vs.go
sudswasavi: one time PUT call in existing deployments right? Maybe okay for now, we need to validate this with a test.

akshayhavile:
Addressed review comments
@chauhanshubham  : Can you please check whether it properly merged with  nsxalbinfra changes.

monotosh-avi (approved):

internal/cache/controller_obj_cache.go
```diff
@@ -988,6 +1007,9 @@ func (c *AviObjCache) AviPopulateOnePKICache(client *clients.AviClient,
 			continue
 		}
 		checksum := lib.SSLKeyCertChecksum(*pkikey.Name, *pkikey.CaCerts[0].Certificate, "")
+		if lib.GetEnableGRBAC() && pkikey.Labels != nil {
```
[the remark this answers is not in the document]
akshayhavile: @chauhanshubham: Each of this Checksum function have different calling points. For example, if we take into account "SSLKeyCertChecksum" ... it has been called while populating cache as well as at layer2 for computing checksum of object from k8 side. Similar scenarios for other component checksum calculations.  This will lead to complicated logic of checking calling point and then according to that add checksum value.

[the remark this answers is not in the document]
chauhanshubham: ok sure.

chauhanshubham (approved):

PR #262, opened 2021-02-03
Introduce L4 DNS formats etc

sudswasavi:
This commit intorduces the following changes:

- Allows an option to disable FQDN generation for L4 VSes.
- Allows specification of formats for FQDNs for L4.
- Allows specification of an annoation to generate custom FQDNs for L4 VSes.

Also removes the code that auto-generated the default FQDN for Shared VS for L4.

### DIFF
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -50,6 +50,7 @@ L7Settings:
 L4Settings:
   advancedL4: "false" # Use this knob to control the settings for the services API usage. Default to not using services APIs: https://github.com/kubernetes-sigs/service-apis
   defaultDomain: "" # If multiple sub-domains are configured in the cloud, use this knob to set the default sub-domain to use for L4 VSes.
+  autoFQDN: "<svc>.<ns>.<subdomain>" # ENUM: <svc>-<ns>.<subdomain>, <svc>.<ns>.<subdomain>, "false" If the value is false then the FQDN generation is disabled.
 
 ### This section outlines settings on the Avi controller that affects AKO's functionality.
 ControllerSettings:
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -50,7 +50,7 @@ L7Settings:
 L4Settings:
   advancedL4: "false" # Use this knob to control the settings for the services API usage. Default to not using services APIs: https://github.com/kubernetes-sigs/service-apis
   defaultDomain: "" # If multiple sub-domains are configured in the cloud, use this knob to set the default sub-domain to use for L4 VSes.
-  autoFQDN: "<svc>.<ns>.<subdomain>" # ENUM: <svc>-<ns>.<subdomain>, <svc>.<ns>.<subdomain>, "false" If the value is false then the FQDN generation is disabled.
+  autoFQDN: "default" # ENUM: default(<svc>.<ns>.<subdomain>), flat (<svc>-<ns>.<subdomain>), "disabled" If the value is disabled then the FQDN generation is disabled.
 
 ### This section outlines settings on the Avi controller that affects AKO's functionality.
 ControllerSettings:
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -47,9 +47,8 @@ var shardSizeMap = map[string]uint32{
 }
 
 var fqdnEnum = map[string]int32{
-	"<svc>.<ns>.<subdomain>": 1,
-	"<svc>-<ns>.<subdomain>": 2,
-	"false":                  3,
+	"default": 1,
+	"flat":    2,
 }
 
 var NamePrefix string
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -46,6 +46,12 @@ var shardSizeMap = map[string]uint32{
 	"SMALL":  1,
 }
 
+var fqdnEnum = map[string]int32{
+	"<svc>.<ns>.<subdomain>": 1,
+	"<svc>-<ns>.<subdomain>": 2,
+	"false":                  3,
+}
+
 var NamePrefix string
 
 func SetNamePrefix() {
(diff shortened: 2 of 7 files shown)


helm/ako/values.yaml
```diff
@@ -50,6 +50,7 @@ L7Settings:
 L4Settings:
   advancedL4: "false" # Use this knob to control the settings for the services API usage. Default to not using services APIs: https://github.com/kubernetes-sigs/service-apis
   defaultDomain: "" # If multiple sub-domains are configured in the cloud, use this knob to set the default sub-domain to use for L4 VSes.
+  autoFQDN: "<svc>.<ns>.<subdomain>" # ENUM: <svc>-<ns>.<subdomain>, <svc>.<ns>.<subdomain>, "false" If the value is false then the FQDN generation is disabled.
```
hemantavi, line 53: Should we instead make the enums as:
```
- "default" (<svc>.<ns>.<subdomain>)
- "withNamespace" (<svc>-<ns>.<subdomain>)
- "false"
```

internal/lib/lib.go
```diff
@@ -46,6 +46,12 @@ var shardSizeMap = map[string]uint32{
 	"SMALL":  1,
 }
 
+var fqdnEnum = map[string]int32{
```
hemantavi, line 49: Shall we use uint32 instead?

internal/lib/lib.go
```diff
@@ -46,6 +46,12 @@ var shardSizeMap = map[string]uint32{
...
+	"<svc>.<ns>.<subdomain>": 1,
+	"<svc>-<ns>.<subdomain>": 2,
+	"false":                  3,
```
hemantavi, line 52: Do we need the "false" key/value here, since we are anyway returning 3 (in `GetL4FqdnFormat`) in case of no match?

internal/nodes/avi_model_l4_translator.go
```diff
@@ -57,9 +64,21 @@ func (o *AviObjectGraph) ConstructAviL4VsNode(svcObj *corev1.Service, key string
...
 		// or would default to the first dns subdomain it gets from the dns profile
 		if strings.HasPrefix(subDomains[0], ".") {
-			fqdn = svcObj.ObjectMeta.Name + "." + svcObj.ObjectMeta.Namespace + subDomains[0]
+			if lib.GetL4FqdnFormat() == 1 {
+				// Generate the FQDN based on the logic: <svc_name>.<namespace>.<sub-domain>
+				fqdn = svcObj.ObjectMeta.Name + "." + svcObj.ObjectMeta.Namespace + subDomains[0]
+			} else if lib.GetL4FqdnFormat() == 2 {
+				// Generate the FQDN based on the logic: <svc_name>-<namespace>.<sub-domain>
+				fqdn = svcObj.ObjectMeta.Name + "-" + svcObj.ObjectMeta.Namespace + subDomains[0]
+			}
```
hemantavi, line 73: Can we move the above block to a separate functon to generate the fqdn and just call that function here and the below section to avoid duplication?

helm/ako/values.yaml
apalsule, line 53: This can have an empty value too, which will be equivalent to default ?

sudswasavi: yeah any value is in fact default. We are just specifying a `false` as way of being symbolic of the feature being turned off.

shubhamavi: yes, i think we should think around strings (all caps maybe), like the shardSize enums, SMALL/MEDIUM/LARGE
I also feel that with the annotation, this might just be adding additional complexity. 

Also, I might have not understood this properly but I see
`nil/"false" == <svc>.<ns>.<subdomain>`
So there are essentially two options: namespacedSubdomain (name.ns.com) and name-namespace separated fqdn (name-ns.com)
namespacedSubdomain is the current default so effectively this could be a bool?
If the only value add from the flag is name-namespace separated fqdn (name-ns.com) then that can actually be specified in the annotation itself?

internal/nodes/avi_model_l4_translator.go
```diff
@@ -57,9 +64,21 @@ func (o *AviObjectGraph) ConstructAviL4VsNode(svcObj *corev1.Service, key string
 		// subDomains[0] would either have the defaultSubDomain value
 		// or would default to the first dns subdomain it gets from the dns profile
 		if strings.HasPrefix(subDomains[0], ".") {
-			fqdn = svcObj.ObjectMeta.Name + "." + svcObj.ObjectMeta.Namespace + subDomains[0]
+			if lib.GetL4FqdnFormat() == 1 {
+				// Generate the FQDN based on the logic: <svc_name>.<namespace>.<sub-domain>
+				fqdn = svcObj.ObjectMeta.Name + "." + svcObj.ObjectMeta.Namespace + subDomains[0]
```
shubhamavi, line 69: `.ObjectMeta.` not required.

internal/lib/lib.go
sudswasavi: Done.

internal/nodes/avi_model_l4_translator.go
sudswasavi: I have optimised the code now.

internal/lib/lib.go
sudswasavi: any benefit?

helm/ako/values.yaml
sudswasavi: Done, called them 'default', 'flat' and 'disabled'

shubhamavi: got the difference between disabled and default thanks

internal/lib/lib.go
hemantavi: Just that the value can never be negative, so thought uint32 would be better.

shubhamavi (approved):

hemantavi (approved):

internal/lib/lib.go
sudswasavi: It's an internal constant and we can directly do comparisons for this number, if it's uint32 - we will have to cast and compare for example if lib.GetL4FqdnFormat() == uint32(3)

PR #264, opened 2021-02-04
Start/Run informers after setting up event handlers

shubhamavi:
We should be running informers after the event handlers have been set-up. Prior to this the order was reverse.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -212,7 +212,7 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 		fastretrywg, _ = waitGroupMap[0]["fastretry"]
 		slowretrywg, _ = waitGroupMap[0]["slowretry"]
 	}
-	c.Start(stopCh)
+
 	/** Sequence:
 	  1. Initialize the graph layer queue.
 	  2. Do a full sync from main thread and publish all the models.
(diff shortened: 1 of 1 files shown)


shubhamavi:
build ako

shubhamavi:
build ako

sudswasavi:
As discussed @monotosh-avi / @shubhamavi - it isn't entirely clear on what this PR is trying to solve. We are trading in a code path that is stable and may introduce regression, while I am not entirely denouncing this PR to be a good to have and not be the cause of any concern. Leave the decision to you folks.

monotosh-avi:
> As discussed @monotosh-avi / @shubhamavi - it isn't entirely clear on what this PR is trying to solve. We are trading in a code path that is stable and may introduce regression, while I am not entirely denouncing this PR to be a good to have and not be the cause of any concern. Leave the decision to you folks.

I think this change is fine, and to me it looks like the current change makes the sequence of operations logical.

monotosh-avi (approved):

shubhamavi:
> As discussed @monotosh-avi / @shubhamavi - it isn't entirely clear on what this PR is trying to solve. We are trading in a code path that is stable and may introduce regression. I am not entirely denouncing this PR and it may be a good to have and eventually may not be the cause of any concern. Leave the decision to you folks.

This is a pre-requisite to further configure informers (like adding cache.Indexers), which won't be possible if we run the informers first and then configure them. Configuring cache.Indexers for NsxAlbInfaSettings will be part of a subsequent PR, and this was intentionally kept separate due to the critical code under which it is present.

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

shubhamavi:
build ako

sudswasavi (approved):

PR #274, opened 2021-02-15
Automatically annotate Services for NPL

monotosh-avi:
AKO would automatically annotate required services for NPL. These
are Services of type LB and backends for ingresses.

- For Service type LB in check if the service is annotated, else
annotate and return.
- For ingress add, annotate all backend services and return.
- For ingress deletion check if annotation can be removed from bakcend service. This
is done after checking if the service is referred by any other ingress.
- For ingress update, check if any new service requires annotation, and if
annotation from old service can be be removed.

Helm template and operator changes to disable this feature by setting autoAnnotateService to false
would be added in next PR


monotosh-avi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

monotosh-avi:
build ako

internal/status/svc_annotation.go
```diff
@@ -0,0 +1,85 @@
...
+	patchPayload := map[string]interface{}{
+		"metadata": map[string]map[string]string{
+			"annotations": ann,
+		},
+	}
+
+	payloadBytes, _ := json.Marshal(patchPayload)
+	_, err = utils.GetInformers().ClientSet.CoreV1().Services(service.Namespace).Patch(context.TODO(), service.Name, types.MergePatchType, payloadBytes, metav1.PatchOptions{})
+	if err != nil {
+		utils.AviLog.Errorf("key: %s, msg: there was an error in updating the Service annotation for NPL: %v", key, err)
```
sudswasavi, line 54: you should probably return here? Else the log below will get printed.

internal/status/svc_annotation.go
```diff
@@ -0,0 +1,85 @@
...
```
sudswasavi, line 82: Same here, if it's errors out - we should return.

internal/status/svc_annotation.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	ann[lib.NPLSvcAnnotation] = "true"
+	patchPayload := map[string]interface{}{
+		"metadata": map[string]map[string]string{
+			"annotations": ann,
+		},
+	}
+
+	payloadBytes, _ := json.Marshal(patchPayload)
+	_, err = utils.GetInformers().ClientSet.CoreV1().Services(service.Namespace).Patch(context.TODO(), service.Name, types.MergePatchType, payloadBytes, metav1.PatchOptions{})
```
sudswasavi, line 52: Can listers get used instead of the clientset?

internal/status/svc_annotation.go
monotosh-avi: Added

internal/status/svc_annotation.go
monotosh-avi: I don't think we can use lister, that can be used to get services only.

internal/status/svc_annotation.go
monotosh-avi: Done

sudswasavi (approved):

tests/npltests/npl_pod_test.go
```diff
@@ -12,7 +12,7 @@
 * limitations under the License.
 */
 
-package hostnameshardtests
+package npltests
```
shubhamavi, line 15: could you please rebase once, i remember doing this change and some addition regarding CRDClientset in a previous PR

internal/status/svc_annotation.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		utils.AviLog.Errorf("key: %s, msg: there was an error in updating the Service annotation for NPL: %v", key, err)
+		return
+	}
+	utils.AviLog.Debugf("key: %s, msg: Deleted NPL annotation from Service: %s/%s", namespace, name)
```
shubhamavi, line 86: key is missing in the formatting argument.

internal/status/svc_annotation.go
```diff
@@ -0,0 +1,87 @@
...
+	service, err := utils.GetInformers().ServiceInformer.Lister().Services(namespace).Get(name)
+	if err != nil {
+		return
+	}
+	ann := service.GetAnnotations()
+	if _, found := ann[lib.NPLSvcAnnotation]; !found {
+		return
+	}
+	if ann == nil {
+		ann = make(map[string]string)
```
shubhamavi, line 71: can we return from here?

internal/nodes/ingress_model_rel.go
```diff
@@ -307,10 +313,28 @@ func IngressChanges(ingName string, namespace string, key string) ([]string, boo
...
-			utils.AviLog.Debugf("key: %s, msg: updating ingress relationship for service: %s", key, svc)
+		_, oldSvcs := objects.SharedSvcLister().IngressMappings(namespace).GetIngToSvc(ingName)
+		currSvcs := parseServicesForIngress(ingObj.Spec, key)
+
+		svcToDel := lib.Difference(oldSvcs, currSvcs)
+		for _, svc := range svcToDel {
+			_, ingrforSvc := objects.SharedSvcLister().IngressMappings(namespace).GetSvcToIng(svc)
+			ingrforSvc = utils.Remove(ingrforSvc, ingName)
+			//objects.SharedSvcLister().IngressMappings(namespace).UpdateSvcToIngMapping(svc, ingrforSvc)
+			if len(ingrforSvc) == 0 {
```
shubhamavi, line 324: should we be checking for AutoAnnotateNPLSvc() here as well? also let's remove the commented out code above.

internal/nodes/ingress_model_rel.go
```diff
@@ -291,7 +292,12 @@ func IngressChanges(ingName string, namespace string, key string) ([]string, boo
 		if k8serrors.IsNotFound(err) {
 			// Remove all the Ingress to Services mapping.
 			// Remove the references of this ingress from the Services
-			objects.SharedSvcLister().IngressMappings(namespace).RemoveIngressMappings(ingName)
+			svcToDel := objects.SharedSvcLister().IngressMappings(namespace).RemoveIngressMappings(ingName)
```
shubhamavi, line 295: this can move inside the conditional below

internal/lib/lib.go
```diff
@@ -550,10 +550,20 @@ func IsNodePortMode() bool {
...
+// ToDo: Set the Service Type only once. But this creates a problem in UTs,
+// because different types of Services needs to be tested in the UTs.
 func GetServiceType() string {
 	return os.Getenv(SERVICE_TYPE)
 }
 
+func AutoAnnotateNPLSvc() bool {
+	autoAnnotateSvc := os.Getenv(autoAnnotateService)
+	if GetServiceType() == NodePortLocal && autoAnnotateSvc != "false" {
+		return true
```
shubhamavi, line 562: this will return true if the user passes the value as "False".

monotosh-avi: corrected to do case insensitive match

internal/nodes/ingress_model_rel.go
monotosh-avi: We need to remove the mapping even if NPL is not enabled, so the function call has to be made, the return value can be ignored.

internal/nodes/ingress_model_rel.go
monotosh-avi: Good catch ! Added the condition.

internal/status/svc_annotation.go
monotosh-avi: Yeah, we can. Added.

internal/status/svc_annotation.go
monotosh-avi: Added

tests/npltests/npl_pod_test.go
monotosh-avi: I think that PR is not merged yet :-)

sudswasavi (approved):

tests/npltests/npl_pod_test.go
shubhamavi: ah, i'll rebase then :) thanks

shubhamavi (approved):

PR #277, opened 2021-02-22
Adding vipNetworkList to config for multivip support

aaha97:


sudswasavi:
This PR will directly replace values on Layer 3 of AKO. This is not recommended. Imagine this:
 - Someone was using AKO 1.3.1
 - They upgraded to AKO 1.4.1
 - They specified networkVipList in the configMap to use the multi-vip feature.
 - None of the VS checksums would change on reboot/upgrade of AKO.
 - None of the VSes would get updated.

So you need to move this logic to layer 2 code of AKO and affect the checksum for it to work.

sudswasavi:
Also - you have made changes for the AKO operator in the same PR. So it would require another code path for testing the installation, I suspect more changes are needed for these fields to work because you have added a struct in the AKOConfig, this would require new `deep copy` methods cc @hemantavi

hemantavi:
>  I suspect more changes are needed for these fields to work because you have added a struct in the AKOConfig, this would require new deep copy methods

Yes, I see some changes to the CRD too @aaha97 . Were they generated from the types.go file or manually? If manually, can you please remove those fields from the CRD's .yaml file and generate the controller code again?

internal/nodes/avi_model_nodes.go
```diff
@@ -761,11 +761,7 @@ func (v *AviVSVIPNode) GetCheckSum() uint32 {
 
 func (v *AviVSVIPNode) CalculateCheckSum() {
 	var checksum uint32
-	if v.NetworkName != nil {
-		checksum = lib.VSVipChecksum(v.FQDNs, v.IPAddress, *v.NetworkName)
-	} else {
-		checksum = lib.VSVipChecksum(v.FQDNs, v.IPAddress, "")
-	}
+	checksum = lib.VSVipChecksum(v.FQDNs, v.IPAddress, v.NetworkNames)
```
sudswasavi, line 764: if you simply do this, then the checksum will change even if the NetworkNames slice is empty and this will re-sync all vips. Note: Empty slices does not mean a 0 checksum.

sudswasavi (commented):
@monotosh-avi can you PTAL.

sudswasavi:
Also - what is your upgrade strategy? If someone uses 1.3.3 and migrate to 1.4.1 - how is this code supposed to work if they didn't use the new `vipNetworkList` parameter?

internal/lib/lib.go
```diff
@@ -620,8 +651,8 @@ func VSVipChecksum(FQDNs []string, IPAddress string, networkName string) uint32
 	if IPAddress != "" {
 		checksum += utils.Hash(IPAddress)
 	}
-	if networkName != "" {
-		checksum += utils.Hash(networkName)
+	if len(networkNames) != 0 {
+		checksum = utils.Hash(utils.Stringify(networkNames))
```
aaha97, line 655: @sudswasavi i believe this check takes care of the case when the vipNetworkList is empty.

monotosh-avi: Does this mean that if fqdn changes, VSVIP won't get updated ? I think that would be a problem.

monotosh-avi (dismissed):

aaha97:
> Also - what is your upgrade strategy? If someone uses 1.3.3 and migrate to 1.4.1 - how is this code supposed to work if they didn't use the new `vipNetworkList` parameter?

If the vipNetworkList is not used, then the checksum shouldn't change as only the first and only element of the list is picked and hashed as a string.

internal/lib/lib.go
aaha97: seems to be a typo, it should be "+="
fixed it.

helm/ako/values.yaml
```diff
@@ -40,6 +40,11 @@ NetworkSettings:
   subnetPrefix: "" # Subnet Prefix of the vip network
   networkName: "" # Network Name of the vip network
   enableRHI: false # This is a cluster wide setting for BGP peering.
+  vipNetworkList: []
+  #vipNetworkList: []
+  #  - networkName: subnet-xxxxxxx
+  #  - subnetIP: 12.0.0.0
```
sudswasavi, line 46: This "-" is probably not required? Since you will have one map of elements within a slice?

aaha97: the user can either provide the networkName OR subnetIP, subnetPrefix pair. The vipNetworkList in the example has 2 elements, one of each.

sudswasavi (commented):
- if networkName is set, ignore vipNetworkList ==> No cloudType check, works like present day.
- if networkName is not set, look at vipNetworkList and check if cloudType == AWS, else restart AKO ==> This implies that currently vipNetworkList is only supported for AWS for all other clouds, the user has to specify networkName.

In Layer 2 code, we must ensure that networkName is being populated into the vsvip model from the environment variable.
Fetch this  value in layer 3, instead of relying on the environment variable. Trust the layer 2 model in Layer 3.

chauhanshubham:
looks like AviInfraSetting UTs also need updates. Could you please modify the tests to check for NetworkName as a slice there as well.

internal/cache/controller_obj_cache.go
```diff
@@ -1246,13 +1250,17 @@ func (c *AviObjCache) AviPopulateOneVsVipCache(client *clients.AviClient,
...
 		for _, vip := range vsvip.Vip {
 			vips = append(vips, *vip.IPAddress.Addr)
 			if ipamNetworkSubnet := vip.IPAMNetworkSubnet; ipamNetworkSubnet != nil {
 				if networkRef := *ipamNetworkSubnet.NetworkRef; networkRef != "" {
-					if networRefName := strings.Split(networkRef, "#"); len(networRefName) == 2 {
-						networkName = networRefName[1]
+					if networkRefName := strings.Split(networkRef, "#"); len(networkRefName) == 2 {
+						networkNames = append(networkNames, networkRefName[1])
+					}
+					if lib.GetCloudType() == lib.CLOUD_AWS {
```
monotosh-avi, line 1261: Is there a chance that in AWS we may add two entries for the same network ? Both of the conditions would satisfy.

internal/rest/avi_obj_vsvip.go
```diff
@@ -497,3 +496,18 @@ func (rest *RestOperations) AviVsVipCacheDel(rest_op *utils.RestOp, vsKey avicac
 
 	return nil
 }
+
+func networkNamesToVips(networkNames []string) []*avimodels.Vip {
+	var vipList []*avimodels.Vip
+	autoAllocate := true
+	for vipIDInt, networkName := range networkNames {
+		vipID := strconv.Itoa(vipIDInt + 2)
```
monotosh-avi, line 504: Just a query: Any reason you had to add 2 here ?

internal/cache/controller_obj_cache.go
aaha97: multiple entries of the same network are allowed and will end up creating multiple VIPs in the same network.

internal/rest/avi_obj_vsvip.go
aaha97: vip ID starts from index 1 and not 0.
earlier argument was that first vip will be created using networkname.
updated it to  "+ 1" instead accordingly

aaha97:
build ako

monotosh-avi (approved):

chauhanshubham:
build ako

aaha97:
build ako

aaha97:
build ako

monotosh-avi (approved):

chauhanshubham (approved):

PR #278, opened 2021-02-23
Prevent object references created by current AKO instance

chauhanshubham:
AKO CRDs should not allow users to refer to Avi Objects that have been previously created by the AKO instance running in the same cluster. While checking for object presence in Avi, this checks for created_by field, which should not be equal to AKO user.
Additional commit that solves a UT failure seen on a previous run. http://10.79.111.162:8080/job/ako_OS-pr-builder/827

### DIFF
--- a/internal/nodes/crd_translator.go
+++ b/internal/nodes/crd_translator.go
@@ -21,8 +21,6 @@ import (
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


monotosh-avi (approved):

sudswasavi (approved):
This requires FT validation @anish-avi

PR #280, opened 2021-02-24
Delete Servers for Service type LB when Pod is deleted

monotosh-avi:
- Added a new object lister SharedPodToLBSvcLister to track this
- For Pod addition this would get updated
- When a Pod is deleted services would be fetched from this lister


sudswasavi (dismissed):

monotosh-avi:
build ako

sudswasavi (approved):

hemantavi (approved):

PR #281, opened 2021-02-25
Update vs uuids in the object annotations

hemantavi:
Piggybacks on the objects' status updates. Once a call to update the
status for a route/ingress/LB service is made, first the status is
updated and then the annotations are updated.
- Sets the UUID for the virtualservice to which the object is mapped
  to, in the `UpdateOptions`.
- Call to update the annotations is made. For route/ingress annotations
  updates:
  * Update the VS uuid for the hostname in the current update context.
  * Remove the non-required hostnames from the annotations.
  * Check if a patch is required (by comparing with the previous
    annotations).

  For LB service annotation updates:
  * Just update the VS uuid for the hostname.
  * Compare the new annotation with the previous one and update if
    required.

This PR also fetches the controller cluster's UUID and sets it in the
objects' annotations field.

Example of both of these annotations:
```
metadata:
  annotations:
    ako.vmware.com/controller-cluster-uuid: cluster-XXXXX
    ako.vmware.com/virtualservices: '{"host1.fqdn":"virtualservice-XXXXXX","host2.fqdn":"virtualservice-YYYYYYYY"}'
```

Added an additional permission for services in the clusterrole.

### DIFF
--- a/internal/status/ing_status.go
+++ b/internal/status/ing_status.go
@@ -42,7 +42,7 @@ type UpdateOptions struct {
 
 const (
 	VSAnnotation         = "ako.vmware.com/virtualservices"
-	ControllerAnnotation = "ako.vmware.com/controller-uuid"
+	ControllerAnnotation = "ako.vmware.com/controller-cluster-uuid"
 )
 
 // VSUuidAnnotation is maps a hostname to the UUID of the virtual service where it is placed.
@@ -239,6 +239,7 @@ func getAnnotationsPayload(vsAnnotations map[string]string, existingAnnotations 
 		existingAnnotations = make(map[string]string)
 	}
 	existingAnnotations[VSAnnotation] = string(vsAnnotationsStr)
+	existingAnnotations[ControllerAnnotation] = avicache.GetControllerClusterUUID()
 	patchPayload := map[string]interface{}{
 		"metadata": map[string]map[string]string{
 			"annotations": existingAnnotations,
--- a/internal/status/ing_status.go
+++ b/internal/status/ing_status.go
@@ -26,6 +26,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	networkingv1beta1 "k8s.io/api/networking/v1beta1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	types "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -41,7 +42,7 @@ type UpdateOptions struct {
 }
 
 const (
-	VSAnnotation         = "ako.vmware.com/virtualservices"
+	VSAnnotation         = "ako.vmware.com/host-fqdn-vs-uuid-map"
 	ControllerAnnotation = "ako.vmware.com/controller-cluster-uuid"
 )
 
@@ -146,8 +147,8 @@ func updateObject(mIngress *networkingv1beta1.Ingress, updateOption UpdateOption
 	return nil
 }
 
-func updateIngAnnotations(mClient kubernetes.Interface, ingObj *networkingv1beta1.Ingress, svcMetaHostnames []string,
-	vsUUID, key string, ingHostList []string, oldIng *networkingv1beta1.Ingress, retryNum ...int) error {
+func updateIngAnnotations(mClient kubernetes.Interface, ingObj *networkingv1beta1.Ingress, hostnamesToBeUpdated []string,
+	vsUUID, key string, ingSpecHostnames []string, oldIng *networkingv1beta1.Ingress, retryNum ...int) error {
 
 	if ingObj == nil {
 		ingObj = oldIng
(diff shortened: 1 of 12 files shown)


hemantavi:
build ako

sudswasavi:
Pls rebase this PR to the latest master. I think the UTs will pass.

hemantavi:
Rebased to the latest master.

hemantavi:
build ako

internal/status/ing_status.go
```diff
@@ -105,28 +115,155 @@ func updateObject(mIngress *networkingv1beta1.Ingress, updateOption UpdateStatus
...
+		return nil
+	}
+	if err = patchIngressAnnotations(ingObj, vsAnnotations, mClient); err != nil {
+		utils.AviLog.Errorf("key: %s, msg: there was an error in updating the ingress annotations: %v", key, err)
 		// fetch updated ingress and feed for update status
-		mIngresses := getIngresses([]string{mIngress.Namespace + "/" + mIngress.Name}, false)
+		mIngresses := getIngresses([]string{ingObj.Namespace + "/" + ingObj.Name}, false)
 		if len(mIngresses) > 0 {
-			return updateObject(mIngresses[mIngress.Namespace+"/"+mIngress.Name], updateOption, retry+1)
+			return updateIngAnnotations(mClient, mIngresses[ingObj.Namespace+"/"+ingObj.Name], svcMetaHostnames, vsUUID, key, ingHostList, oldIng, retry+1)
```
sudswasavi, line 196: Possibly another legacy code problem but if the `err` above resulted into `NotFound` that is the Ingress you are trying to update has been deleted, there's no point of recursing the call over 3 loops and we could just quit the loop and return.

internal/status/ing_status.go
```diff
@@ -105,28 +115,155 @@ func updateObject(mIngress *networkingv1beta1.Ingress, updateOption UpdateStatus
...
+	if !ok {
+		if len(newVSAnnotations) > 0 {
+			return true
+		}
+		return false
+	}
+
+	var oldVSAnnotations map[string]string
+	if err := json.Unmarshal([]byte(oldVSAnnotationsStr), &oldVSAnnotations); err != nil {
+		utils.AviLog.Errorf("error in unmarshalling old vs annotations %s: %v", oldVSAnnotationsStr, err)
```
sudswasavi, line 214: If you weren't able to unmarshall the annotations, could there be a bigger problem here and you fail the update instead of returning true?

internal/status/ing_status.go
```diff
@@ -105,28 +115,155 @@ func updateObject(mIngress *networkingv1beta1.Ingress, updateOption UpdateStatus
...
+		return true
+	}
+
+	if len(oldVSAnnotations) != len(newVSAnnotations) {
+		return true
+	}
+	for oldHost, oldVS := range oldVSAnnotations {
+		newVS, exists := newVSAnnotations[oldHost]
+		if !exists {
```
sudswasavi, line 223: Minor nit: the two if blocks can be combined into 1.

internal/status/ing_status.go
```diff
@@ -193,28 +330,86 @@ func deleteObject(svc_mdata_obj avicache.ServiceMetadataObj, key string, isVSDel
...
 	}
 
-	patchPayload, _ := json.Marshal(map[string]interface{}{
-		"status": mIngress.Status,
-	})
-	if len(mIngress.Status.LoadBalancer.Ingress) == 0 {
-		patchPayload, _ = json.Marshal(map[string]interface{}{
-			"status": nil,
-		})
+	if err = deleteIngressAnnotation(updatedIng, svc_mdata_obj, isVSDelete, hostListIng, key, mClient, mIngress); err != nil {
```
sudswasavi, line 359: Again can this not be done within the same call above?

internal/status/ing_status.go
```diff
@@ -105,28 +115,155 @@ func updateObject(mIngress *networkingv1beta1.Ingress, updateOption UpdateStatus
...
+		if err := json.Unmarshal([]byte(value), &vsAnnotations); err != nil {
+			// just print an error and continue, this will be taken care of during the update
+			utils.AviLog.Errorf("key: %s, error in unmarshalling Ingress %s/%s annotations for VS: %v",
+				key, ingObj.Namespace, ingObj.Name, err)
+		}
+	}
+
+	// update the existing hostname vs uuid for the current update
+	for i := 0; i < len(svcMetaHostnames); i++ {
+		vsAnnotations[svcMetaHostnames[i]] = vsUUID
```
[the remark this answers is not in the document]
monotosh-avi: I think svcMetaHostnames contain service metadata from same VS .. so this should be fine.

internal/status/ing_status.go
hemantavi: Sure, will add that check here.

internal/status/ing_status.go
hemantavi: The annotations wouldn't be in proper format in this case, and hence, it would be better to return true and update the annotations I think.

internal/status/ing_status.go
hemantavi: Will update. Thanks.

monotosh-avi (approved):

sudswasavi (approved):

PR #283, opened 2021-02-25
Fix for problem in multiple UTs

monotosh-avi:
- deleting multiple models at the start of the tests which uses multiple Hosts


chauhanshubham (approved):

chauhanshubham:
thanks @monotosh-avi

sudswasavi (approved):
Great catch. So some stale models were left in memory and needed cleanup.

PR #284, opened 2021-02-26
fix ingressclass template for defaultIngController

chauhanshubham:
`defaultIngController` in helm values.yaml is of type string. While we add the default ingress class annotation `ingressclass.kubernetes.io/is-default-class` assuming that the value is a boolean. This ends up making Avi IngressClass default even if the value set is `"false"`.

### DIFF
--- a/helm/ako/templates/ingressclass.yaml
+++ b/helm/ako/templates/ingressclass.yaml
@@ -3,7 +3,7 @@ apiVersion: networking.k8s.io/v1
 kind: IngressClass
 metadata:
   name: avi-lb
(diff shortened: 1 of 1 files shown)


sudswasavi (approved):

chauhanshubham:
This is a bit confusing since `defaultIngController` in AkoConfig is of type boolean, but of type string in helm `values.yaml`.
But since there is no validation on the `values.yaml` file, it would be better to keep the type as is.

monotosh-avi (approved):

PR #289, opened 2021-03-02
Fix problem in deleting NPL annotation from Service

monotosh-avi:
- Setting value for NPL annotation key to nil
- Instead of returning podLBSvcInstance podSvcInstance was being returned in SharedPodToLBSvcLister, fixed this.

### DIFF
--- a/internal/status/svc_annotation.go
+++ b/internal/status/svc_annotation.go
@@ -72,9 +72,6 @@ func DeleteSvcAnnotation(key, namespace, name string) {
 	}
 
 	payloadValue := make(map[string]*string)
-	for key, val := range ann {
-		payloadValue[key] = &val
-	}
 	// To delete an annotation with patch call, the value has to be set to nil
 	payloadValue[lib.NPLSvcAnnotation] = nil
 
--- a/internal/status/svc_annotation.go
+++ b/internal/status/svc_annotation.go
@@ -54,7 +54,7 @@ func CheckUpdateSvcAnnotation(key, namespace, name string) bool {
 		utils.AviLog.Errorf("key: %s, msg: there was an error in updating the Service annotation for NPL: %v", key, err)
 		return false
 	}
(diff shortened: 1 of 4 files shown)


internal/status/svc_annotation.go
```diff
@@ -70,18 +70,25 @@ func DeleteSvcAnnotation(key, namespace, name string) {
...
 	}
-	delete(ann, lib.NPLSvcAnnotation)
-	patchPayload := map[string]interface{}{
-		"metadata": map[string]map[string]string{
-			"annotations": ann,
+
+	payloadValue := make(map[string]*string)
+	for key, val := range ann {
+		payloadValue[key] = &val
+	}
```
hemantavi, line 77: Any reason you are setting back the old annotations to the service? For deleting a specific annotation, this is not required, the statement below will do, as this is not a `replace` operation but a `merge` operation.

internal/status/svc_annotation.go
```diff
@@ -54,7 +54,7 @@ func CheckUpdateSvcAnnotation(key, namespace, name string) bool {
 		utils.AviLog.Errorf("key: %s, msg: there was an error in updating the Service annotation for NPL: %v", key, err)
 		return false
 	}
-	utils.AviLog.Debugf("key: %s, msg: updated NPL annotation from Service: %s/%s", key, namespace, name)
+	utils.AviLog.Infof("key: %s, msg: updated NPL annotation from Service: %s/%s", key, namespace, name)
```
hemantavi, line 57: Do we really need an Infof for this log? Its completely your call though.

internal/status/svc_annotation.go
monotosh-avi: Thanks ! I thought it would replace the old annotations. I'll remove it.

internal/status/svc_annotation.go
monotosh-avi: I would prefer to keep this, else we would have no idea about the status. I think we do this for some other status updates too.

hemantavi (approved):

sudswasavi (approved):

PR #290, opened 2021-03-04
fix default ingressclass transitioning issues

chauhanshubham:
Changing the default ingress class annotation in Avi ingress class was not removing the VSes/Pools from Avi and does not remove status from the ingress objects. This PR fixes this issue.

This also includes a related ingress status update fix wherein the status is removed only when host is removed from the spec. But with new additions of namespace selector and ingress selection via ingress class, the status can be removed from ingress even when the host is still present in the spec. The check for the same has been removed from the workflow.

### DIFF
--- a/internal/nodes/ingress_model_rel.go
+++ b/internal/nodes/ingress_model_rel.go
@@ -300,23 +300,15 @@ func IngressChanges(ingName string, namespace string, key string) ([]string, boo
 					status.DeleteSvcAnnotation(key, namespace, svc)
 				}
 			}
-			objects.SharedSvcLister().IngressMappings(metav1.NamespaceAll).RemoveIngressClassMappings(namespace + "/" + ingName)
+			if utils.GetIngressClassEnabled() {
+				objects.SharedSvcLister().IngressMappings(metav1.NamespaceAll).RemoveIngressClassMappings(namespace + "/" + ingName)
+			}
 		}
 	} else {
-
 		// simple validator check for duplicate hostpaths, logs Warning if duplicates found
 		success := validateSpecFromHostnameCache(key, ingObj.Namespace, ingObj.Name, ingObj.Spec)
 		if !success {
 			return ingresses, false
-		}
-
-		if ingObj.Spec.IngressClassName != nil {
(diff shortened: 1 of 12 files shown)


sudswasavi:
Pls put some meaningful description on the commit message.

chauhanshubham:
> Pls put some meaningful description on the commit message.
yes i'll do that and add reviewers later

sudswasavi:
@chauhanshubham shall we cherry-pick this to release-1.3.4?

chauhanshubham:
build ako

akshayhavile:
@chauhanshubham : I am bit confused about files getting compared here. For example, ing_status.go doesn't seems to be upto date as present in master. Annotation related code, I am not able to find here. Am i missing something?

chauhanshubham:
build ako

chauhanshubham:
build ako

internal/nodes/ingress_model_rel.go
```diff
@@ -311,6 +311,9 @@ func IngressChanges(ingName string, namespace string, key string) ([]string, boo
 
 		if ingObj.Spec.IngressClassName != nil {
 			objects.SharedSvcLister().IngressMappings(metav1.NamespaceAll).UpdateIngressClassMappings(namespace+"/"+ingName, *ingObj.Spec.IngressClassName)
+		} else if aviIngClassName, found := isAviLBDefaultIngressClass(); found {
```
monotosh-avi, line 314: I wonder what would happen if the ingress class is added after adding the ingress. Also take a look at this: 
https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/299
There might be come conflicts.

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

sudswasavi (approved):
@monotosh-avi / @akshayhavile  PTAL and merge.

monotosh-avi (approved):

akshayhavile (approved):

PR #292, opened 2021-03-05
L4 Namespace Sync

akshayhavile:
This PR addresses:
1) L4 Namespace Sync: Sync L4 services from valid namespace with AVI controller. Do not address advance L4
2) Code optimization.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -274,7 +274,7 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 		err = c.FullSyncK8s()
 		if err != nil {
 			// Something bad sync. We need to return and shutdown the API server
-			utils.AviLog.Errorf("Couldn't run full sync successfully on bootup, going to shutdown AKO: %s", err)
+			utils.AviLog.Errorf("Couldn't run full sync successfully on bootup, going to shutdown AKO")
 			lib.ShutdownApi()
 			return
 		}
@@ -389,20 +389,26 @@ func (c *AviController) FullSyncK8s() error {
 	if err != nil {
 		utils.AviLog.Errorf("Unable to retrieve the services during full sync: %s", err)
 		return err
-	} else {
-		for _, svcObj := range svcObjs {
-			isSvcLb := isServiceLBType(svcObj)
-			var key string
-			if isSvcLb && !lib.GetLayer7Only() {
-				key = utils.L4LBService + "/" + utils.ObjKey(svcObj)
-			} else {
-				if lib.GetAdvancedL4() {
-					continue
-				}
-				key = utils.Service + "/" + utils.ObjKey(svcObj)
+	}
+	for _, svcObj := range svcObjs {
+
+		isSvcLb := isServiceLBType(svcObj)
+		var key string
+		if isSvcLb && !lib.GetLayer7Only(){
+			svcLabel := utils.ObjKey(svcObj)
+			ns := strings.Split(svcLabel, "/")
+			//Add key if ServiceAPI or advance L4 enabled or NS is valid
+			if !utils.IsServiceNSValid(ns[0]) {
+				continue
 			}
-			nodes.DequeueIngestion(key, true)
+			key = utils.L4LBService + "/" + utils.ObjKey(svcObj)
+		} else {
+			if lib.GetAdvancedL4() {
+				continue
+			}
+			key = utils.Service + "/" + utils.ObjKey(svcObj)
 		}
+		nodes.DequeueIngestion(key, true)
 	}
 
 	if lib.GetServiceType() == lib.NodePortLocal {
@@ -457,8 +463,8 @@ func (c *AviController) FullSyncK8s() error {
 				for _, ingObj := range ingObjs {
 					ingLabel := utils.ObjKey(ingObj)
 					ns := strings.Split(ingLabel, "/")
-					if utils.CheckIfNamespaceAccepted(ns[0], utils.GetGlobalNSFilter(), nil, true) {
-						key := utils.Ingress + "/" + ingLabel
+					if utils.CheckIfNamespaceAccepted(ns[0], nil, true) {
+						key := utils.Ingress + "/" + utils.ObjKey(ingObj)
 						utils.AviLog.Debugf("Dequeue for ingress key: %v", key)
 						nodes.DequeueIngestion(key, true)
 					}
@@ -476,8 +482,8 @@ func (c *AviController) FullSyncK8s() error {
 					// to do move to container-lib
 					routeLabel := utils.ObjKey(routeObj)
 					ns := strings.Split(routeLabel, "/")
-					if utils.CheckIfNamespaceAccepted(ns[0], utils.GetGlobalNSFilter(), nil, true) {
-						key := utils.OshiftRoute + "/" + routeLabel
+					if utils.CheckIfNamespaceAccepted(ns[0], nil, true) {
+						key := utils.OshiftRoute + "/" + utils.ObjKey(routeObj)
 						utils.AviLog.Debugf("Dequeue for route key: %v", key)
 						nodes.DequeueIngestion(key, true)
 					}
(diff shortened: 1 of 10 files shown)


internal/k8s/ako_init.go
```diff
@@ -387,20 +387,26 @@ func (c *AviController) FullSyncK8s() error {
...
-				key = utils.Service + "/" + utils.ObjKey(svcObj)
+	}
+	for _, svcObj := range svcObjs {
+
+		isSvcLb := isServiceLBType(svcObj)
+		var key string
+		if isSvcLb {
+			svcLabel := utils.ObjKey(svcObj)
+			ns := strings.Split(svcLabel, "/")
+			//Add key if ServiceAPI or advance L4 enabled or NS is valid
```
monotosh-avi, line 398: I think this comment can be reworded. Now it seems like add key if( ServiceAPI) or (advance L4 enabled) or (NS is valid).

internal/k8s/controller.go
```diff
@@ -190,6 +190,29 @@ func AddRoutesFromNSToIngestionQueue(numWorkers uint32, c *AviController, namesp
...
+}
+func AddServicesFromNSToIngestionQueue(numWorkers uint32, c *AviController, namespace string, msg string) {
+	// For Advancd L4 and service api , do not handle. services already been taken care
+	// in service handler
+	if lib.GetAdvancedL4() || lib.UseServicesAPI() {
+		return
+	}
+	svcObjs, err := utils.GetInformers().ServiceInformer.Lister().Services(namespace).List(labels.Set(nil).AsSelector())
+	if err != nil {
+		utils.AviLog.Errorf("Unable to retrieve the services during full sync: %s", err)
```
monotosh-avi, line 202: This comment has to be changed.

internal/k8s/ako_init.go
```diff
@@ -455,8 +461,8 @@ func (c *AviController) FullSyncK8s() error {
 				for _, ingObj := range ingObjs {
 					ingLabel := utils.ObjKey(ingObj)
 					ns := strings.Split(ingLabel, "/")
-					if utils.CheckIfNamespaceAccepted(ns[0], utils.GetGlobalNSFilter(), nil, true) {
-						key := utils.Ingress + "/" + ingLabel
+					if utils.CheckIfNamespaceAccepted(ns[0], nil, true) {
```
monotosh-avi, line 464: Why is NS filter not required anymore ?

internal/k8s/ako_init.go
akshayhavile: will change it.

internal/k8s/ako_init.go
akshayhavile: Now CheckIfNamespaceAccepted() function directly access NSFilter object present in same file(utils). Old code was un-necessary making an extra function call to fetch that variable(through utils.GetGlobalNSFilter()).

internal/k8s/controller.go
akshayhavile: Will change it.

internal/k8s/ako_init.go
```diff
@@ -272,7 +272,7 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 		err = c.FullSyncK8s()
 		if err != nil {
 			// Something bad sync. We need to return and shutdown the API server
-			utils.AviLog.Errorf("Couldn't run full sync successfully on bootup, going to shutdown AKO: %s", err)
+			utils.AviLog.Errorf("Couldn't run full sync successfully on bootup, going to shutdown AKO")
```
sudswasavi, line 275: why did you remove the error?

internal/k8s/ako_init.go
```diff
@@ -387,20 +387,26 @@ func (c *AviController) FullSyncK8s() error {
...
+	}
+	for _, svcObj := range svcObjs {
+
+		isSvcLb := isServiceLBType(svcObj)
+		var key string
+		if isSvcLb {
+			svcLabel := utils.ObjKey(svcObj)
+			ns := strings.Split(svcLabel, "/")
+			//Add key if ServiceAPI or advance L4 enabled or NS is valid
+			if !utils.IsServiceNSValid(ns[0]) {
```
sudswasavi, line 399: technically you can even stop the service of ClusterIP from getting processed because the namespace is not eligible right? I understand the ingress object won't be processed so we are good that way but this could be another optimisation.

internal/k8s/ako_init.go
akshayhavile: In FullSyncK8s(), we are printing out error and returning it to caller.. So to avoid same error getting printed twice, thought of removing it.

internal/k8s/controller.go
```diff
@@ -190,6 +190,29 @@ func AddRoutesFromNSToIngestionQueue(numWorkers uint32, c *AviController, namesp
 		utils.AviLog.Debugf("key: %s, msg: %s for namespace: %s", key, msg, namespace)
 	}
 
+}
+func AddServicesFromNSToIngestionQueue(numWorkers uint32, c *AviController, namespace string, msg string) {
+	// For Advancd L4 and service api , do not handle. services already been taken care
+	// in service handler
+	if lib.GetAdvancedL4() || lib.UseServicesAPI() {
```
sudswasavi, line 197: I think we should handle it for the ServicesAPI in the next PR.

internal/k8s/controller.go
```diff
@@ -190,6 +190,29 @@ func AddRoutesFromNSToIngestionQueue(numWorkers uint32, c *AviController, namesp
...
+		return
+	}
+	for _, svcObj := range svcObjs {
+		isSvcLb := isServiceLBType(svcObj)
+
+		if isSvcLb {
+			key := utils.L4LBService + "/" + utils.ObjKey(svcObj)
+			bkt := utils.Bkt(namespace, numWorkers)
+			c.workqueue[bkt].AddRateLimited(key)
+			utils.AviLog.Debugf("key: %s, msg: %s for namespace: %s", key, msg, namespace)
```
sudswasavi, line 212: here also - we might just handles "Services" in general as opposed to only service of type "Loadbalancer"

pkg/utils/utils.go
```diff
@@ -345,32 +345,32 @@ func GetNSFilter(obj *K8ValidNamespaces) (string, string) {
...
-	defer obj.validNSList.lock.Unlock()
-	delete(obj.validNSList.nsList, namespace)
+func DeleteNamespaceFromFilter(namespace string) {
+	globalNSFilterObj.validNSList.lock.Lock()
+	defer globalNSFilterObj.validNSList.lock.Unlock()
+	delete(globalNSFilterObj.validNSList.nsList, namespace)
 }
 
-func CheckIfNamespaceAccepted(namespace string, obj *K8ValidNamespaces, nsLabels map[string]string, nonNSK8ResFlag bool) bool {
+func CheckIfNamespaceAccepted(namespace string, nsLabels map[string]string, nonNSK8ResFlag bool) bool {
```
sudswasavi, line 360: It looks like you are mostly passing a `nil` for nsLabels in most of the calls to this method - in such a case should we move this as a optional parameter (third in this arg list). That way all your methods can just pass 2 args to this method except the ones that need to pass the nsLabels.

internal/k8s/controller.go
akshayhavile: Will Handle ServiceAPI in next PR

akshayhavile (commented):
As part of this checkin, Following things addressed:
1. Added namespace filter for clusterIP services.
2. Review Comments addressed.
3. Changed CheckIfNamespaceAccepted() to accept optional parameters.

akshayhavile:
@sudswasavi @monotosh-avi : Please review PR as merged conflicts with Pure L7 ingress changes and doing that touched some NPL annotation related code.

akshayhavile:
build ako

internal/nodes/dequeue_ingestion.go
```diff
@@ -194,6 +199,11 @@ func handlePod(key, namespace, podName string, fullsync bool) {
 			if ok {
 				utils.AviLog.Debugf("key: %s, msg: handling l4 Services %v", key, lbSvcs)
 				for _, lbSvc := range lbSvcs {
+					ns := strings.Split(lbSvc, "/")
+					//Do not append L4 service if namespace is invalid
+					if !utils.IsServiceNSValid(ns[0]) {
```
monotosh-avi, line 204: I think we can do this check only once, as all the services would be in the same NS

akshayhavile:
build ako

sudswasavi (approved):
LGTM.

monotosh-avi (approved):

PR #294, opened 2021-03-08
Allow AKO to function as a pure layer 7 ingress controller

sudswasavi:
This commit introduces a flag that would allow AKO to function as
a pure Layer 7 ingress controller. This flag should be used where
AKO isn't needed for layer 4 functionality. Note, this flag can be
only set during AKO bootup and cannot be changed while the AKO pod
is already in running state.

If the system had layer 4 LBs synced as virtual services, AKO would
delete those if the flag is set to `true` and AKO is rebooted.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -63,10 +63,26 @@ func GetNamePrefix() string {
 }
 
 var DisableSync bool
+var layer7Only bool
 
 func SetDisableSync(state bool) {
 	DisableSync = state
 	utils.AviLog.Infof("Setting Disable Sync to: %v", state)
+}
+
+func SetLayer7Only(val string) {
+	if boolVal, err := strconv.ParseBool(val); err == nil {
+		layer7Only = boolVal
+
+	} else {
+		// If the variable could not be converted, default to this being false
+		layer7Only = false
+	}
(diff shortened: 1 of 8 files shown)


parimanur (approved):

internal/lib/lib.go
```diff
@@ -63,12 +63,28 @@ func GetNamePrefix() string {
...
 	utils.AviLog.Infof("Setting Disable Sync to: %v", state)
 }
 
+func SetLayer7Only(val string) {
+	if boolVal, err := strconv.ParseBool(val); err == nil {
+		layer7Only = boolVal
+
+	} else {
+		// If the variable could not be converted, default to this being false
+		layer7Only = false
```
hemantavi, line 79: Nit: No need to set the false value here, as in case of error, `ParseBool` returns a false anyway.

hemantavi (commented):
LGTM, just one nit.

hemantavi (approved):

sudswasavi:
> @sudswasavi : I think we need to change this also : internal/nodes/dequeue_ingestion.go:198 Couldn't comment on line so pasting as comment.

Thanks @akshayhavile , fixed now. PTAL.

akshayhavile (approved):
LGTM

parimanur (approved):

PR #298, opened 2021-03-09
Introduce dedicated VS for ingress FQDNs

sudswasavi:
This commit introduces another ENUM value for the SHARD_SIZE called
DEDICATED. If this value is set, then AKO will create a dedicated
virtualservice per FQDN. If there are multiple paths configured
for the same FQDN across multiple ingress objects, then AKO would
create a single virtual service for the FQDN and patch all the paths
to the same VS.

### DIFF
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -47,7 +47,7 @@ L7Settings:
   defaultIngController: "true"
   l7ShardingScheme: "hostname"
   serviceType: ClusterIP #enum NodePort|ClusterIP
-  shardVSSize: "LARGE" # Use this to control the layer 7 VS numbers. This applies to both secure/insecure VSes but does not apply for passthrough. ENUMs: LARGE, MEDIUM, SMALL
+  shardVSSize: "LARGE" # Use this to control the layer 7 VS numbers. This applies to both secure/insecure VSes but does not apply for passthrough. ENUMs: LARGE, MEDIUM, SMALL, DEDICATED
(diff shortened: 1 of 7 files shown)


sudswasavi:
@monotosh-avi I have not taken care of passthrough virtualservice. Do you think we should take care of that as well?

monotosh-avi (approved):

chauhanshubham:
build ako

akshayhavile (commented):
@sudswasavi: I think we have to take care of evh model also for this. Also CRD translator uses ShardSize in certain way in addSeGroupLabel and checkRefOnController... Do you think we should take care of that also?

chauhanshubham (approved):

chauhanshubham:
> @sudswasavi: I think we have to take care of evh model also for this. Also CRD translator uses ShardSize in certain way in addSeGroupLabel and checkRefOnController... Do you think we should take care of that also?

In the crd translator, we use the GetShardsize func to ensure that we use the last REST goroutine to make the CRD related API calls to Avi (for validation). I think in this case, the functionality is preserved, so looks good to me.

sudswasavi:
> @sudswasavi: I think we have to take care of evh model also for this. Also CRD translator uses ShardSize in certain way in addSeGroupLabel and checkRefOnController... Do you think we should take care of that also?

The changes are not meant for EVH. EVH in first phase will not support dedicated VSes since we need the full openshift support for that, as this is a global flag setting. As far as the CRDs are concerned, we have not `yet` built the functionality for it - we want tot complete it as a global flag first and then move to CRDs.

akshayhavile (approved):
LGTM

PR #299, opened 2021-03-09
Update NPL annotation in Service based on Ingress Class

monotosh-avi:
Services whcih are ingress backends should only be annotated with NPL annotation,
only if the ingress is valid as per ingress class filtering through validateIngressForClass()

- Updating ingress-service map only if the ingress has valid ingress class. This is unlike the current
method, where we populate this map irrespective of Ingress Class.
- Upon Ingress Class update, repopulate this map and update corresponding service annotation.

### DIFF
--- a/internal/nodes/ingress_model_rel.go
+++ b/internal/nodes/ingress_model_rel.go
@@ -303,6 +303,7 @@ func IngressChanges(ingName string, namespace string, key string) ([]string, boo
 			objects.SharedSvcLister().IngressMappings(metav1.NamespaceAll).RemoveIngressClassMappings(namespace + "/" + ingName)
 		}
 	} else {
+
 		// simple validator check for duplicate hostpaths, logs Warning if duplicates found
 		success := validateSpecFromHostnameCache(key, ingObj.Namespace, ingObj.Name, ingObj.Spec)
 		if !success {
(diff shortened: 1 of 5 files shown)


monotosh-avi:
build ako

internal/nodes/ingress_model_rel.go
```diff
@@ -315,6 +316,18 @@ func IngressChanges(ingName string, namespace string, key string) ([]string, boo
 			objects.SharedSvcLister().IngressMappings(metav1.NamespaceAll).RemoveIngressClassMappings(namespace + "/" + ingName)
 		}
 
+		// If the Ingress Class is not found or is not valid, then return.
+		// When the correct Ingress Class is added, then the Ingress would be processed again.
+		if !validateIngressForClass(key, ingObj) {
```
chauhanshubham, line 321: This case probably gets covered as part of the L2 translator code.
https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/d1b6568c9a1caa4e5273ac5a8e5e2828463d7561/internal/nodes/avi_model_l7_translator.go#L54
and
https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/d1b6568c9a1caa4e5273ac5a8e5e2828463d7561/internal/nodes/avi_model_routeingr_hostname_shard.go#L140
where the relevant objects corresponding to an invalid ingress gets deleted (or are not created), if the ingressclass is found to be invalid.

chauhanshubham (approved):

sudswasavi (approved):

PR #302, opened 2021-03-15
cleanup test namespace in oshiftroutetests

monotosh-avi:
- If the test NS is not cleaned up properly then it can create problem in multi NS tests

### DIFF
--- a/tests/oshiftroutetests/oshift_l4_l7_nodeport_test.go
+++ b/tests/oshiftroutetests/oshift_l4_l7_nodeport_test.go
@@ -558,9 +558,9 @@ func TestSecureToInsecureRouteInNodePort(t *testing.T) {
 	TearDownTestForRouteInNodePort(t, defaultModelName)
 }
 
-// Removing this testcase for now, as we are missing events from non default Namespace with fakeclient in case of Routes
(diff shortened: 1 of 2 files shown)


akshayhavile:
@monotosh-avi: Changes looks good to me. I was thinking instead of creating labelled namespaces at each test, should we create all namespaces at TestMain() to avoid delay. As we have seen that for specific labelled namespace(in this case test), no ingestion queue event seen which caused failure.

monotosh-avi:
> @monotosh-avi: Changes looks good to me. I was thinking instead of creating labelled namespaces at each test, should we create all namespaces at TestMain() to avoid delay. As we have seen that for specific labelled namespace(in this case test), no ingestion queue event seen which caused failure.

We can do that too. But I think creating and deleting within a test case makes it self contained and possibly tests more conditions. For example, creating NS and then adding objects tests potential simultaneous event handlings.

chauhanshubham (approved):

sudswasavi:
cc @akshayhavile merging this - hope you are okay with the changes.

sudswasavi (approved):
Thanks @monotosh-avi  hopefully the pipelines would be back up now.

PR #304, opened 2021-03-16
AviInfraSetting integration with IngressClass and Routes

chauhanshubham:
This introduces AviInfraSetting integration with IngressClass via parametersRef, and with Routes using annotations.
The commit also changes the controller value in AKO GatewayClass to be `ako.vmware.com/avi-lb` instead of `lbapi.run.tanzu.vmware.com/avi-lb` which is meant for WCP specific GatewayClass usage.  
Updated `AviInfraSetting` to take multiple networkNames.


internal/nodes/avi_model_l7_translator.go
```diff
@@ -655,3 +662,55 @@ func RemoveRedirectHTTPPolicyInModel(vsNode *AviVsNode, hostname, key string) {
...
+
+	if !utils.GetIngressClassEnabled() {
+		return
+	} else if ingClassName == "" {
+		if ingClassName, found = isAviLBDefaultIngressClass(); !found {
+			return
+		}
+	}
+
+	if ingClassName != "" {
```
monotosh-avi, line 683: I think we can either skip this check or return if ingClassName == "", because we don't have to call buildWithInfraSetting in that case

internal/nodes/avi_model_routeingr_hostname_shard.go
```diff
@@ -529,3 +540,18 @@ func updateHostPathCacheV2(ns, ingress string, oldHostMap, newHostMap map[string
 		}
 	}
 }
+
+func buildL7RouteInfraSetting(key string, vs *AviVsNode, vsvip *AviVSVIPNode, routeIgrObj RouteIngressModel) {
```
monotosh-avi, line 544: Nit: I think this function can be in the file avi_model_l7_translator.go with buildL7IngressInfraSetting

monotosh-avi (approved):

parimanur (approved):

sudswasavi (approved):

PR #305, opened 2021-03-16
Handle error in rest layer to sync requests from multiple K8s objects

monotosh-avi:
Currently if we get an error in rest layer, then all other requests mapped to the
same shared VS are skipped. This leads to a scenario where problem in one ingress affects
other ingresses mapped to the same shared VS.

In this PR, not all requests are skipped in case of an error.
- If the failed request was anything other than POST call then we ignore the error and coninue.
- IF the failed request was POST then we attempt to remove its reference from other objects, then continue.

Also as part of this PR macro calls have been removed. In the current model it doesn't provide
any extra benifit as one macro call contains ony one request.

### DIFF
--- a/internal/rest/rest_operation.go
+++ b/internal/rest/rest_operation.go
@@ -1,0 +1,248 @@
+/*
+ * Copyright 2020-2021 VMware, Inc.
+ * All Rights Reserved.
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*   http://www.apache.org/licenses/LICENSE-2.0
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package rest
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
+
+	"github.com/avinetworks/sdk/go/clients"
+	avimodels "github.com/avinetworks/sdk/go/models"
+	"github.com/avinetworks/sdk/go/session"
+)
+
+// modelSchema defines an interface to handle rest operations for an object type.
+// For each object type (e.g. VirtualService), a new Model has to be implemented for this interface,
+// and the instance should be added in supportedModelTypes.
+// Usually the Model should have a refMap which has a mapping between reference object type and
+// a function which removes the reference. The function of the interface - RemoveRef should return
+// different functions based on the key of refMap.
+type modelSchema interface {
+	GetType() string
+	RemoveRef(reftype string) func(*utils.RestOp, string) bool
+}
+
+var (
+	virtualServiceModel = initVSModel()
+	poolGroupModel      = initPGModel()
+
+	// List of Models for which we would try to remove object refs in case of error
+	supportedModelTypes = map[string]modelSchema{
+		"VirtualService": virtualServiceModel,
+		"PoolGroup":      poolGroupModel,
+	}
+
+	// List of Objects for which we would handle error
+	supportedObjForError = map[string]struct{}{
+		"Pool":                 {},
+		"VsVip":                {},
+		"SSLKeyAndCertificate": {},
+	}
+)
+
+type poolGroupSchema struct {
+	Type   string
+	refMap map[string]func(*utils.RestOp, string) bool
+}
+
+func (v *poolGroupSchema) GetType() string {
+	return v.Type
+}
+func (v *poolGroupSchema) RemoveRef(refType string) func(*utils.RestOp, string) bool {
+	return v.refMap[refType]
+}
+
+func initPGModel() *poolGroupSchema {
+	pg := poolGroupSchema{}
+	pg.Type = "PoolGroup"
+	pg.refMap = map[string]func(*utils.RestOp, string) bool{
+		"Pool": pg.removePoolRef,
+	}
+	return &pg
+}
+
+func (v *poolGroupSchema) removePoolRef(op *utils.RestOp, objRef string) bool {
+	pg, ok := op.Obj.(*avimodels.PoolGroup)
+	if !ok {
+		utils.AviLog.Infof("Failed to remove Pool ref, object is not of type PoolGroup")
+		return false
+	}
+
+	for i := range pg.Members {
+		if strings.EqualFold(*pg.Members[i].PoolRef, objRef) {
+			pg.Members = append(pg.Members[:i], pg.Members[i+1:]...)
+			utils.AviLog.Infof("Successfully removed pool ref %s from PoolGroup: %s", objRef, *pg.Name)
+			break
+		}
+	}
+	op.Obj = pg
+	return true
+}
+
+type virtualserviceSchema struct {
+	Type   string
+	refMap map[string]func(*utils.RestOp, string) bool
+}
+
+func (v *virtualserviceSchema) GetType() string {
+	return v.Type
+}
+
+func initVSModel() *virtualserviceSchema {
+	vs := virtualserviceSchema{}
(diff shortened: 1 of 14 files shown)


monotosh-avi:
build ako

internal/rest/rest_operation.go
```diff
@@ -0,0 +1,248 @@
...
+func (v *virtualserviceSchema) removePoolRef(op *utils.RestOp, objRef string) bool {
+	vs, ok := op.Obj.(*avimodels.VirtualService)
+	if !ok {
+		utils.AviLog.Infof("Failed to remove Pool ref, object is not of type Virtualservice")
+		return false
+	}
+	if strings.EqualFold(*vs.PoolRef, objRef) {
+		vs.PoolRef = nil
+	}
+	for i := range vs.ServicePoolSelect {
```
sudswasavi, line 146: I don't think we should have ServicePoolSelect anymore. It should be l4policyset.

tests/integrationtest/l4_service_test.go
```diff
@@ -423,22 +423,30 @@ func TestCreateServiceLBWithFaultCacheSync(t *testing.T) {
...
 				w.WriteHeader(http.StatusBadRequest)
 				fmt.Fprintln(w, `{"error": "bad request"}`)
 			} else {
-				rName := rData["name"].(string)
+				if strings.Contains(url, "virtualservice") {
+					rModelName = "virtualservice"
+				} else if strings.Contains(url, "vsvip") {
+					rModelName = "vsvip"
+				} else if strings.Contains(url, "l4policy") {
+					rModelName = "l4policy"
```
sudswasavi, line 441: Can we call it the full name viz. l4policyset?

internal/rest/rest_operation.go
```diff
@@ -0,0 +1,248 @@
...
+			utils.AviLog.Errorf("Unknown RestOp %v", op.Method)
+			op.Err = fmt.Errorf("Unknown RestOp %v", op.Method)
+		}
+		if op.Err != nil {
+			utils.AviLog.Warnf(`RestOp method %v path %v tenant %v Obj %s returned err %s with response %s`,
+				op.Method, op.Path, op.Tenant, utils.Stringify(op.Obj), utils.Stringify(op.Err), utils.Stringify(op.Response))
+			// Wrap the error into a websync error.
+			err := &utils.WebSyncError{Err: op.Err, Operation: string(op.Method)}
+			aviErr := op.Err.(session.AviError)
+			if !isErrorRetryable(aviErr.HttpStatusCode) {
```
sudswasavi, line 229: So looking at this logic, it appears the following might happen:
- if you got a non-retryable error and the operation is not post then you will get rid of the ref from the object model for the first object.
- However, the method would return from Line 241 because of the continue - so if you effectively had more than 1 REST OP that required reference removal, it will not work?

internal/rest/rest_operation.go
```diff
@@ -0,0 +1,248 @@
...
+				if op.Method != utils.RestPost {
```
sudswasavi, line 230: Can you pls re-write the methods as:
if !isErrorRetryable(aviErr.HttpStatusCode) && op.Method == utils.RestPost {
      if removeObjRefFromRestOps(rest_ops, op.ObjName, op.Model) {
          return err // If you want to return right after 1 ref removal which appears to be happening here
    } 
}

internal/rest/rest_operation.go
sudswasavi: Ah looks like it's the exact opposite of what I described that should happen here? That is - you will remove all the references by use of that continue?

internal/rest/rest_operation.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/avinetworks/sdk/go/session"
+)
+
+// modelSchema defines an interface to handle rest operations for an object type.
+// For each object type (e.g. VirtualService), a new Model has to be implemented for this interface,
+// and the instance should be added in supportedModelTypes.
+// Usually the Model should have a refMap which has a mapping between reference object type and
+// a function which removes the reference. The function of the interface - RemoveRef should return
+// different functions based on the key of refMap.
+type modelSchema interface {
```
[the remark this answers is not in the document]
monotosh-avi: I thought that if this is not being used outside this file, better not to export it. If we need it later, we can export it.

internal/rest/rest_operation.go
monotosh-avi: Missed it, thanks. I have added one schema for L4PolicySet

internal/rest/rest_operation.go
monotosh-avi: What I wanted was .. in case of Put/Del if we get a non retryable error, we should continue with the next requests. Because Ref is not being changed, so there should be no impact on other requests, hence I added the continue statement. Without that we would follow the instructions after if block and return with error.

tests/integrationtest/l4_service_test.go
```diff
@@ -423,22 +423,30 @@ func TestCreateServiceLBWithFaultCacheSync(t *testing.T) {
...
 		url := r.URL.EscapedPath()
 
-		if strings.Contains(url, "macro") && r.Method == "POST" {
+		rModelName := ""
+		if r.Method == "POST" && !strings.Contains(url, "login") {
 			data, _ := ioutil.ReadAll(r.Body)
 			json.Unmarshal(data, &resp)
-			rData, rModelName := resp["data"].(map[string]interface{}), strings.ToLower(resp["model_name"].(string))
-			if rModelName == "virtualservice" && injectFault {
+			fmt.Printf("xxx: resp: %v", resp)
```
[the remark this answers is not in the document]
monotosh-avi: Removed

tests/integrationtest/l4_service_test.go
monotosh-avi: Done

sudswasavi:
LGTM

sudswasavi (approved):

parimanur (approved):

tests/integrationtest/lib.go
```diff
@@ -922,25 +922,30 @@ func NormalControllerServer(w http.ResponseWriter, r *http.Request, args ...stri
 	addrPrefix := "10.250.250"
 	object := strings.Split(strings.Trim(url, "/"), "/")
 
-	if strings.Contains(url, "macro") && r.Method == "POST" {
+	if r.Method == "POST" && !strings.Contains(url, "login") {
 		data, _ := ioutil.ReadAll(r.Body)
 		json.Unmarshal(data, &resp)
-		rData, rModelName := resp["data"].(map[string]interface{}), strings.ToLower(resp["model_name"].(string))
-		rName := rData["name"].(string)
+		_, rModelName := resp, ""
```
akshayhavile, line 928: Minor: we can simplify this line as rModelName := "" ... as resp is not considered

tests/integrationtest/lib.go
```diff
@@ -922,25 +922,30 @@ func NormalControllerServer(w http.ResponseWriter, r *http.Request, args ...stri
...
+		rName := resp["name"].(string)
 		objURL := fmt.Sprintf("https://localhost/api/%s/%s-%s-%s#%s", rModelName, rModelName, rName, RANDOMUUID, rName)
```
akshayhavile, line 930: With line no 928, rModelName will be empty so objURL will be "https://localhost/api//-rname-uuid#rname". will this be ok if not for virtual service or vsvip?

internal/rest/rest_operation.go
```diff
@@ -0,0 +1,288 @@
...
+			op.Err = c.AviSession.Patch(op.Path, op.Obj, op.PatchOp,
+				&op.Response)
+		case utils.RestDelete:
+			op.Err = c.AviSession.Delete(op.Path)
+		default:
+			utils.AviLog.Errorf("Unknown RestOp %v", op.Method)
+			op.Err = fmt.Errorf("Unknown RestOp %v", op.Method)
+		}
+		if op.Err != nil {
+			utils.AviLog.Warnf(`RestOp method %v path %v tenant %v Obj %s returned err %s with response %s`,
```
akshayhavile, line 264: Minor: op.path, op.tenant is string. May be we can change %v to %s.

akshayhavile (approved):
Few minor comments. LGTM.

PR #307, opened 2021-03-17
Fixing the ApplicationProfile and associating the redirect policy for EVH child VS.

parimanur:
- DEFAULT_L7__APP_PROFILE is used for secure and insecure ingresses for the EVH Vs.
- The redirect policy should be bound to the child EVH VS instead of the Parent.
- Removed ServiceMetadata information from pool as it was causing a race in status update for insecure ingress path update.
Tested the secure, insecure and the CRD usecase with this change.


parimanur:
build ako

parimanur:
build ako

monotosh-avi (approved):

sudswasavi (approved):
LGTM

PR #308, opened 2021-03-18
L4 Namespace Sync for GatewayAPI/ServiceAPI services

akshayhavile:
This PR has changes for:
1. L4 namespace sync for Service API
2. Few code corrections.

@sudswasavi @chauhanshubham : Recently, K8 Upstream changed project name from ServiceAPI to GatewayAPI.  I think we should also reflect out that change in our code base. (https://github.com/kubernetes-sigs/gateway-api).

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -194,7 +194,7 @@ func AddRoutesFromNSToIngestionQueue(numWorkers uint32, c *AviController, namesp
 func AddServicesFromNSToIngestionQueue(numWorkers uint32, c *AviController, namespace string, msg string) {
 	// For Advancd L4 and service api , do not handle. services already been taken care
 	// in service handler
-	if lib.GetAdvancedL4() || lib.UseServicesAPI() {
+	if lib.GetAdvancedL4() {
 		return
 	}
 	var key string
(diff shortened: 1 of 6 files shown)


akshayhavile:
build ako

internal/k8s/controller.go
```diff
@@ -557,12 +583,12 @@ func (c *AviController) SetupEventHandlers(k8sinfo K8sinformers) {
 			namespace, _, _ := cache.SplitMetaNamespaceKey(utils.ObjKey(svc))
 			if isSvcLb && !lib.GetLayer7Only() {
 				if !utils.IsServiceNSValid(namespace) {
-					utils.AviLog.Debugf("L4 Service delete event: Namespace: %s didn't qualify filter. Not deleting service.", namespace)
```
monotosh-avi, line 560: We usually don't ahem Info logs here, else there may be too many logs.

akshayhavile: @monotosh-avi :True. Changes has been made to make debugging easier when namespace feature enabled in customer env as well as in our integration pipeline. If you remember, 2 week back, there was issue in multinamespace oshift integration test cases where we were not able to determine whether namespace filtering has been applied to namespace or not. To take care such scenarios, changed this.

monotosh-avi: I think we would end up getting too many logs messages. May be we can run the UTs in Debug mode ?

monotosh-avi (approved):

internal/k8s/controller.go
```diff
@@ -217,6 +220,21 @@ func AddServicesFromNSToIngestionQueue(numWorkers uint32, c *AviController, name
 	}
 
 }
+func AddGatewaysFromNSToIngestionQueue(numWorkers uint32, c *AviController, namespace string, msg string) {
+	//TODO: Add code for gateway
```
sudswasavi, line 224: You have added the code for Gateway or are you planning to add some more code?

sudswasavi (approved):

PR #310, opened 2021-03-22
Temporary support for http caching via pools

sudswasavi:
Currently http caching is not supported on PoolGroups. This means
that if customers had a httppolicyset refer to a PG instead of a pool,
then this feature cannot work.

This commit addresses this concern with an interim fix that removes the
poolgroup references from the httppolicies and instead references the pools
directly. This flag is not recommeded for use unless http caching is
required because removing Poolgroups would prohibit us from using
canary flows. Hence these are the restrictions of the commit:

 - Only works for SNI VSes (no other thing, not insecure, nor EVH).
 - Will be deprecated once the Avi controller builds the http caching
functionality on poolgroups.
 - Flipping the flag to `true` in a running AKO setup will not take
effect unless AKO is rebooted. Some stale PGs could linger if they
were pre-created in SNIs. In which case, another reboot of AKO will
get rid of them as per the stale object deletion workflow that kicks
on bootup.

### DIFF
--- a/internal/nodes/avi_model_l7_hostname_shard.go
+++ b/internal/nodes/avi_model_l7_hostname_shard.go
@@ -217,6 +217,10 @@ func (o *AviObjectGraph) ManipulateSniNode(currentSniNodeName, ingName, namespac
 				} else {
 					sniPool = lib.GetSniPoolName(ingName, namespace, hostname, path, svc)
 				}
+				// Pls decprecate when PGs have http caching
+				if lib.GetNoPGForSNI() && isIngr {
+					sniPool = sniPool + "--" + lib.PoolNameSuffixForHttpPolToPool
+				}
 				o.RemovePoolNodeRefsFromSni(sniPool, modelSniNode)
 				o.RemovePoolRefsFromPG(sniPool, pgNode)
 			}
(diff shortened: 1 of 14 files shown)


sudswasavi:
build ako

internal/nodes/avi_model_l7_hostname_shard.go
```diff
@@ -217,6 +217,10 @@ func (o *AviObjectGraph) ManipulateSniNode(currentSniNodeName, ingName, namespac
 				} else {
 					sniPool = lib.GetSniPoolName(ingName, namespace, hostname, path, svc)
 				}
+				// Pls decprecate when PGs have http caching
+				if lib.GetNoPGForSNI() && isIngr {
+					sniPool = sniPool + "--" + "policy-to-pool"
```
monotosh-avi, line 222: Should we add this to a common function or add policy-to-pool to a const ? sniPool name is being generated two times.

monotosh-avi (commented):
Looks good to me. One minor comment.

sudswasavi:
I tried adding a suite which deletes the configmap and updates the flag to `true` but it had to be classified into a new suite with a single test in it to work. Reviewers, pls let me know if I should remove it or I can keep it.

internal/nodes/avi_model_l7_hostname_shard.go
sudswasavi: Added as a constant.

monotosh-avi:
> I tried adding a suite which deletes the configmap and updates the flag to `true` but it had to be classified into a new suite with a single test in it to work. Reviewers, pls let me know if I should remove it or I can keep it.

@sudswasavi I'm fine with another test suit. But can we be more specific instead of misc ? Else this may be used in future to add all kind of test cases.

monotosh-avi (approved):

sudswasavi:
build ako

chauhanshubham (approved):

PR #311, opened 2021-03-22
deprecate namespace shard scheme

chauhanshubham:


internal/k8s/ako_init.go
```diff
@@ -225,37 +225,21 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
...
-			// Setting the numWorkers to 1 as single node update in L2 affects multiple ingresses.
-			// Cannot have multiple workers working on ingress and node updates.
-			numWorkers = 1
-		} else {
-			numWorkers = utils.NumWorkersIngestion
-		}
-		ingestionQueueParams := utils.WorkerQueue{NumWorkers: numWorkers, WorkqueueName: utils.ObjectIngestionLayer}
-		graphQueueParams := utils.WorkerQueue{NumWorkers: utils.NumWorkersGraph, WorkqueueName: utils.GraphLayer}
-		graphQueue = utils.SharedWorkQueue(&ingestionQueueParams, &graphQueueParams, &slowRetryQParams, &fastRetryQParams).GetQueueByName(utils.GraphLayer)
+	numWorkers := uint32(1)
```
parimanur, line 234: do we want to limit this to 1?  only nodeport mode we had to scale down the workers.

chauhanshubham: I don't see any other possible values for this in case of hostname shard. hostname shard initializes this to 1 for all cases.

sudswasavi: I think hostname shard was using 1 thread for the graph layer earlier.

internal/nodes/avi_model_l7_hostname_shard.go
```diff
@@ -292,7 +268,7 @@ func sniNodeHostName(routeIgrObj RouteIngressModel, tlssetting TlsSettings, ingN
 		}
 		SharedHostNameLister().Save(sniHost, ingressHostMap)
 		sniHosts = append(sniHosts, sniHost)
-		shardVsName := DeriveHostNameShardVS(sniHost, key)
+		shardVsName := DeriveShardVS(sniHost, key)
```
sudswasavi, line 271: prefer the older name?

sudswasavi (approved):
what about the upgrade implications? I suppose none?

internal/nodes/avi_model_l7_hostname_shard.go
chauhanshubham: I thought of removing the explicit `HostName` usage since there's no other alternative.

monotosh-avi (approved):

PR #314, opened 2021-03-24
skip status deletion for Ingress if hostname is present in spec

monotosh-avi:
Without this fix we may delete an ingress status on path update, if
the pool for old path gets deleted in the end after addition of new pool.

### DIFF
--- a/internal/status/ing_status.go
+++ b/internal/status/ing_status.go
@@ -309,11 +309,19 @@ func deleteObject(svc_mdata_obj avicache.ServiceMetadataObj, key string, isVSDel
 	}
 
 	oldIngressStatus := mIngress.Status.LoadBalancer.DeepCopy()
+	var hostListIng []string
+	for _, rule := range mIngress.Spec.Rules {
+		hostListIng = append(hostListIng, rule.Host)
+	}
 
 	for i, status := range mIngress.Status.LoadBalancer.Ingress {
 		for _, host := range svc_mdata_obj.HostNames {
 			if status.Hostname == host {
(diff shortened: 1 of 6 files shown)


internal/status/ing_status.go
```diff
@@ -309,11 +309,19 @@ func deleteObject(svc_mdata_obj avicache.ServiceMetadataObj, key string, isVSDel
...
+	var hostListIng []string
+	for _, rule := range mIngress.Spec.Rules {
+		hostListIng = append(hostListIng, rule.Host)
+	}
 
 	for i, status := range mIngress.Status.LoadBalancer.Ingress {
 		for _, host := range svc_mdata_obj.HostNames {
 			if status.Hostname == host {
-				mIngress.Status.LoadBalancer.Ingress = append(mIngress.Status.LoadBalancer.Ingress[:i], mIngress.Status.LoadBalancer.Ingress[i+1:]...)
+				if !utils.HasElem(hostListIng, host) || isVSDelete {
```
akshayhavile, line 320: @monotosh-avi: We have to add namespace validation case here when namespace sync feature is enabled. Can you please check this deleted code and add namespace related same code? https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/290/files#diff-a4c1507719112824bd2d1e477c3835596c3ad4c884db9e52d1fe775f9bbc885dL321

monotosh-avi: Thanks Akshay, updated.

chauhanshubham (approved):

sudswasavi (approved):

sudswasavi:
build ako

akshayhavile (approved):

PR #316, opened 2021-03-24
Layer7 AviInfraSetting shardSize based modelName

chauhanshubham:
This commit introduces modelName generation based on AviInfraSetting CRD, and derives the shardVsName via the shardSize mentioned in the AviInfraSetting CR. To simplify transition cases between valid AviInfraSetting CRs, the Avi object names now include AviInfraSetting Name as well.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -43,7 +43,7 @@ var ShardSchemeMap = map[string]string{
 	"namespace": "namespace",
 }
 
-var shardSizeMap = map[string]uint32{
+var ShardSizeMap = map[string]uint32{
 	"LARGE":     8,
 	"MEDIUM":    4,
 	"SMALL":     1,
@@ -120,10 +120,10 @@ func GetEnableCtrl2014Features() bool {
 func GetshardSize() uint32 {
 	if GetAdvancedL4() {
 		// shard to 8 go routines in the REST layer
-		return shardSizeMap["LARGE"]
+		return ShardSizeMap["LARGE"]
 	}
 	shardVsSize := os.Getenv("SHARD_VS_SIZE")
-	shardSize, ok := shardSizeMap[shardVsSize]
+	shardSize, ok := ShardSizeMap[shardVsSize]
 	if ok {
 		return shardSize
 	} else {
@@ -176,9 +176,14 @@ func GetL7SharedPGName(vsName string) string {
 	return vsName
 }
 
-func GetL7PoolName(priorityLabel, namespace, ingName string, args ...string) string {
+func GetL7PoolName(priorityLabel, namespace, ingName, infrasetting string, args ...string) string {
 	priorityLabel = strings.ReplaceAll(priorityLabel, "/", "_")
-	poolName := NamePrefix + priorityLabel + "-" + namespace + "-" + ingName
+	var poolName string
+	if infrasetting != "" {
+		poolName = NamePrefix + infrasetting + "-" + priorityLabel + "-" + namespace + "-" + ingName
+	} else {
+		poolName = NamePrefix + priorityLabel + "-" + namespace + "-" + ingName
+	}
 	if len(args) > 0 {
 		svcName := args[0]
 		poolName = poolName + "-" + svcName
@@ -194,16 +199,25 @@ func GetHeaderRewritePolicy(vsName, localHost string) string {
 	return vsName + "--host-hdr-re-write" + "--" + localHost
 }
 
-func GetSniNodeName(ingName, namespace, secret string, sniHostName ...string) string {
-	if len(sniHostName) > 0 {
-		return NamePrefix + sniHostName[0]
+func GetSniNodeName(ingName, namespace, secret, settingName string, sniHostName ...string) string {
+	namePrefix := NamePrefix
+	if settingName != "" {
+		namePrefix = NamePrefix + "-" + settingName + "-"
 	}
-	return NamePrefix + ingName + "-" + namespace + "-" + secret
+	if len(sniHostName) > 0 {
+		return namePrefix + sniHostName[0]
+	}
+	return namePrefix + ingName + "-" + namespace + "-" + secret
 }
 
-func GetSniPoolName(ingName, namespace, host, path string, args ...string) string {
+func GetSniPoolName(ingName, namespace, host, path, infrasetting string, args ...string) string {
 	path = strings.ReplaceAll(path, "/", "_")
-	poolName := NamePrefix + namespace + "-" + host + path + "-" + ingName
+	var poolName string
+	if infrasetting != "" {
+		poolName = NamePrefix + infrasetting + "-" + namespace + "-" + host + path + "-" + ingName
+	} else {
+		poolName = NamePrefix + namespace + "-" + host + path + "-" + ingName
+	}
 	if len(args) > 0 {
 		svcName := args[0]
 		poolName = poolName + "-" + svcName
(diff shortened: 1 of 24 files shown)


chauhanshubham:
Given that this is a long PR, with some critical Layer 2 changes, I would prefer pushing this with the UTs, therefore marking this `in-progress`. Please proceed with the review though.

chauhanshubham:
build ako

chauhanshubham:
build ako

internal/lib/lib.go
```diff
@@ -176,9 +176,14 @@ func GetL7SharedPGName(vsName string) string {
 	return vsName
 }
 
-func GetL7PoolName(priorityLabel, namespace, ingName string, args ...string) string {
+func GetL7PoolName(priorityLabel, namespace, ingName, infrasetting string, args ...string) string {
 	priorityLabel = strings.ReplaceAll(priorityLabel, "/", "_")
-	poolName := NamePrefix + priorityLabel + "-" + namespace + "-" + ingName
+	var poolName string
+	if infrasetting != "" {
+		poolName = NamePrefix + infrasetting + "-" + priorityLabel + "-" + namespace + "-" + ingName
```
sudswasavi, line 183: K8s 1.19 onwards would we force people to use an Ingress class? If so - would they be forced to use an AviInfraSetting for the default ingress class or would it work by simply referring to the global settings (hence the else block) cc @monotosh-avi

internal/lib/lib.go
```diff
@@ -190,30 +195,46 @@ func GetL7HttpRedirPolicy(vsName string) string {
 	return vsName
 }
 
-func GetSniNodeName(ingName, namespace, secret string, sniHostName ...string) string {
+func GetSniNodeName(ingName, namespace, secret, settingName string, sniHostName ...string) string {
+	namePrefix := NamePrefix
```
sudswasavi, line 199: why simply initialise another variable here?

internal/nodes/avi_model_advl4_translator.go
```diff
@@ -202,7 +202,9 @@ func (o *AviObjectGraph) ConstructSvcApiL4VsNode(gatewayName, namespace, key str
 		}
 
 		// configures VS and VsVip nodes using infraSetting object (via CRD).
-		buildL4InfraSetting(key, avi_vs_meta, vsVipNode, nil, &gw.Spec.GatewayClassName)
+		if infraSetting := getL4InfraSetting(key, nil, &gw.Spec.GatewayClassName); infraSetting != nil {
```
sudswasavi, line 205: Previously there was no check for whether infraSetting is nil or not, now if it's nil - does it behave like older behaviour?

internal/lib/lib.go
```diff
@@ -190,30 +195,46 @@ func GetL7HttpRedirPolicy(vsName string) string {
...
 	if len(sniHostName) > 0 {
-		return NamePrefix + sniHostName[0]
+		return namePrefix + sniHostName[0]
 	}
-	return NamePrefix + ingName + "-" + namespace + "-" + secret
+	return namePrefix + ingName + "-" + namespace + "-" + secret
 }
 
-func GetSniPoolName(ingName, namespace, host, path string, args ...string) string {
+func GetSniPoolName(ingName, namespace, host, path, infrasetting string, args ...string) string {
```
sudswasavi, line 209: Can't the infraSetting be an optional parameter as well? Since the args ...string is a slice. This way you don't have to touch code where you don't want to specify the infraSetting param. You can assume that args[0] is infraSetting?

internal/nodes/dequeue_ingestion.go
```diff
@@ -470,26 +471,70 @@ func GetShardVSPrefix(key string) string {
...
+		}
+	}
+
+	aviInfraSetting := routeIgrObj.GetAviInfraSetting()
+	return oldAviInfraSetting, aviInfraSetting
+}
+
+// returns old and new models if changed, else just the current one.
+func DeriveShardVS(hostname string, key string, routeIgrObj RouteIngressModel) (string, string) {
+	// Read the value of the num_shards from the environment variable or the AviInfraSetting CRD.
```
sudswasavi, line 513: this comment says that we are reading the num_shards value, but I don't see it here...so a bit confused.

internal/nodes/avi_model_routeingr_hostname_shard.go
```diff
@@ -462,21 +290,55 @@ func DeleteStaleData(routeIgrObj RouteIngressModel, key string, modelList *[]str
...
+	}
+
+	utils.AviLog.Debugf("key: %s, msg: hosts to delete %s", key, utils.Stringify(hostMap))
+	for host, hostData := range hostMap {
+		oldShardVsName, newShardVsName := DeriveShardVS(host, key, routeIgrObj)
+		if oldShardVsName == newShardVsName {
+			continue
+		}
+
+		shardVsName := oldShardVsName
```
sudswasavi, line 322: Why do we have do this assignment? Can't we continue to use a single variable - in this case `oldShardVsName`?

internal/nodes/avi_model_routeingr_hostname_shard.go
```diff
@@ -227,15 +45,20 @@ func HostNameShardAndPublish(objType, objname, namespace, key string, fullsync b
 		utils.AviLog.Infof("key: %s, starting unsupported object type: %s", key, objType)
 		return
 	}
+
+	// delete old Models in case the modelNames changes because of shardSize updates via AviInfraSetting
+	if !lib.IsEvhEnabled() {
```
sudswasavi, line 50: why is this blocked for EVH? Can't EVH have a model switch as well?

sudswasavi: I think you are blocking this feature for EVH?

internal/nodes/avi_model_routeingr_hostname_shard.go
```diff
@@ -462,21 +290,55 @@ func DeleteStaleData(routeIgrObj RouteIngressModel, key string, modelList *[]str
...
+
+		modelName := lib.GetModelName(lib.GetTenant(), shardVsName)
+		found, aviModel := objects.SharedAviGraphLister().Get(modelName)
+		if !found || aviModel == nil {
+			utils.AviLog.Warnf("key: %s, msg: model not found during delete: %s", key, modelName)
+			continue
+		}
+
+		// Delete the pool corresponding to this host
+		aviModel.(*AviObjectGraph).DeletePoolForHostname(shardVsName, host, routeIgrObj, hostData.PathSvc, key, infraSettingName, true, true, false)
```
sudswasavi, line 333: is SNI deletion not required?

sudswasavi (commented):
@monotosh-avi / @parimanur can you folks review this PR carefully.

internal/lib/lib.go
chauhanshubham: AviInfraSetting is not forced in any way, so we'll fall back to defaults.

internal/nodes/avi_model_advl4_translator.go
chauhanshubham: yes, same behaviour.

internal/nodes/avi_model_routeingr_hostname_shard.go
chauhanshubham: I'll cover EVH in a subsequent PR.

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

monotosh-avi (approved):

sudswasavi (approved):

PR #320, opened 2021-03-26
Introduce GSLB host header re-write for insecure hosts

sudswasavi:
If a GSLB host header is specified in the HostRule CRD and the FQDN
is insecure then this commit introduces a way by which AKO programs
a URL re-write rule as a httppolicyset on the shared VS.

When the GSLB request is made to the global FQDN the request would
land on the VS which hosts the local FQDN. The HTTP re-write rule
would then modify the request header to the cluster local host header.

Following this, the header (which is the local host header) would be
matched with the local pools priority labels via the datascript.

This change requires a datascript change which is only available
with 20.1.4 and hence couple of things happen:

 - This feature only works with 20.1.4 and above controllers.
 - On switching the controller to 20.1.4 and starting an AKO during
upgrade would auto-matically upgrade the datascript.

### DIFF
--- a/internal/nodes/avi_model_l7_translator.go
+++ b/internal/nodes/avi_model_l7_translator.go
@@ -148,6 +148,9 @@ func (o *AviObjectGraph) ConstructShardVsPGNode(vsName string, key string, vsNod
 
 func (o *AviObjectGraph) ConstructHTTPDataScript(vsName string, key string, vsNode *AviVsNode) *AviHTTPDataScriptNode {
 	scriptStr := utils.HTTP_DS_SCRIPT
+	if lib.GetEnableCtrl2014Features() {
+		scriptStr = utils.HTTP_DS_SCRIPT_MODIFIED
+	}
 	evt := utils.VS_DATASCRIPT_EVT_HTTP_REQ
 	var poolGroupRefs []string
 	pgName := lib.GetL7SharedPGName(vsName)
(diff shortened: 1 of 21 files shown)


sudswasavi:
build ako

hemantavi:
build ako

sudswasavi:
build ako

monotosh-avi:
build ako

internal/nodes/avi_model_l7_translator.go
```diff
@@ -431,6 +434,25 @@ func (o *AviObjectGraph) BuildPolicyRedirectForVS(vsNode []*AviVsNode, hostname
...
+		SourceHost: gslbHost,
+		TargetHost: localHost,
+	}
+
+	rewritePolicy := &AviHttpPolicySetNode{
+		Tenant:        lib.GetTenant(),
+		Name:          policyname,
+		HeaderReWrite: &rewriteRule,
+	}
+	if policyFound := FindAndReplaceHeaderRewriteHTTPPolicyInModel(vsNode[0], rewritePolicy, gslbHost, key); !policyFound && gslbHost != "" {
```
monotosh-avi, line 449: should we check if gslbHost == "" and return before calling FindAndReplaceHeaderRewriteHTTPPolicyInModel ?

monotosh-avi (approved):

hemantavi (approved):

PR #332, opened 2021-04-01
AviInfraSetting shardSize configuration for EVH VSes

chauhanshubham:
AviInfraSetting provides a way to configure ShardSizes for Shared VSes created by AKO. This PR facilitates modifying shardSize via the AviInfraSetting CRD for VSes created in EVH mode. In case of a shardSize change, the objects map to a differently named graph model, and objects from the old model are removed in  layer 2.

This commit also removes a few naming schemes that were previously used in namespace shardScheme.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -199,15 +199,12 @@ func GetHeaderRewritePolicy(vsName, localHost string) string {
 	return vsName + "--host-hdr-re-write" + "--" + localHost
 }
 
-func GetSniNodeName(ingName, namespace, secret, infrasetting string, sniHostName ...string) string {
+func GetSniNodeName(ingName, infrasetting, sniHostName string) string {
 	namePrefix := NamePrefix
 	if infrasetting != "" {
 		namePrefix += infrasetting + "-"
 	}
-	if len(sniHostName) > 0 {
-		return namePrefix + sniHostName[0]
-	}
-	return namePrefix + ingName + "-" + namespace + "-" + secret
+	return namePrefix + sniHostName
 }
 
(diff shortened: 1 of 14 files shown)


monotosh-avi (approved):
@parimanur please take a look at this PR to make sure that EVH part is fine.

internal/lib/lib.go
```diff
@@ -199,15 +199,12 @@ func GetHeaderRewritePolicy(vsName, localHost string) string {
...
+func GetSniNodeName(ingName, infrasetting, sniHostName string) string {
 	namePrefix := NamePrefix
 	if infrasetting != "" {
 		namePrefix += infrasetting + "-"
 	}
-	if len(sniHostName) > 0 {
-		return namePrefix + sniHostName[0]
-	}
-	return namePrefix + ingName + "-" + namespace + "-" + secret
+	return namePrefix + sniHostName
```
sudswasavi, line 207: does this change the naming convention of existing objects since it may have had the `secret` variable in the end?

chauhanshubham: this is dead code for namespace sharding which is deprecated now.

monotosh-avi: Isn't this used for hostname shard too .. from GetSniNodeName() ?

chauhanshubham: yes, but the naming scheme where we use the `secret` was only used for namespace shard.

sudswasavi (approved):
@parimanur can you PTAL?

internal/lib/lib.go
parimanur: Yes, Secret is passed as "". The secret can be removed.

parimanur (approved):

Issue #313, opened 2021-03-22
AKO should support for selector-less load balancer type of service.


XudongLiuHarold:
AKO should add support for selector-less `LoadBalacaner` type of service for customizing endpoints.

AKO needs to use IPs and Ports directly from `Endpoints` object to create Pool Members for a `selector-less` LoadBalancer type of Service. 

For example, if I want to specify endpoints by my own like this:
```yaml
apiVersion: v1
kind: Service
metadata:
 name: HA-lb
spec:
  ports:
  - protocol: TCP
    port: 6443
    targetPort: 6443
  type: LoadBalancer
---
apiVersion: v1
kind: Endpoints
metadata:
  name: ha-lb-eps
  namespace: default
subsets:
- addresses:
  - ip: 100.96.0.2
    nodeName: controlplane-node-1
  - ip: 100.96.0.7
    nodeName: controlplane-node-2
  ports:
  - port: 6443
    protocol: TCP
```

AKO should directly add IPs and Ports from `Endpoints` `ha-lb-eps` to virtual service's pool members.

sudswasavi:
@XudongLiuHarold how about your create the ep with the same name as the `service`?
That should solve your use case I think.

XudongLiuHarold:
> @XudongLiuHarold how about your create the ep with the same name as the `service`?
> That should solve your use case I think.

Hi @sudswasavi , I updated the description, sorry for the misleading.

andrewsykim:
@XudongLiuHarold I think the issue description is still misleading. I believe selectorless Services are supported already, but I think you're asking for the serviceType (ClusterIP or NodePorT) to be configurable per Service LoadBalancer right?

sudswasavi:
This should be available in AKO 1.4.1

PR #333, opened 2021-04-05
Introduce skipnodeport annotation

sudswasavi:
This PR introduces a new annotation called skipnodeport. In scenarios
where the global `serviceType` is set to `NodePort` the NodeIPs are
processed as Servers inside the Avi pool.

But in case, the user wants to skip these nodeports for such Services
and instead rely on an Endpoint object for the servers, the services
could be annotated with:

skipnodeport.ako.vmware.com/enabled: "true"

If this is set on a service, then AKO will look for the actual endpoint
object to figure out the servers for the pools instead of deriving the
nodeport info.

It is the responsibility of the network admin to ensure that the endpoints
are routable from the Avi Service Engines since static routes will be
skipped if service is set to NodePort.

This annotation will only work for nodeport mode. If the serviceType is 
set to NodePortLocal - this annotation has no effect. 
This annotation will also not have a net effect if the serviceType is set to ClusterIP. 
For example, if someone set this annotation to true or false, the net effect in ClusterIP mode is the same.

Implements: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/issues/313

### DIFF
--- a/internal/nodes/avi_model_l4_translator.go
+++ b/internal/nodes/avi_model_l4_translator.go
@@ -151,6 +151,11 @@ func (o *AviObjectGraph) ConstructAviL4PolPoolNodes(svcObj *corev1.Service, vsNo
 			if servers := PopulateServersForNPL(poolNode, svcObj.ObjectMeta.Namespace, svcObj.ObjectMeta.Name, false, key); servers != nil {
 				poolNode.Servers = servers
 			}
+		} else if _, ok := svcObj.GetAnnotations()[lib.SkipNodePortAnnotation]; ok {
+			// This annotation's presence on the svc object means that the node ports should be skipped.
+			if servers := PopulateServers(poolNode, svcObj.ObjectMeta.Namespace, svcObj.ObjectMeta.Name, false, key); servers != nil {
+				poolNode.Servers = servers
+			}
 		} else if serviceType == lib.NodePort {
 			if servers := PopulateServersForNodePort(poolNode, svcObj.ObjectMeta.Namespace, svcObj.ObjectMeta.Name, false, key); servers != nil {
 				poolNode.Servers = servers
--- a/internal/lib/constants.go
+++ b/internal/lib/constants.go
@@ -124,6 +124,7 @@ const (
 	NPLPodAnnotation               = "nodeportlocal.antrea.io"
 	NPLSvcAnnotation               = "nodeportlocal.antrea.io/enabled"
 	InfraSettingNameAnnotation     = "aviinfrasetting.ako.vmware.com/name"
+	SkipNodePortAnnotation         = "skipnodeport.ako.vmware.com/enabled"
 
 	// Specifies command used in namespace event handler
 	NsFilterAdd                    = "ADD"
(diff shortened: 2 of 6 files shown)


internal/nodes/avi_model_l4_translator.go
```diff
@@ -149,6 +149,11 @@ func (o *AviObjectGraph) ConstructAviL4PolPoolNodes(svcObj *corev1.Service, vsNo
 			if servers := PopulateServersForNPL(poolNode, svcObj.ObjectMeta.Namespace, svcObj.ObjectMeta.Name, false, key); servers != nil {
 				poolNode.Servers = servers
 			}
+		} else if _, ok := svcObj.GetAnnotations()[lib.SkipNodePortAnnotation]; ok {
```
monotosh-avi, line 152: May be we can change this condition to 
ok || serviceType != lib.NodePort  
and remove the last else condition.

monotosh-avi, line 152: Can we change this condition to 
ok || serviceType == lib.NodePort and remove the last else block ?

But the current checks might be simpler .. so this is fine too.

sudswasavi:
build ako

monotosh-avi (approved):

internal/nodes/avi_model_l4_translator.go
maplain: Is it the intention to populate the servers from Endpoints as well when serviceType is ClusterIP, and this annotation is present?

maplain:
thx @sudswasavi for making the change!

> But in case, the user wants to skip these nodeports for such Services
and instead rely on an Endpoint object for the servers, the services
could be annotated with:
>
> skipnodeport.ako.vmware.com/enabled: "true"
>
> If this is set on a service, then AKO will look for the actual endpoint
object to figure out the servers for the pools instead of deriving the
nodeport info.

One minor comment: from the code it seems like the presence of this annotation is enough to trigger the behavioral change, could you please update the description to reflect that?

internal/nodes/avi_model_l4_translator.go
sudswasavi: If someone uses this annotation while the serviceType is clusterIP, the effect is the same as not having the annotation at all.

sudswasavi:
> thx @sudswasavi for making the change!
> 
> > But in case, the user wants to skip these nodeports for such Services
> > and instead rely on an Endpoint object for the servers, the services
> > could be annotated with:
> > skipnodeport.ako.vmware.com/enabled: "true"
> > If this is set on a service, then AKO will look for the actual endpoint
> > object to figure out the servers for the pools instead of deriving the
> > nodeport info.
> 
> One minor comment: from the code it seems like the presence of this annotation is enough to trigger the behavioral change, could you please update the description to reflect that?

This annotation will only work for nodeport mode. If the `serviceType` is set to NodePortLocal - this annotation has no effect. This annotation will also not have a net effect if the `serviceType` is set to ClusterIP. For example, if someone set this annotation to `true` or `false`, the net effect in ClusterIP mode is the same.

hemantavi (approved):

PR #336, opened 2021-04-05
update status of k8s objects through a queue

monotosh-avi:
Currently function for updating and deleting status are called directly from rest layer.
This means multiple threads of rest layer may try to update the same object simultaneously,
To avoid this race condition, introducing a queue where objects would be added for status update,
which would be dequeued and processed later.

Unlike other worker queues in AKO, for status udpate we can not use only the key, hence the complete
object is added in the queue.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -212,12 +212,14 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 	var graphwg *sync.WaitGroup
 	var fastretrywg *sync.WaitGroup
 	var slowretrywg *sync.WaitGroup
+	var statusWG *sync.WaitGroup
 	if len(waitGroupMap) > 0 {
 		// Fetch all the waitgroups
 		ingestionwg, _ = waitGroupMap[0]["ingestion"]
 		graphwg, _ = waitGroupMap[0]["graph"]
 		fastretrywg, _ = waitGroupMap[0]["fastretry"]
 		slowretrywg, _ = waitGroupMap[0]["slowretry"]
+		statusWG, _ = waitGroupMap[0]["status"]
 	}
 
 	/** Sequence:
@@ -233,6 +235,7 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 	slowRetryQParams := utils.WorkerQueue{NumWorkers: retryQueueWorkers, WorkqueueName: lib.SLOW_RETRY_LAYER, SlowSyncTime: lib.SLOW_SYNC_TIME}
 	fastRetryQParams := utils.WorkerQueue{NumWorkers: retryQueueWorkers, WorkqueueName: lib.FAST_RETRY_LAYER}
 
+	statusQueueParams := utils.WorkerQueue{NumWorkers: 8, WorkqueueName: utils.StatusQueue}
 	numWorkers := uint32(1)
 	ingestionQueueParams := utils.WorkerQueue{NumWorkers: numWorkers, WorkqueueName: utils.ObjectIngestionLayer}
 	numGraphWorkers := lib.GetshardSize()
@@ -241,7 +244,7 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 		numGraphWorkers = 8
 	}
 	graphQueueParams := utils.WorkerQueue{NumWorkers: numGraphWorkers, WorkqueueName: utils.GraphLayer}
-	graphQueue = utils.SharedWorkQueue(&ingestionQueueParams, &graphQueueParams, &slowRetryQParams, &fastRetryQParams).GetQueueByName(utils.GraphLayer)
+	graphQueue = utils.SharedWorkQueue(&ingestionQueueParams, &graphQueueParams, &slowRetryQParams, &fastRetryQParams, &statusQueueParams).GetQueueByName(utils.GraphLayer)
 
 	// Setup and start event handlers for objects.
 	c.SetupEventHandlers(informers)
@@ -289,6 +292,11 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 	slowRetryQueue := utils.SharedWorkQueue().GetQueueByName(lib.SLOW_RETRY_LAYER)
 	slowRetryQueue.SyncFunc = SyncFromSlowRetryLayer
 	slowRetryQueue.Run(stopCh, slowretrywg)
+
+	statusQueue := utils.SharedWorkQueue().GetQueueByName(utils.StatusQueue)
+	statusQueue.SyncFunc = SyncFromStatusQueue
+	statusQueue.Run(stopCh, statusWG)
+
 LABEL:
 	for {
 		select {
@@ -306,6 +314,7 @@ LABEL:
 	graphQueue.StopWorkers(stopCh)
 	fastRetryQueue.StopWorkers(stopCh)
 	slowRetryQueue.StopWorkers(stopCh)
+	statusQueue.StopWorkers(stopCh)
 }
 
 func (c *AviController) FullSync() {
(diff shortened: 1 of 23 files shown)


internal/k8s/ako_init.go
```diff
@@ -233,6 +235,7 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 	slowRetryQParams := utils.WorkerQueue{NumWorkers: retryQueueWorkers, WorkqueueName: lib.SLOW_RETRY_LAYER, SlowSyncTime: lib.SLOW_SYNC_TIME}
 	fastRetryQParams := utils.WorkerQueue{NumWorkers: retryQueueWorkers, WorkqueueName: lib.FAST_RETRY_LAYER}
 
+	statusQueueParams := utils.WorkerQueue{NumWorkers: 8, WorkqueueName: utils.StatusQueue}
```
sudswasavi, line 238: Instead of hard coding 8 - should be the same as numGraphWorkers?

internal/k8s/ako_init.go
```diff
@@ -642,31 +651,64 @@ func (c *AviController) DeleteModels() {
...
+func SyncFromIngestionLayer(key interface{}, wg *sync.WaitGroup) error {
 	// This method will do all necessary graph calculations on the Graph Layer
 	// Let's route the key to the graph layer.
 	// NOTE: There's no error propagation from the graph layer back to the workerqueue. We will evaluate
 	// This condition in the future and visit as needed. But right now, there's no necessity for it.
 	//sharedQueue := SharedWorkQueueWrappers().GetQueueByName(queue.GraphLayer)
-	nodes.DequeueIngestion(key, false)
+
+	keyStr, ok := key.(string)
+	if !ok {
```
sudswasavi, line 662: If the key isn't a string then what can it be?
If so - does it point to a problem in the previous layer?
Also - if it does occur, and we return nil in Line 663 - looks like we won't be able to debug as to why a certain worker couldn't process anything, so putting a log statement would aid in that.

internal/rest/avi_obj_vs.go
```diff
@@ -446,12 +465,24 @@ func (rest *RestOperations) AviVsCacheAdd(rest_op *utils.RestOp, key string) err
 							pool_cache_obj, found := pool_cache.(*avicache.AviPoolCache)
 							if found {
 								if pool_cache_obj.ServiceMetadataObj.Namespace != "" {
-									status.UpdateRouteIngressStatus([]status.UpdateOptions{{
+									utils.AviLog.Infof("xxx updating status2")
```
sudswasavi, line 468: pls remove this log

pkg/utils/types.go
```diff
@@ -37,6 +37,7 @@ const (
 	DeleteEv            EvType = "DELETE"
 	NumWorkersIngestion uint32 = 2
 	NumWorkersGraph     uint32 = 2
+	NumWorkersStatus    uint32 = 2
```
sudswasavi, line 40: where is this variable getting used?

internal/rest/avi_obj_vs.go
```diff
@@ -406,34 +406,53 @@ func (rest *RestOperations) AviVsCacheAdd(rest_op *utils.RestOp, key string) err
...
-							ServiceMetadata: svc_mdata_obj,
-							Key:             key,
-						}}, false)
-					} else {
-						status.UpdateGatewayStatusAddress([]status.UpdateOptions{{
-							Vip:             vs_cache_obj.Vip,
-							ServiceMetadata: svc_mdata_obj,
-							Key:             key,
-						}}, false)
+						statusOption.ObjType = lib.SERVICES_API
```
sudswasavi, line 420: @chauhanshubham presume there's no need for service of type LB checks here?

sudswasavi (commented):
Overall looks great. Some comments inline.

sudswasavi (commented):
@monotosh-avi the changes look good, however are you going to address the initial bootup sync status call in a separate PR?

monotosh-avi:
> @monotosh-avi the changes look good, however are you going to address the initial bootup sync status call in a separate PR?

@sudswasavi  I have addressed it in the new commit in the file  internal/rest/k8s_utils.go

internal/status/status_toiler.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func DequeueStatus(objIntf interface{}) error {
+	obj, ok := objIntf.(StatusOptions)
+	if !ok {
+		utils.AviLog.Warnf("key: %s, object is not of type StatusOptions, %T", obj.Options.Key, objIntf)
+		return nil
+	}
+	if obj.ObjType == utils.L4LBService {
+		if obj.Op == "update" {
```
sudswasavi, line 35: one minor nit - are these strings better being constants? "delete", "update"?

sudswasavi:
> > @monotosh-avi the changes look good, however are you going to address the initial bootup sync status call in a separate PR?
> 
> @sudswasavi I have addressed it in the new commit in the file internal/rest/k8s_utils.go

Ah sorry, I see it now, thx.

sudswasavi (approved):

internal/status/status_toiler.go
monotosh-avi: Yeah, that would be better, updated.

monotosh-avi:
build ako

sudswasavi (approved):

internal/lib/constants.go
```diff
@@ -105,6 +105,8 @@ const (
 	DefaultRouteCert                           = "router-certs-default"
 	autoAnnotateService                        = "AUTO_ANNOTATE_SERVICE"
 	SeGroupLabelKey                            = "clustername"
+	UpdateOperation                            = "update"
```
chauhanshubham, line 108: nit: could we use UpdateStatus/DeleteStatus instead?

internal/status/ing_status.go
```diff
@@ -266,7 +266,12 @@ func patchIngressAnnotations(ingObj *networkingv1beta1.Ingress, vsAnnotations ma
 	return nil
 }
 
-func DeleteIngressStatus(svc_mdata_obj avicache.ServiceMetadataObj, isVSDelete bool, key string) error {
+func DeleteIngressStatus(options []UpdateOptions, isVSDelete bool, key string) error {
+	//func DeleteIngressStatus(svc_mdata_obj avicache.ServiceMetadataObj, isVSDelete bool, key string) error {
```
chauhanshubham, line 270: stale comment. Also when would the options length be 0?

internal/rest/k8s_utils.go
```diff
@@ -110,10 +110,41 @@ func (rest *RestOperations) SyncObjectStatuses() {
...
-		status.UpdateGatewayStatusAddress(allGatewayUpdateOptions, true)
+		for i := range allGatewayUpdateOptions {
+			statusOption := status.StatusOptions{
+				ObjType: lib.Gateway,
+				Op:      lib.UpdateOperation,
+				Options: &allGatewayUpdateOptions[i],
+			}
+			statusQueue := utils.SharedWorkQueue().GetQueueByName(utils.StatusQueue)
+			bkt := utils.Bkt(allGatewayUpdateOptions[i].ServiceMetadata.Gateway, statusQueue.NumWorkers)
+			statusQueue.Workqueue[bkt].AddRateLimited(statusOption)
```
chauhanshubham, line 121: i guess these can be moved to a separate PublishToStatusQueue function?

internal/rest/avi_obj_vs.go
chauhanshubham: lgtm

internal/lib/constants.go
monotosh-avi: Done

internal/status/ing_status.go
monotosh-avi: Removed. The length should not be zero, though I think it would be always one now. I'll see if we can remove the slice in a PR later, don't want to reformat that now.

internal/rest/k8s_utils.go
monotosh-avi: Done

chauhanshubham (approved):

monotosh-avi:
build ako

sudswasavi (approved):

PR #339, opened 2021-04-06
Move all CRD validations to the ingestion layer

sudswasavi:
The graph layer is single threaded. All the validation for all CRDs
which can involve ref checks on the controller, was happening in
this layer. This could potentially block updates for other objects
from getting processed in parallel. Since graph layer is single
threaded, we should stay way from any I/O operations on it.

This PR, moves all the CRD validations to the ingestion layer
and handles them as a part of the event handler threads.

This should improve performance of AKO when the system is heavily
loaded with multiple CRDs.


monotosh-avi:
build ako

sudswasavi:
build ako

sudswasavi:
build ako

internal/k8s/crdcontroller.go
```diff
@@ -208,3 +219,214 @@ func (c *AviController) SetupAKOCRDEventHandlers(numWorkers uint32) {
 
 	return
 }
+
+// validateHostRuleObj would do validation checks
+// update internal CRD caches, and push relevant ingresses to ingestion
+func ValidateHostRuleObj(key string, hostrule *akov1alpha1.HostRule) error {
```
monotosh-avi, line 225: I don't see the return value getting used anymore. We can skip it.

monotosh-avi (dismissed):

internal/k8s/crdcontroller.go
```diff
@@ -208,3 +219,214 @@ func (c *AviController) SetupAKOCRDEventHandlers(numWorkers uint32) {
...
+	var err error
+	fqdn := hostrule.Spec.VirtualHost.Fqdn
+	foundHost, foundHR := objects.SharedCRDLister().GetFQDNToHostruleMapping(fqdn)
+	if foundHost && foundHR != hostrule.Namespace+"/"+hostrule.Name {
+		err = fmt.Errorf("duplicate fqdn %s found in %s", fqdn, foundHR)
+		status.UpdateHostRuleStatus(key, hostrule, status.UpdateCRDStatusOptions{
+			Status: lib.StatusRejected,
+			Error:  err.Error(),
+		})
+		utils.AviLog.Warnf("key: %s, msg: %v", key, err)
```
hemantavi, line 235: Is there a reason for this warning here?

internal/k8s/crdcontroller.go
hemantavi: Actually, it would probably be better to return the error from here and print it in the caller function IMHO. What do you think @sudswasavi @monotosh-avi

internal/k8s/crdcontroller.go
```diff
@@ -208,3 +219,214 @@ func (c *AviController) SetupAKOCRDEventHandlers(numWorkers uint32) {
...
+// checkRefOnController checks whether a provided ref on the controller
+func checkRefOnController(key, refKey, refValue string) error {
+	uri := fmt.Sprintf("/api/%s?name=%s&fields=name,type,labels,created_by", refModelMap[refKey], refValue)
+	clients := avicache.SharedAVIClients()
+
+	// assign the last avi client for ref checks
+	aviClientLen := lib.GetshardSize()
+	result, err := lib.AviGetCollectionRaw(clients.AviClient[aviClientLen], uri)
+	if err != nil {
+		utils.AviLog.Warnf("key: %s, msg: Get uri %v returned err %v", key, uri, err)
```
hemantavi, line 308: If we are returning an error below, there's no need to print the warning here, right?

internal/k8s/crdcontroller.go
sudswasavi: Removed.

hemantavi (approved):

monotosh-avi (approved):

PR #341, opened 2021-04-06
Introduce skipnodeport annotation

sudswasavi:
This PR introduces a new annotation called skipnodeport. In scenarios
where the global `serviceType` is set to `NodePort` the NodeIPs are
processed as Servers inside the Avi pool.

But in case, the user wants to skip these nodeports for such Services
and instead rely on an Endpoint object for the servers, the services
could be annotated with:

skipnodeport.ako.vmware.com/enabled: "true"

If this is set on a service, then AKO will look for the actual endpoint
object to figure out the servers for the pools instead of deriving the
nodeport info.

It is the responsibility of the network admin to ensure that the endpoints
are routable from the Avi Service Engines since static routes will be
skipped if service is set to NodePort.

Implements: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/issues/313

### DIFF
--- a/internal/lib/constants.go
+++ b/internal/lib/constants.go
@@ -124,6 +124,7 @@ const (
 	NPLPodAnnotation               = "nodeportlocal.antrea.io"
 	NPLSvcAnnotation               = "nodeportlocal.antrea.io/enabled"
 	InfraSettingNameAnnotation     = "aviinfrasetting.ako.vmware.com/name"
+	SkipNodePortAnnotation         = "skipnodeport.ako.vmware.com/enabled"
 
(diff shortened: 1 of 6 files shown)


sudswasavi:
build ako

sudswasavi:
build ako

anish-avi:
build ako

sudswasavi:
build ako

sudswasavi:
build ako

monotosh-avi (approved):

chauhanshubham (approved):

PR #343, opened 2021-04-06
Logging improvement when serviceType is NodePortLocal

sudswasavi:
AKO does not support services of type NodePort in NodePortLocal mode
This commit will print a log while populating the servers inside a
pool when an ingress points to a service of type NodePort


monotosh-avi (approved):

sudswasavi:
build ako

sudswasavi:
build ako

hemantavi (approved):

PR #351, opened 2021-04-06
adding missing fields to AKO config and ako-operator values.yaml

chauhanshubham:


sudswasavi (approved):

hemantavi (commented):
Looks good overall, however we need to add the missing fields in the samples file as well: ako-operator/config/samples/ako_v1alpha1_akoconfig.yaml. Please take care of that and the FTs in a subsequent PR.

hemantavi (approved):

PR #362, opened 2021-04-09
Changes in rest layer to use Markers instead of Labels.

parimanur:
Disable GRBAC if avi controller build is less than 20.1.5.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -462,8 +462,8 @@ func (c *AviObjCache) AviPopulateAllPkiPRofiles(client *clients.AviClient, pkiDa
 			continue
 		}
 		checksum := lib.SSLKeyCertChecksum(*pki.Name, string(*pki.CaCerts[0].Certificate), "")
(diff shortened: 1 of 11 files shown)


sudswasavi (approved):
Thanks for getting this done quickly Parikshit. Do we need any change in UTs? I realised while disabling this feature using the flag that none of the UTs probably cover this. @monotosh-avi we can merge this issue if you are fine and then work on a UT PR subsequently.

parimanur:
> Thanks for getting this done quickly Parikshit. Do we need any change in UTs? I realised while disabling this feature using the flag that none of the UTs probably cover this. @monotosh-avi we can merge this issue if you are fine and then work on a UT PR subsequently.

I will test the PR and update it here. @monotosh-avi @akshayhavile  Can you review and merge post that?

monotosh-avi (approved):

akshayhavile (approved):
@parimanur: Changes looks good.I think we should modify description in values.yaml, for AKOSetting:GRBAC variable, to include controller version from which GRBAC will be enabled.  Once testing is successful, we can merge PR. Thanks.

PR #365, opened 2021-04-09
skip other operations if sni child deletion fails

monotosh-avi:
This fixes a problem where objects belonging to the parent VS could not be deleted,
when deleteConfig flag was set to true.


sudswasavi:
build ako

PR #369, opened 2021-04-13
Fix issue related with ordering of httppolicysets of VS having same host with different Paths for Openshift route, K8 ingress

akshayhavile:
As part of this PR, httppolicysets of same VS are sorted in ascending order of path.
Changes has been done for SNI and EVH models.

### DIFF
--- a/internal/rest/avi_obj_vs.go
+++ b/internal/rest/avi_obj_vs.go
@@ -257,16 +257,38 @@ func (rest *RestOperations) AviVsSniBuild(vs_meta *nodes.AviVsNode, rest_method 
 		var httpPolicyCollection []*avimodels.HTTPPolicies
 		internalPolicyIndexBuffer := int32(11)
 
-		sort.Slice(vs_meta.HttpPolicyRefs, func(i, j int) bool {
-			return vs_meta.HttpPolicyRefs[i].HppMap[0].Path[0] < vs_meta.HttpPolicyRefs[j].HppMap[0].Path[0]
+		var httpsWithHppMap []*nodes.AviHttpPolicySetNode
+		var httpsNoHppMap []*nodes.AviHttpPolicySetNode
+		for _, http := range vs_meta.HttpPolicyRefs {
+			if http.HppMap != nil && http.HppMap[0].Path != nil {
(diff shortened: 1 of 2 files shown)


akshayhavile:
build ako

internal/rest/avi_obj_vs.go
```diff
@@ -255,6 +256,11 @@ func (rest *RestOperations) AviVsSniBuild(vs_meta *nodes.AviVsNode, rest_method
 
 		var httpPolicyCollection []*avimodels.HTTPPolicies
 		internalPolicyIndexBuffer := int32(11)
+
+		sort.Slice(vs_meta.HttpPolicyRefs, func(i, j int) bool {
```
sudswasavi, line 260: For this type of sorting, we aren't considering the noPath condition. but with EVH we are doing that. If you had a host without any path - will path[0] have a "/"?

akshayhavile: @sudswasavi : Thanks. Yes. I missed that. With latest changes fixed ordering of no path and with path httpps in VS

hemantavi (commented):
Both functions `AviVsChildEvhBuild` and `AviVsSniBuild` are using the same methodology to figure out the index and assign the policies. Can't we take the common stuff out into a function and use that instead? Because, if something had to be changed later, we would have to do at multiple places, instead of just one. @akshayhavile

sudswasavi (approved):

chauhanshubham (commented):
We would need to reverse sort during checksum calculation in layer 2 in order to force a PUT call while publishing in the REST layer
cc: @sudswasavi @monotosh-avi

sudswasavi:
Hi @chauhanshubham I spoke with @akshayhavile . Your concern is a valid one but we plan to address it in a subsequent PR. Currently we create a httppolicyset object per host/path combination which is sub-optimal. We would create a single httppolicyset object per host and then use a sorting function to sort the paths in descending order in layer 2 and publish the model to Layer 3. This will get addressed in the next PR. Hence can we pls merge this for now? cc @hemantavi

hemantavi (approved):

PR #371, opened 2021-04-14
sslkeycert, pkiprofile, l4policy, vsdatascript layer 3 checksum without GRBAC fixes

chauhanshubham:


internal/cache/controller_obj_cache.go
```diff
@@ -461,15 +461,12 @@ func (c *AviObjCache) AviPopulateAllPkiPRofiles(client *clients.AviClient, pkiDa
...
-		if lib.GetGRBACSupport() && pki.Markers != nil {
-			checksum += lib.ObjectLabelChecksum(pki.Markers)
-		}
+
 		pkiCacheObj := AviPkiProfileCache{
 			Name:             *pki.Name,
 			Uuid:             *pki.UUID,
 			Tenant:           lib.GetTenant(),
-			CloudConfigCksum: checksum,
+			CloudConfigCksum: lib.SSLKeyCertChecksum(*pki.Name, string(*pki.CaCerts[0].Certificate), "", pki.Markers),
```
akshayhavile, line 469: @chauhanshubham : Markers field introduced in 20.1.5. So for controller version 20.1.4 and less, I think this will cause exception.

sudswasavi: +1 to @akshayhavile . Unless it's nil because of us using the 20.1.5 SDK.

chauhanshubham: I believe it would cause an Exception if pki is nil, which won't be the case here. We are using the 20.1.5 sdk which would probably omit the `Markers` key if it's not fetched in case of a 20.1.4 controller (or less). 
We were doing a `pki.Markers != nil` check prior to this, which has been moved inside the checksum function. So in case of an incompatible controller, we would just be passing `nil` here.

chauhanshubham: verified with Avi v20.1.4

internal/lib/lib.go
```diff
@@ -670,9 +671,16 @@ func VrfChecksum(vrfName string, staticRoutes []*models.StaticRoute) uint32 {
...
 
-func DSChecksum(pgrefs []string) uint32 {
+func DSChecksum(pgrefs []string, markers []*models.RoleFilterMatchLabel) uint32 {
 	sort.Strings(pgrefs)
 	checksum := utils.Hash(utils.Stringify(pgrefs))
+	if GetGRBACSupport() {
+		if markers != nil {
+			checksum += ObjectLabelChecksum(markers)
+		} else {
+			checksum += GetClusterLabelChecksum()
```
akshayhavile, line 681: @chauhanshubham : This will require same logic as that of ObjectLabelChecksum(markers) function.  If we consider existing 20.1.4 deployments where existing DS, L4 etc. objects will not labeled. So here if we add label cksum by default to it while adding it to cache , we will not do put call for adding labels at layer2 as computed and cache cksum will be same and those objects will not be labelled at controller side.

internal/rest/dequeue_nodes.go
```diff
@@ -1260,6 +1260,7 @@ func (rest *RestOperations) DatascriptCU(ds_nodes []*nodes.AviHTTPDataScriptNode
 						dsCacheObj := ds_cache.(*avicache.AviDSCache)
 						if dsCacheObj.CloudConfigCksum != ds.GetCheckSum() {
 							utils.AviLog.Debugf("key: %s, msg: datascript checksum changed, updating - %s", key, ds.Name)
+							utils.AviLog.Infof("DS CHECKSUM CHECK: cache: %v\n model: %v", utils.Stringify(dsCacheObj), utils.Stringify(ds))
```
akshayhavile, line 1263: Do we want this as Info?

internal/rest/dequeue_nodes.go
```diff
@@ -1479,6 +1480,7 @@ func (rest *RestOperations) KeyCertCU(sslkey_nodes []*nodes.AviTLSKeyCertNode, c
 				if ok {
 					cache_ssl_nodes = Remove(cache_ssl_nodes, ssl_key)
 					ssl_cache_obj, _ := ssl_cache.(*avicache.AviSSLCache)
+					utils.AviLog.Infof("SSL CHECKSUM CHECK: cache: %v\n model: %v", utils.Stringify(ssl_cache_obj), utils.Stringify(ssl))
```
akshayhavile, line 1483: Do we want this msg as info?

sudswasavi (approved):

parimanur (approved):

internal/lib/lib.go
sudswasavi: Yeah I think I am not able to understand why we have to ever add the ClusterLabelChecksum? the GRBAC support is either with markers or it does not exist.

akshayhavile: I think.. he is trying to support label stuff also.. as it is there in 20.1.4.

chauhanshubham: This does not change the existing functionality, just moves the checksum calculation based on labels inside the helper function. This might have been incorrect before as well?
cc: @parimanur

internal/rest/dequeue_nodes.go
chauhanshubham: removed

internal/rest/dequeue_nodes.go
chauhanshubham: removed

sudswasavi (dismissed):

internal/lib/lib.go
```diff
@@ -711,17 +719,33 @@ func InformersToRegister(oclient *oshiftclient.Clientset, kclient *kubernetes.Cl
...
+func L4PolicyChecksum(ports []int64, protocol string, markers []*models.RoleFilterMatchLabel, populateCache bool) uint32 {
 	var portsInt []int
 	for _, port := range ports {
 		portsInt = append(portsInt, int(port))
 	}
 	sort.Ints(portsInt)
-	return utils.Hash(utils.Stringify(portsInt)) + utils.Hash(protocol)
+	checksum := utils.Hash(utils.Stringify(portsInt)) + utils.Hash(protocol)
+	if GetGRBACSupport() {
+		if markers != nil && populateCache {
```
akshayhavile, line 742: @chauhanshubham , @sudswasavi : I feel this logic still have an issue. Lets say customer with older avi controller (having DS, sslkey objects etc) migrates to 20.1.5. So when AKO fetches objects (L3 side) from AVI controller, for DS objects say, Markers field will not be there. So it will be marked nil. So we will go to Else part and add labels value to checksum. Now ate L2 side, while computing we will also use same cksum logic. So cache obj cksum and computed cksum will be same, so we will not do PUT call to add markers to these objects. So these object will remain without Markers(labels).

internal/lib/lib.go
```diff
@@ -96,10 +96,9 @@ func SetGRBACSupport(val string) {
 	controllerVersion := utils.CtrlVersion
 	if gRBAC && CheckControllerVersionCompatibility(controllerVersion, "<", ControllerVersion2015) {
 		// GRBAC is supported from 20.1.5 and above
-		utils.AviLog.Infof("Disabling GRBAC as controller version is less than %v", ControllerVersion2015)
+		utils.AviLog.Infof("Disabling GRBAC as current controller version %s is less than %s.", ControllerVersion2015)
```
akshayhavile, line 99: Nit: Missed "controllerversion" parameter while printing msg.

sudswasavi (approved):
Great review @akshayhavile

akshayhavile (approved):
@chauhanshubham Thanks for changes. One nit comment. LGTM .

internal/lib/lib.go
sudswasavi: Another good one - I am taking back the review till this is fixed.

chauhanshubham: ah thanks will do

akshayhavile (approved):

chauhanshubham:
build ako

anish-avi:
build ako

anish-avi:
Test new credentials

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

anish-avi:
build ako

anish-avi:
build ako

anish-avi:
build ako

sudswasavi:
build ako

sudswasavi (approved):

PR #385, opened 2021-04-21
Check AVI controller state during boot up

monotosh-avi:
Only start AKO if AVI controller state is active, else we may face DB sync issue after upgrade.

### DIFF
--- a/cmd/ako-main/main.go
+++ b/cmd/ako-main/main.go
@@ -25,6 +25,8 @@ import (
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
 	crd "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/client/v1alpha1/clientset/versioned"
 
+	avicache "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/cache"
+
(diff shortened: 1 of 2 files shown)


sudswasavi:
build ako

sudswasavi (approved):

cmd/ako-main/main.go
```diff
@@ -140,6 +142,12 @@ func InitializeAKC() {
 		}
 	}
 
+	aviObjCache := avicache.SharedAviObjCache()
+	aviRestClientPool := avicache.SharedAVIClients()
+	if !aviObjCache.IsAviClusterActive(aviRestClientPool.AviClient[0]) {
+		utils.AviLog.Fatalf("Avi Controller Cluster state is not Active, shutting down AKO")
```
chauhanshubham, line 148: Can we shutdown using the following  mechanism.
https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/99405920a9e3fb701a59206be5ee0f22758902ee/internal/k8s/ako_init.go#L116

monotosh-avi: Is there a benefit in doing that .. we are checking this during boot up, so I thought we can just exit

hemantavi (approved):

PR #387, opened 2021-04-21
Remove SE group labels on deleteConfig

sudswasavi:
This commit removes the SE group labels configured for the AKO cluster.
This operation is carried out if `deleteConfig` is set to `true`.

The operation is only done for the label added for the cluster.

In order for this change to take effect, the existing SEs have to
be rebooted or new SEs have to be created within the SE group.


chauhanshubham (approved):

chauhanshubham:
build ako

monotosh-avi (approved):

PR #391, opened 2021-04-22
skip other operations if sni child deletion fails

monotosh-avi:
This fixes a problem where some objects could not be deleted, when deleteConfig flag was set to true.
This happens when session gets timed out, retry would fix the errors.


sudswasavi (approved):

chauhanshubham (approved):

sudswasavi:
build ako

PR #395, opened 2021-04-22
remove networkName and use vipNetworkList for all clouds

monotosh-avi:
Currently there are two ways to specify vip network - vipNetworkList for AWS and
networkName for others. Removing networkName to use vipNetworkList for all cases.
But vipNetworkList can contain multiple entries only for AWS.

### DIFF
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -40,11 +40,10 @@ NetworkSettings:
   #       - 11.0.0.1/24
   subnetIP: "" # Subnet IP of the vip network
   subnetPrefix: "" # Subnet Prefix of the vip network
-  networkName: "" # Network Name of the vip network
   enableRHI: false # This is a cluster wide setting for BGP peering.
-  vipNetworkList: []
-  # vipNetworkList: []
-  #  - networkName: subnet-xxxxxxx
+  vipNetworkList: [] # Network Name of the VIP network, multiple networks allowed only for AWS Cloud.
+  # vipNetworkList:
+  #  - networkName: net1
 
 ### This section outlines all the knobs  used to control Layer 7 loadbalancing settings in AKO.
 L7Settings:
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -2700,15 +2700,12 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
 func checkPublicCloud(client *clients.AviClient) bool {
 	if lib.IsPublicCloud() {
 		// Handle all public cloud validations here
-		if lib.GetCloudType() != lib.CLOUD_GCP {
-			vipNetworkList, err := lib.GetVipNetworkList()
-			if err == nil && len(vipNetworkList) != 0 {
-				return true
-			}
+		vipNetworkList, err := lib.GetVipNetworkList()
+		if err != nil {
+			utils.AviLog.Errorf("Got error while fetching VIP network list: %s", err.Error())
+			return false
+		} else if len(vipNetworkList) == 0 {
 			utils.AviLog.Errorf("vipNetworkList not specified, syncing will be disabled.")
-			if err != nil {
-				utils.AviLog.Errorf("%s", err.Error())
-			}
 			return false
 		}
 	}
@@ -2785,6 +2782,12 @@ func checkAndSetVRFFromNetwork(client *clients.AviClient) bool {
 		utils.AviLog.Warnf("Network name net specified, skipping fetching of the VRF setting from network")
 		return true
 	}
+
+	if !validateNetworkNames(client, networkList) {
+		utils.AviLog.Warnf("Failed to validate Network Names specified in VIP Network List")
+		return false
+	}
+
 	networkName := networkList[0]
 
 	uri := "/api/network/?include_name&name=" + networkName + "&cloud_ref.name=" + utils.CloudName
@@ -2821,6 +2824,29 @@ func checkAndSetVRFFromNetwork(client *clients.AviClient) bool {
 	vrfName := strings.Split(vrfRef, "#")[1]
 	utils.AviLog.Infof("Setting VRF %s found from network %s", vrfName, networkName)
 	lib.SetVrf(vrfName)
+	return true
+}
+
+func validateNetworkNames(client *clients.AviClient, networkList []string) bool {
+	for _, networkName := range networkList {
+		uri := "/api/network/?include_name&name=" + networkName + "&cloud_ref.name=" + utils.CloudName
+		result, err := lib.AviGetCollectionRaw(client, uri)
+		if err != nil {
+			utils.AviLog.Warnf("Get uri %v returned err %v", uri, err)
+			return false
+		}
+		elems := make([]json.RawMessage, result.Count)
+		err = json.Unmarshal(result.Results, &elems)
+		if err != nil {
+			utils.AviLog.Warnf("Failed to unmarshal data, err: %v", err)
+			return false
+		}
+
+		if result.Count == 0 {
+			utils.AviLog.Warnf("No networks found for networkName: %s", networkName)
+			return false
+		}
+	}
 	return true
 }
 
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -2700,25 +2700,15 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
 func checkPublicCloud(client *clients.AviClient) bool {
 	if lib.IsPublicCloud() {
 		// Handle all public cloud validations here
-		networkName := lib.GetNetworkName()
-		if networkName == "" && lib.GetCloudType() != lib.CLOUD_GCP {
-			if lib.GetCloudType() == lib.CLOUD_AWS {
-				//check for vipNetworkList if networkName not present for AWS Clouds
-				utils.AviLog.Infof("networkName not specified, checking for vipNetworkList as cloud is AWS")
-				vipNetworkList, err := lib.GetVipNetworkList()
-				if err == nil && len(vipNetworkList) != 0 {
-					utils.AviLog.Infof("Proceeding as multi-vip enabled")
-					return true
-				} else {
-					utils.AviLog.Errorf("vipNetworkList not specified, syncing will be disabled.")
-					if err != nil {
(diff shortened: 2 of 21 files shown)


monotosh-avi:
build ako

internal/cache/controller_obj_cache.go
```diff
@@ -2700,25 +2700,15 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
-					return false
-				}
+		if lib.GetCloudType() != lib.CLOUD_GCP {
+			vipNetworkList, err := lib.GetVipNetworkList()
+			if err == nil && len(vipNetworkList) != 0 {
+				return true
+			}
+			utils.AviLog.Errorf("vipNetworkList not specified, syncing will be disabled.")
+			if err != nil {
+				utils.AviLog.Errorf("%s", err.Error())
```
sudswasavi, line 2710: Can we print a more meaningful message?

internal/cache/controller_obj_cache.go
```diff
@@ -2700,25 +2700,15 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
-					utils.AviLog.Infof("Proceeding as multi-vip enabled")
-					return true
-				} else {
-					utils.AviLog.Errorf("vipNetworkList not specified, syncing will be disabled.")
-					if err != nil {
-						utils.AviLog.Errorf("%s", err.Error())
-					}
-					return false
-				}
+		if lib.GetCloudType() != lib.CLOUD_GCP {
```
sudswasavi, line 2703: Do you know why we make an exception in case of GCP?

internal/cache/controller_obj_cache.go
```diff
@@ -2785,11 +2775,17 @@ func checkAndSetVRFFromNetwork(client *clients.AviClient) bool {
...
-	networkName := lib.GetNetworkName()
-	if networkName == "" {
-		utils.AviLog.Warnf("Param networkName not specified, skipping fetching of the VRF setting from network")
+	networkList, err := lib.GetVipNetworkList()
+	if err != nil {
+		utils.AviLog.Warnf("Error gettting Network name: %s, skipping fetching of the VRF setting from network", err.Error())
+		return true
+	}
+
+	if len(networkList) == 0 {
```
sudswasavi, line 2784: is it better to combine this with the if statement on line 2779?

internal/nodes/avi_model_advl4_translator.go
```diff
@@ -112,8 +112,8 @@ func (o *AviObjectGraph) ConstructAdvL4VsNode(gatewayName, namespace, key string
 		}
 
 		if networkNames, err := lib.GetNetworkNamesForVsVipNode(); err != nil {
-			utils.AviLog.Warnf("key: %s, msg: error when getting vipNetworkList: ", key, err)
-			return nil
+			utils.AviLog.Warnf("key: %s, msg: error when getting vipNetworkList: %s", key, err.Error())
+			//return nil
```
sudswasavi, line 116: can we remove the commented code?

sudswasavi: Is not returning intentional?

sudswasavi (commented):
looks good overall. @chauhanshubham shall we change the aviinfraCRD's field or keep it as is?

internal/cache/controller_obj_cache.go
```diff
@@ -2785,11 +2775,17 @@ func checkAndSetVRFFromNetwork(client *clients.AviClient) bool {
...
+		utils.AviLog.Warnf("Network name net specified, skipping fetching of the VRF setting from network")
 		return true
 	}
+	networkName := networkList[0]
```
chauhanshubham, line 2788: do we need to check for all networks specified in vipNetworkList. I guess the logic @aaha97 wrote for AviInfraSetting CRD might come in handy here?

internal/lib/lib.go
```diff
@@ -419,26 +419,17 @@ func GetSubnetPrefixInt() int32 {
 }
 
 func GetNetworkNamesForVsVipNode() ([]string, error) {
```
chauhanshubham, line 421: We can just use the GetVipNetworkList now. This function was intended to cover two different kinds of inputs.

chauhanshubham (commented):
`helm/bin/ako` binary is added to the commit that needs to be removed.

chauhanshubham:
> looks good overall. @chauhanshubham shall we change the aviinfraCRD's field or keep it as is?

Do you mean changing the field name? it is currently provided via `.spec.network.names`. Also, we have another setting that comes under the network settings as `.spec.network.enableRhi`. Do you propose changing it to something more explicit like `.spec.vipNetwork.names` | `.spec.network.vipNetworkNames`

internal/cache/controller_obj_cache.go
monotosh-avi: I think it is better to keep them separate. Error messages are different and it would be easier to debug.

internal/cache/controller_obj_cache.go
monotosh-avi: Good point. I had followed the existing logic of ignoring GCP cloud. 
@parimanur can you please review this PR  .. specifically here if we need the  check for GCP too ?

parimanur: @sudswasavi @monotosh-avi  In GCP we can accept ipamdnsproviderprofile when creating the VIP. We can change the condition for IsPublicCloud to specifically check for AWS and Azure.

monotosh-avi: @parimanur I saw that the function checkPublicCloud() is being called only from ValidateUserInput(), In that case are we fine if you remove the GCP cloud check ? 
I didn't understand the ipamdnsproviderprofile part ..

internal/lib/lib.go
```diff
@@ -461,6 +453,12 @@ func GetVipNetworkList() ([]string, error) {
 	for _, subnet := range vipNetworkListObj {
 		vipNetworkList = append(vipNetworkList, subnet.NetworkName)
 	}
+	if len(vipNetworkList) > 1 {
+		// Only AWS cloud supports multiple VIP networks
+		if !(IsPublicCloud() && GetCloudType() != CLOUD_AWS) {
```
parimanur, line 458: We can validate for only AWS

helm/ako/values.yaml
```diff
@@ -40,11 +40,10 @@ NetworkSettings:
   #       - 11.0.0.1/24
   subnetIP: "" # Subnet IP of the vip network
   subnetPrefix: "" # Subnet Prefix of the vip network
-  networkName: "" # Network Name of the vip network
   enableRHI: false # This is a cluster wide setting for BGP peering.
-  vipNetworkList: []
```
parimanur, line 45: Can you change the comment, multiple networks are only supported for AWS cloud?

internal/cache/controller_obj_cache.go
parimanur: @monotosh-avi Yes, you can delete the check. I was talking about in general where the behaviour of GCP vs other public clouds.

sudswasavi (dismissed):

internal/cache/controller_obj_cache.go
```diff
@@ -2700,27 +2700,15 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
-			utils.AviLog.Errorf("Required param networkName not specified, syncing will be disabled.")
-			return false
+		vipNetworkList, err := lib.GetVipNetworkList()
+		if err == nil && len(vipNetworkList) != 0 {
+			return true
 		}
+		utils.AviLog.Errorf("vipNetworkList not specified, syncing will be disabled.")
+		if err != nil {
+			utils.AviLog.Errorf("Got error while fetching VIP network list: %s", err.Error())
+		}
```
hemantavi, line 2710: The error message at 2707 seems misleading if there was an error in getting the vip network list. We will print both the error messages in that case. Can we check for the specific conditions for both the cases:
1. vip network list is not specified.
2. error is not nil while getting vip network list.

internal/lib/lib.go
```diff
@@ -418,29 +418,21 @@ func GetSubnetPrefixInt() int32 {
...
+			return nil, fmt.Errorf("More than one network specified in VIP Network List and Cloud type is not AWS")
+		}
 	}
-	return ""
-}
+	return vipNetworkList, nil
+}*/
```
hemantavi, line 433: Can you please remove the commented out code?

internal/lib/lib.go
```diff
@@ -418,29 +418,21 @@ func GetSubnetPrefixInt() int32 {
...
 
+//GetNetworkNamesForVsVipNode, GetVipNetworkList
```
hemantavi, line 435: We should remove this too.

helm/ako/values.yaml
```diff
@@ -40,11 +40,10 @@ NetworkSettings:
...
-  # vipNetworkList: []
-  #  - networkName: subnet-xxxxxxx
+  vipNetworkList: [] # Network Name of the VIP network, multiple networks allowed only for AWS Cloud.
```
hemantavi, line 44: @monotosh-avi can we add somewhere in the documentation that we will take the default VIP n/w (IPAM) if the `vipNetworkList` is empty?

internal/lib/lib.go
monotosh-avi: Thats what we are checking in the if condition. If the cloud type is AWS return error.

helm/ako/values.yaml
monotosh-avi: We only support multi Vip for AWS added that in comment  in line no 44. Do you mean we can do this for other cloud too ?

internal/cache/controller_obj_cache.go
monotosh-avi: Got it. Done

helm/ako/values.yaml
monotosh-avi: Sure. I'll add.

internal/lib/lib.go
monotosh-avi: Done

internal/lib/lib.go
monotosh-avi: Done

sudswasavi (approved):

hemantavi (approved):

PR #421, opened 2021-04-30
fix for checking cloud type in GetVipNetworkList

monotosh-avi:


internal/lib/lib.go
```diff
@@ -440,7 +440,7 @@ func GetVipNetworkList() ([]string, error) {
 	}
 	if len(vipNetworkList) > 1 {
 		// Only AWS cloud supports multiple VIP networks
-		if !(IsPublicCloud() && GetCloudType() != CLOUD_AWS) {
+		if !(IsPublicCloud() && GetCloudType() == CLOUD_AWS) {
```
sudswasavi, line 443: why are we checking for IsPublicCloud if AWS is guaranteed to be a public cloud?

sudswasavi (approved):

chauhanshubham (approved):

PR #423, opened 2021-05-03
Removing default pool group reference from EVH Child.

akshayhavile:
Due to default pg reference, evh child was responding to each and every request.
With this change, evh child will responds to only path for which it has corresponding pgs.


monotosh-avi (approved):

sudswasavi:
Ah - @akshayhavile thanks for the fix. This seems like a major bug in 1.4.1?

sudswasavi (approved):

PR #424, opened 2021-05-03
Process SwitchingAction in http policy rule only if it contains SwitchingAction

monotosh-avi:
While populating http policy in cache we check for the ref type in SwitchingAction.
But not all rules contain SwitchingAction, for which we get a panic because we try to process a nil value.


akshayhavile (approved):

monotosh-avi:
build ako

sudswasavi (approved):

PR #432, opened 2021-05-04
Introduce bgpPeerLabels to enable selection of BGP peers for vsvip advertisement

chauhanshubham:
This commit
- adds a new field in the config map (`bgpPeerLabels`), and AviInfraSetting CRD (`.Network.BGPPeerLabels`).
- Updates the AKOConfig in ako-operator accordingly, to accomodate `bgpPeerLabels`.
- blocks setting the bgp peer labels on VsVip if the corresponding VS is not a BGP VS (disabled RHI).
- blocks setting bgp peer labels in values.yaml/AviInfraSetting CR, if `enableRhi` is set to `false`.


sudswasavi:
build ako

sudswasavi (commented):
Can we add a check that if bgppeerlabels are specified but enableRHI is set to true, AKO will not bootup?

internal/cache/controller_obj_cache.go
```diff
@@ -2829,6 +2837,16 @@ func checkAndSetVRFFromNetwork(client *clients.AviClient) bool {
 	return true
 }
 
+func checkBGPParams() bool {
+	enableRhi := lib.GetEnableRHI()
+	bgpPeerLabels := lib.GetBgpPeerLabels()
+	if !enableRhi && len(*bgpPeerLabels) > 0 {
```
sudswasavi, line 2843: why are you checking the pointers length here?

internal/k8s/crdcontroller.go
```diff
@@ -460,6 +460,16 @@ func validateHTTPRuleObj(key string, httprule *akov1alpha1.HTTPRule) error {
 // validateAviInfraSetting would do validaion checks on the
 // ingested AviInfraSetting objects
 func validateAviInfraSetting(key string, infraSetting *akov1alpha1.AviInfraSetting) error {
+	if ((infraSetting.Spec.Network.EnableRhi != nil && !*infraSetting.Spec.Network.EnableRhi) || infraSetting.Spec.Network.EnableRhi == nil) &&
+		len(infraSetting.Spec.Network.BgpPeerLabels) > 0 {
```
sudswasavi, line 464: You seem to be checking Network.EnableRhi for nil value, can the Network.BgpPeerLabels also be nil? In such a case, will this length check fail?

internal/lib/lib.go
```diff
@@ -447,6 +447,16 @@ func GetVipNetworkList() ([]string, error) {
 	return vipNetworkList, nil
 }
 
+func GetBgpPeerLabels() *[]string {
```
sudswasavi, line 450: Can we rename this method to GetGlobalBgpPeerLabels()?

internal/nodes/avi_model_advl4_translator.go
```diff
@@ -100,17 +103,22 @@ func (o *AviObjectGraph) ConstructAdvL4VsNode(gatewayName, namespace, key string
...
 		vsVipNode := &AviVSVIPNode{
-			Name:       lib.GetL4VSVipName(gatewayName, namespace),
-			Tenant:     lib.GetTenant(),
-			EastWest:   false,
-			VrfContext: lib.GetVrf(),
+			Name:          lib.GetL4VSVipName(gatewayName, namespace),
+			Tenant:        lib.GetTenant(),
+			EastWest:      false,
+			VrfContext:    lib.GetVrf(),
+			BGPPeerLabels: nil,
```
sudswasavi, line 110: is this initialisation to nil required? Isn't it implicit?

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -1484,6 +1493,16 @@ func buildWithInfraSettingForEvh(key string, vs *AviEvhVsNode, vsvip *AviVSVIPNo
 			vs.EnableRhi = &enableRhi
 		}
 
+		if vs.EnableRhi != nil && *vs.EnableRhi {
+			if infraSetting.Spec.Network.BgpPeerLabels != nil {
+				vsvip.BGPPeerLabels = &infraSetting.Spec.Network.BgpPeerLabels
+			} else {
+				vsvip.BGPPeerLabels = lib.GetBgpPeerLabels()
```
sudswasavi, line 1500: I wonder if this is the right behavior. What if the user just wanted the RHI to be true but they didn't want to specify a BGP peer label in the infra setting because they don't want this feature for this IngressClass/AviInfraCRD. Shouldn't we just have enable RHI with no BGP peer labels in that case?

internal/nodes/avi_model_nodes.go
```diff
@@ -892,6 +892,7 @@ type AviVSVIPNode struct {
 	SubnetIP                string
 	SubnetPrefix            int32
 	NetworkNames            []string
+	BGPPeerLabels           *[]string
```
sudswasavi, line 895: Why is this a pointer to a slice and not just a slice of string? Are you planning to pass this variable around and manipulate them inside functions? Slices should otherwise always be pass by reference.

internal/rest/avi_obj_vsvip.go
```diff
@@ -95,8 +95,19 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, cache_
...
 		}
 
-		// Override the Vip in VsVip tto bring in updates, keeping everything else as is.
+		if !vsCache.EnableRhi && vsvip_meta.BGPPeerLabels != nil && len(*vsvip_meta.BGPPeerLabels) > 0 {
+			err = fmt.Errorf("To use selective vip advertisement, %s VS must advertise vips via BGP. Please recreate the VS.", vsCache.Name)
+			utils.AviLog.Error(err)
+			return nil, err
+		}
+
+		if vsvip_meta.BGPPeerLabels != nil {
```
sudswasavi, line 104: if enableRHI is `true` and then someone updates the BGPPeerLabels, this code assumes that the BGPPeerLabels can be updated on the vsvip using a PUT call.  I guess this is the only exception in vsvip PUT call with RHI enabled?

internal/rest/avi_obj_vsvip.go
```diff
@@ -95,8 +95,19 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, cache_
 			}
 		}
 
-		// Override the Vip in VsVip tto bring in updates, keeping everything else as is.
+		if !vsCache.EnableRhi && vsvip_meta.BGPPeerLabels != nil && len(*vsvip_meta.BGPPeerLabels) > 0 {
```
monotosh-avi, line 98: What happens in the reverse case .. does the user have to recreate the VS ?

tests/hostnameshardtests/ingressclass_hostname_test.go
```diff
@@ -890,3 +890,81 @@ func TestUpdateIngressClassWithoutInfraSetting(t *testing.T) {
 	TeardownIngressClass(t, ingClassName2)
 	VerifyVSNodeDeletion(g, modelName)
 }
+
+func TestBGPConfigurationWithInfraSetting(t *testing.T) {
```
monotosh-avi, line 894: We can add few more tests, like changing peer labels.

internal/rest/dequeue_nodes.go
```diff
@@ -1306,7 +1306,7 @@ func (rest *RestOperations) VSVipCU(vsvip_nodes []*nodes.AviVSVIPNode, vs_cache_
 							utils.AviLog.Debugf("key: %s, msg: the checksums are same for VSVIP %s, not doing anything", key, vsvip_cache_obj.Name)
 						} else {
 							// The checksums are different, so it should be a PUT call.
-							restOp, err := rest.AviVsVipBuild(vsvip, vsvip_cache_obj, key)
+							restOp, err := rest.AviVsVipBuild(vsvip, vs_cache_obj, vsvip_cache_obj, key)
```
monotosh-avi, line 1309: In case we do get an error, the vsvip operation would be skipped, but do we want to proceed with other operation ?... e.g. pool also has a rhi enabled flag.

internal/k8s/crdcontroller.go
chauhanshubham: EnableRhi is a pointer to bool which will be nil if not specified. BgpPeerLabels is of type string slice which will be an empty slice if not specified.

internal/nodes/avi_model_evh_nodes.go
chauhanshubham: Yes that's covered as part of the  conditional i believe. If enableRhi is set to true, then the bgp peer labels are set, from either the infra CRD, or the config map. if not set then the labels are not added/removed.

internal/rest/avi_obj_vsvip.go
chauhanshubham: yes, labels can be updated freely.

chauhanshubham:
build ako

monotosh-avi (dismissed):

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

tests/ingresstests/l7_evh_rest_test.go
```diff
@@ -31,15 +31,18 @@ import (
...
-	defer integrationtest.DisableEVH()
-
-	g := gomega.NewGomegaWithT(t)
-	modelName := "admin/cluster--Shared-L7-EVH-0"
-
 	SetupDomain()
 	SetUpTestForIngress(t, integrationtest.AllModels...)
+	modelName := "admin/cluster--Shared-L7-EVH-0"
 	integrationtest.PollForCompletion(t, modelName, 5)
+	// _, aviModel := objects.SharedAviGraphLister().Get(modelName)
```
monotosh-avi, line 38: Lets remove the commented out code

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

monotosh-avi (approved):

sudswasavi (approved):

PR #433, opened 2021-05-04
Use index while accessing network from networkList in rest layer for AWS multiVip

monotosh-avi:
using networkName instead of index results in same network multiple times because we are using a pointer.

### DIFF
--- a/internal/rest/avi_obj_vsvip.go
+++ b/internal/rest/avi_obj_vsvip.go
@@ -29,6 +29,7 @@ import (
 
 	avimodels "github.com/avinetworks/sdk/go/models"
 	"github.com/davecgh/go-spew/spew"
+	"google.golang.org/protobuf/proto"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


internal/rest/avi_obj_vsvip.go
```diff
@@ -504,13 +504,13 @@ func (rest *RestOperations) AviVsVipCacheDel(rest_op *utils.RestOp, vsKey avicac
 func networkNamesToVips(networkNames []string) []*avimodels.Vip {
 	var vipList []*avimodels.Vip
 	autoAllocate := true
-	for vipIDInt, networkName := range networkNames {
+	for vipIDInt := range networkNames {
```
parimanur, line 507: we can retain the range and use the util function `proto.String`

monotosh-avi: Done

chauhanshubham (approved):

parimanur (approved):

PR #436, opened 2021-05-07
Slow retry forever on error - No available free IPs

monotosh-avi:
If we get an error in rest layer when no free IP is avialable, then
retry forever until a free IP is avialable. Usually users can increase
avialable IP range and then after sometime the VsVip would get a free IP automatically.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ npltests:
 
 .PHONY: int_test
 int_test:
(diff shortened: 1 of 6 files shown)


monotosh-avi:
This can be merged to master after 1.4.2 branch is created.

sudswasavi (approved):
There isn't any potential risk of getting into 1.4.2, so let's merge it. @chauhanshubham can you PTAL.

hemantavi (approved):

PR #444, opened 2021-05-12
SAML auth for AKO using AVI authtoken

aaha97:


internal/cache/avi_ctrl_clients.go
```diff
@@ -34,8 +34,9 @@ func SharedAVIClients() *utils.AviRestClientPool {
 
 	ctrlUsername := os.Getenv("CTRL_USERNAME")
 	ctrlPassword := os.Getenv("CTRL_PASSWORD")
+	ctrlAuthtoken := os.Getenv("CTRL_AUTHTOKEN")
 	ctrlIpAddress := os.Getenv("CTRL_IPADDRESS")
-	if ctrlUsername == "" || ctrlPassword == "" || ctrlIpAddress == "" {
+	if ctrlUsername == "" || (ctrlPassword == "" && ctrlAuthtoken == "") || ctrlIpAddress == "" {
```
monotosh-avi, line 39: Do we need the username when auth token is used ?

internal/k8s/ako_init.go
```diff
@@ -265,6 +266,8 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 	graphQueue.Run(stopCh, graphwg)
 	fullSyncInterval := os.Getenv(utils.FULL_SYNC_INTERVAL)
 	interval, err := strconv.ParseInt(fullSyncInterval, 10, 64)
+
+	refreshAuthtokenInterval := 12 //hours
```
monotosh-avi, line 270: Lets add this to a const.

internal/k8s/ako_init.go
```diff
@@ -328,6 +338,69 @@ LABEL:
...
+		utils.AviLog.Debugf("new token: %+v", token)
+
+		//TODO:
+		// update secret
+		aviSecret, err := c.informers.KubeClientIntf.ClientSet.CoreV1().Secrets("avi-system").Get(context.TODO(), "avi-secret", metav1.GetOptions{})
+		if err != nil {
+			utils.AviLog.Warnf("failed to get secret, err: %+v", err)
+			return
+		}
+		utils.AviLog.Debugf("arif debug: old authtoken %+v", aviSecret.Data["authtoken"])
```
monotosh-avi, line 386: Change the debug logs

internal/k8s/ako_init.go
```diff
@@ -290,6 +294,12 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 		} else {
 			utils.AviLog.Warnf("Full sync interval set to 0, will not run full sync")
 		}
+		ctrlAuthtoken := os.Getenv("CTRL_AUTHTOKEN")
```
monotosh-avi, line 297: CTRL_AUTHTOKEN is being used from multiple functions, lets add this to const.

internal/k8s/ako_init.go
```diff
@@ -290,6 +294,12 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
...
+		if ctrlAuthtoken != "" {
+			worker = utils.NewFullSyncThread(time.Duration(refreshAuthtokenInterval) * time.Hour)
```
monotosh-avi, line 299: this would end up overwriting worker variable which would create problem. 
Lets evaluate if we need a new full sync worker and then use the functionalities, e.g. gracefully shutting down the threads.

monotosh-avi (commented):
Lets add the UTs if possible after adding the functionalities.

sudswasavi (commented):
what happens if AKO is booted up with a stale auth token?

hemantavi:
@aaha97 As discussed today, can we please re-structure the code in such a way that AMKO can also use it? Perhaps `pkg/utils` can be an appropriate place to put the refresh token logic.

aaha97:
build ako

internal/cache/avi_ctrl_clients.go
aaha97: authtoken is associated with the username, and used same as password

internal/lib/lib.go
```diff
@@ -971,3 +971,59 @@ func IsAviLBDefaultIngressClassWithClient(kc kubernetes.Interface) (string, bool
...
+		return ctrlUsername, ctrlPassword, "", nil
+	}
+	if aviSecret.Data["authtoken"] != nil {
+		ctrlAuthtoken := string(aviSecret.Data["authtoken"])
+		return ctrlUsername, "", ctrlAuthtoken, nil
+	}
+	return ctrlUsername, "", "", nil
+}
+
+func RefreshAuthtoken(kc kubernetes.Interface) {
```
aaha97, line 992: Read creds from avi-secret
create a new aviclient
get user token details
create a new token
write the new token to avi secret
delete the old auth token from avi

pkg/utils/utils.go
```diff
@@ -413,3 +413,25 @@ func GetAKONamespace() string {
 	}
 	return AKO_DEFAULT_NS
 }
+
+func GetTokenFromRestObj(robj interface{}, ctrlAuthtoken string) (string, bool) {
```
aaha97, line 417: takes response object from /api/user-token and return with matching user token id (for deletion) and boolean that determines if refresh needs to performed
in case matching token is not found, it returns a no token id and true

hemantavi (commented):
@aaha97 Can you please modify your variable naming schemes (like `authtoken` to `authToken`) according to golang conventions?

internal/lib/lib.go
```diff
@@ -971,3 +971,59 @@ func IsAviLBDefaultIngressClassWithClient(kc kubernetes.Interface) (string, bool
...
+	if err != nil {
+		utils.AviLog.Errorf("Filed to read from avi secret, err:", err)
+	}
+	aviClient := utils.NewAviRestClientWithToken(ctrlIpAddress, ctrlUsername, ctrlAuthtoken)
+	if aviClient != nil {
+		userTokensListResp, err := utils.AuthtokenGet(aviClient)
+		if err != nil {
+			return
+		}
+		if oldTokenID, ok := utils.GetTokenFromRestObj(userTokensListResp, ctrlAuthtoken); ok {
```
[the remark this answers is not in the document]
aaha97: the condition may also be false if the token is found but the is not old enough..
maybe we can have an info message stating skipping token refresh simply?

internal/lib/lib.go
```diff
@@ -971,3 +971,59 @@ func IsAviLBDefaultIngressClassWithClient(kc kubernetes.Interface) (string, bool
...
+			newTokenResp, err := utils.AuthtokenCreate(aviClient)
+			if err != nil {
+				return
+			}
+			token := fmt.Sprintf("%v", newTokenResp.(map[string]interface{})["token"])
+			utils.AviLog.Infof("new token: %+v", token)
```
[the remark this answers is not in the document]
aaha97: since the user already has access to that information, i thought it would be fine to reveal it in the logs

chauhanshubham:
hey could you please create a fork of the upstream repo, and create a PR from the fork. Let's not  create branches in upstream. We can't have unprotected branches on upstream.

PR #448, opened 2021-05-19
Support for Passthough hosts in Ingress

monotosh-avi:
Adding support for passthrough hosts similar to openshift Route in kubernetes Ingress.
This can be used if user wants to send secure traffic a host which has to be terminated in the backend,
and AVI woould act only as passthrough. Similar to openshift route, multiple passthrough hosts can be
handled by same AVI VS. Insecure -> secure redirect policy would be enabled for passthourgh hosts in
Ingresses.

To use this feature, an Ingress has to be annoated with passthrough.ako.vmware.com/enabled: "true".
Other annotations like ako.vmware.com/enable-tls would be ignored. Similarly, TLS configuration,
if specified would be ignored. Multiple ingresses with same hostname should not be created in this case.


sudswasavi (dismissed):

monotosh-avi:
build ako

sudswasavi:
build ako

tests/hostnameshardtests/l7_hostname_rest_test.go
```diff
@@ -996,6 +1005,6 @@ func TestBootupIngressStatusPersistence(t *testing.T) {
 			return ingress.Status.LoadBalancer.Ingress[0].IP
 		}
 		return ""
-	}, 20*time.Second).Should(gomega.Equal("10.250.250.10"))
+	}, 40*time.Second).Should(gomega.Equal("10.250.250.10"))
 	TearDownIngressForCacheSyncCheck(t, modelName)
-}
+}*/
```
sudswasavi, line 1010: is this a syntactical problem?

monotosh-avi: I don't think so, this is likely a test issue. I think we need to refresh the cache before refreshing status like we do during boot up, but that may not be straight forward. 

This test was added recently to catch a bug in Initial status refresh. But I think we need to analyse this issue more and then add a add a test which takes care of the problem. This is blocking all the PRs, so I want to proceed for now.

tests/hostnameshardtests/ingress_annotation_test.go
```diff
@@ -0,0 +1,421 @@
...
+	VerifyPassthroughIngressDeletion(t, g, DefaultPassthroughModel, 0, 0)
+
+	err = KubeClient.NetworkingV1beta1().Ingresses("default").Delete(context.TODO(), passthroughIngressName, metav1.DeleteOptions{})
+	if err != nil {
+		t.Fatalf("error in adding Ingress: %v", err)
+	}
+	TearDownTestForIngress(t, DefaultPassthroughModel)
+}
+
+func TestPassthroughIngressAddAnnotation(t *testing.T) {
```
akshayhavile, line 183: @monotosh-avi : Should we add a test case where two hosts maps to single shared VS to check multiple DSset creation and only single redirect httpps.? Does split traffic allowed in k8 implementation of passthrough as allowed in openshift?

sudswasavi (approved):

internal/nodes/validator.go
```diff
@@ -266,6 +279,12 @@ func (v *Validator) ParseHostPathForIngress(ns string, ingName string, ingSpec n
 		}
 	}
 
+	if passthroughEnabled {
+		ingressConfig.PassthroughCollection = passConfig
+		utils.AviLog.Infof("key: %s, msg: host path config from passthrough enabled ingress: %+v", key, utils.Stringify(ingressConfig))
```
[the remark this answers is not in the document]
chauhanshubham: i think we can probable leave logging the ingressconfig object here since there is a log that prints it at the end of the function.

chauhanshubham (commented):
minor comment around logging.

internal/nodes/validator.go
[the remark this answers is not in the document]
monotosh-avi: We are returning from the function here, so  added a similar log that we print at the end of the function.

tests/hostnameshardtests/ingress_annotation_test.go
monotosh-avi: Good idea Akshay. Added a test to check multiple ingress case. 
Unlike openshift routes, Ingresses don't have alternate backends, we are not adding that functionality here.

monotosh-avi:
build ako

chauhanshubham (approved):

monotosh-avi:
build ako

akshayhavile (approved):

PR #451, opened 2021-05-21
Fixing Continuous POST calls for sslkeyandcertificate in 409 error and DELETE calls for other AVI objects in 404 error

akshayhavile:
This PR fixes
1. Continuous  post for SSLkeyandcertificate even continuous 409 errors are there in retry layer for that sslkeyandcertificate 
2. Continuous delete call for AVI Objects even 404 error are there in retry layer.

This fixes S1 issue AKO: AKO continuously does POST for sslkeyandcertificate and fails to configure all VS.

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -724,7 +724,9 @@ func (rest *RestOperations) RefreshCacheForRetryLayer(parentVsKey string, aviObj
 				case avimodels.Pool:
 					poolObjName = *rest_op.Obj.(avimodels.Pool).Name
 				}
-				rest_op.ObjName = poolObjName
+				if poolObjName != "" {
+					rest_op.ObjName = poolObjName
+				}
 				rest.AviPoolCacheDel(rest_op, aviObjKey, key)
 			case "PoolGroup":
 				var pgObjName string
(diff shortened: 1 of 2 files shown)


akshayhavile:
build ako

internal/rest/dequeue_nodes.go
```diff
@@ -724,7 +724,9 @@ func (rest *RestOperations) RefreshCacheForRetryLayer(parentVsKey string, aviObj
 				case avimodels.Pool:
 					poolObjName = *rest_op.Obj.(avimodels.Pool).Name
 				}
-				rest_op.ObjName = poolObjName
+				if poolObjName != "" {
```
monotosh-avi, line 727: When would poolObjName be empty ?

akshayhavile: @monotosh-avi : When we do delete call , we do not fill up rest_op.obj object so rest_op.Obj is nil... in that case switch will not be executed. This will result in empty poolObjName. So when AKO will try to remove this entry from cache of Pool.. but as pool object name is empty, it will not be removed from cache. This will trigger loop of DELETE calls till retry count.  Same logic applicable for other objects. I have seen repeated DELETE call when i did out of band secure VS deletion and afterward ingress updation for path.

monotosh-avi: Good catch !

monotosh-avi (approved):

chauhanshubham (approved):

PR #453, opened 2021-05-27
EVH: Encoding AVI object names

akshayhavile:


internal/lib/lib.go
```diff
@@ -267,38 +279,52 @@ func GetEvhVsPoolNPgName(ingName, namespace, host, path, infrasetting string, ar
 		svcName := args[0]
 		poolName = poolName + "-" + svcName
 	}
+	return Encode(poolName)
+}
+func GetEvhPoolName(ingName, namespace, host, path, infrasetting, svcName string) string {
```
monotosh-avi, line 284: GetEvhPoolName and GetEvhVsPoolNPgName can potentially create a confusion. GetEvhVsPoolNPgName gives us the encoded name and GetEvhPoolName gives us the unencoded name. I think we came rename this function, e.g. GetEvhPoolNameUnencoded (or anything suitable) to indicate that this is not encoded pool name, else in future we may end up using this function for populating pools. 

Currently GetEvhPoolName and GetEvhVsPoolNPgName are independent, so one can make a logical change in GetEvhPoolName without changing GetEvhVsPoolNPgName and it would create problems. We can call GetEvhPoolName from GetEvhVsPoolNPgName and then apply encoding on the return value. 

Though not introduced by this PR, we can thin about breaking up GetEvhVsPoolNPgName - one for PG, one for Pool. I think initially same function was used for both but now we need sic name for pool, but it is not needed for PG, so we can use different functions.

akshayhavile:
build ako

akshayhavile:
build ako

internal/lib/constants.go
```diff
@@ -114,6 +114,7 @@ const (
 	NPLService                                 = "NPLService"
 	SyncStatusKey                              = "syncstatus"
 	NoFreeIPError                              = "No available free IPs"
+	AKO_ENCODED                                = "--ako-encoded"
```
sudswasavi, line 117: @monotosh-avi / @akshayhavile should be prefix this value instead of suffixing?
Also can we instead call it - "ako-object-id--"

akshayhavile: @sudswasavi : Encoded object name currently looks like <cluster-name>--encodedvalue--ako-encoded.  With Prefix "ako-object-id--", user might think that this is uuid field of object (which is not the case). Also I think that prefixing instead of suffixing will make it difficult to understand format with <cluster-name> coming between prefix and encoded value (ako--encoded--<cluster-name>--encodedvalue>)

monotosh-avi (approved):

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

monotosh-avi (dismissed):

internal/lib/constants.go
sudswasavi: @akshayhavile I meant the format to be like this"

cluster-name--ako-object-id--<SHA1> 

Does that make it better?

sudswasavi: If you want to simply get rid of `ako-object-id--" and say cluster-name--<SHA1> - that also is good enough in my opinion.

akshayhavile:
build ako

monotosh-avi (approved):

chauhanshubham (approved):

sudswasavi (approved):

monotosh-avi (approved):

PR #461, opened 2021-05-31
fix sync status bootup tests

chauhanshubham:


chauhanshubham:
build ako

sudswasavi:
@chauhanshubham can we fix this PR and then move towards merging?

chauhanshubham:
it might just be better to have these in the functional tests, will mirror these there.

PR #463, opened 2021-06-01
Do not perform bulk status update on resources of un-accepted namespaces when namespace sync feature is enabled.

akshayhavile:
This PR handles following objects:
1. Ingress
2. Routes
3. Gateway
4. LB Service
5. Advance LB service

For services and routes, we do not have gated condition other than namespace sync feature to use.
So if namespace sync feature is disabled, we will reset status of all services and routes, irrespective whether it is handled by AKO or not.
That is not case with Ingress, gateways, as it has specific class which is only handled by AKO.

### DIFF
--- a/internal/status/ing_status.go
+++ b/internal/status/ing_status.go
@@ -478,7 +478,6 @@ func getIngresses(ingressNSNames []string, bulk bool, retryNum ...int) map[strin
 		// to return all AKO ingestable Ingresses.
 		aviIngClasses := make(map[string]bool)
 		if utils.GetIngressClassEnabled() {
-			//It seems v1bet1/ingressclass deprecated in 1.19+. Should we use networking/v1 ingress class ?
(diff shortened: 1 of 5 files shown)


internal/status/ing_status.go
```diff
@@ -478,6 +478,7 @@ func getIngresses(ingressNSNames []string, bulk bool, retryNum ...int) map[strin
 		// to return all AKO ingestable Ingresses.
 		aviIngClasses := make(map[string]bool)
 		if utils.GetIngressClassEnabled() {
+			//It seems v1bet1/ingressclass deprecated in 1.19+. Should we use networking/v1 ingress class ?
```
chauhanshubham, line 481: hey can we remove the comment? updating to networkingv1 will be an epic and will require refactors.

monotosh-avi (approved):

akshayhavile:
build ako

chauhanshubham (approved):

akshayhavile:
build ako

PR #468, opened 2021-06-07
Support for dynamic namespaces in AKO

### LINKED ISSUE #449: AKO should support customize deployed namespace
Currently, AKO can be deployed only in a particular namespace like `avi-system`, `vmware-system`.

Ideally, AKO should support deployed in a namespace that the user can specify.

chauhanshubham:
AKO would identify it's namespace based on the statefulset
through which it runs.
Fixes #449


chauhanshubham:
build ako

monotosh-avi (approved):

sudswasavi (approved):

PR #470, opened 2021-06-09
Support NSX-T cloud in AKO

sudswasavi:
This PR adds supports for the NSX-T cloud in AKO.

The user needs to specify the backend network's T1Lr mapping as a
config option in the values.yaml on which the Service Engine would
create the backend network.

### DIFF
--- a/helm/ako/templates/configmap.yaml
+++ b/helm/ako/templates/configmap.yaml
@@ -25,6 +25,7 @@ data:
   subnetIP: {{ .Values.NetworkSettings.subnetIP | quote }}
   enableRHI: {{ .Values.NetworkSettings.enableRHI | quote }}
   subnetPrefix: {{ .Values.NetworkSettings.subnetPrefix | quote }}
+  nsxt1LR: {{ .Values.NetworkSettings.nsxt1LR | quote }}
   logLevel: {{ .Values.AKOSettings.logLevel | quote }}
   deleteConfig: {{ .Values.AKOSettings.deleteConfig | quote }}
   advancedL4: {{ .Values.L4Settings.advancedL4 | quote }}
--- a/helm/ako/templates/configmap.yaml
+++ b/helm/ako/templates/configmap.yaml
@@ -25,7 +25,7 @@ data:
   subnetIP: {{ .Values.NetworkSettings.subnetIP | quote }}
   enableRHI: {{ .Values.NetworkSettings.enableRHI | quote }}
   subnetPrefix: {{ .Values.NetworkSettings.subnetPrefix | quote }}
-  nsxt1LR: {{ .Values.NetworkSettings.nsxt1LR | quote }}
+  nsxtT1LR: {{ .Values.NetworkSettings.nsxtT1LR | quote }}
   logLevel: {{ .Values.AKOSettings.logLevel | quote }}
   deleteConfig: {{ .Values.AKOSettings.deleteConfig | quote }}
   advancedL4: {{ .Values.L4Settings.advancedL4 | quote }}
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -144,6 +144,13 @@ spec:
                 name: avi-k8s-config
                 key: syncNamespace
           {{ end }}
+           {{ if .Values.NetworkSettings.nsxt1LR }}
+          - name: NSXT_T1_LR
+            valueFrom:
+              configMapKeyRef:
+                name: avi-k8s-config
+                key: nsxt1LR
+          {{ end }}
           - name: NAMESPACE_SYNC_LABEL_KEY
             valueFrom:
               configMapKeyRef:
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -149,7 +149,7 @@ spec:
             valueFrom:
               configMapKeyRef:
                 name: avi-k8s-config
-                key: nsxt1LR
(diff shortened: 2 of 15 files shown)


sudswasavi:
build ako

monotosh-avi (dismissed):

helm/ako/templates/configmap.yaml
```diff
@@ -25,6 +25,7 @@ data:
   subnetIP: {{ .Values.NetworkSettings.subnetIP | quote }}
   enableRHI: {{ .Values.NetworkSettings.enableRHI | quote }}
   subnetPrefix: {{ .Values.NetworkSettings.subnetPrefix | quote }}
+  nsxt1LR: {{ .Values.NetworkSettings.nsxt1LR | quote }}
```
chauhanshubham, line 28: values.yaml defines it as `nsxT1LR` with an uppercase `T`. This won't match correctly. a we use it consistently as either `nsxt1LR` or `nsxT1LR`. 

Also should it be `nsxtT1LR`, one `t` for the nsx variant and the other for `Tier1` (like the env variable in statefulset to be NSXT_T1_LR).

internal/lib/lib.go
```diff
@@ -457,6 +457,15 @@ func GetGlobalBgpPeerLabels() []string {
 	return bgpPeerLabels
 }
 
+func GetT1LRPath() string {
+	t1LrPath := os.Getenv("NSXT_T1_LR")
+	if t1LrPath == "" {
+		utils.AviLog.Error("Required param nsxt1LR, missing from the configuration")
```
chauhanshubham, line 463: Do we need the error log here? This might lead to wrong error messages, as it suggests the field to be required?
For instance, in the translator and the rest layer code where we don't have the cloud type check, and we just call this function, with a non NSX-T cloud it would still print this error.

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -574,6 +578,9 @@ func (o *AviObjectGraph) ConstructAviL7SharedVsNodeForEvh(vsName, key string, ro
 		VrfContext: vrfcontext,
 	}
 
+	if lib.GetT1LRPath() != "" {
+		vsVipNode.T1Lr = lib.GetT1LRPath()
```
chauhanshubham, line 582: minor: this can go in with the above conditional itself.

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -545,9 +546,12 @@ func (o *AviObjectGraph) ConstructAviL7SharedVsNodeForEvh(vsName, key string, ro
...
 	avi_vs_meta.NetworkProfile = utils.DEFAULT_TCP_NW_PROFILE
 	avi_vs_meta.EVHParent = true
-
-	vrfcontext := lib.GetVrf()
-	avi_vs_meta.VrfContext = vrfcontext
+	if lib.GetT1LRPath() != "" {
+		vrfcontext = ""
+	} else {
+		vrfcontext = lib.GetVrf()
+		avi_vs_meta.VrfContext = vrfcontext
```
chauhanshubham, line 553: we seem to be changing the vrfcontext for the VS node based on the Tier1 LR, so do we need to have the vrfcontext accomodated in the VSNode's CalculateChecksum as well?

chauhanshubham (commented):
I think this would also require ako-operator additions? @hemantavi @sudswasavi

helm/ako/templates/configmap.yaml
sudswasavi: I think we can call it nsxtT1LR consistently. It's easier for people to type. Thanks for the catch, I will make the change.

internal/lib/lib.go
sudswasavi: Good catch. Let me remove the log from here.

internal/nodes/avi_model_evh_nodes.go
sudswasavi: I think if I move this, then I have to shuffle a bit more of the code. the vsVipNode is only declared after that previous conditional.

chauhanshubham: ah yes, i did not see the vs/vsvip difference, my bad.

internal/nodes/avi_model_evh_nodes.go
sudswasavi: I would think no. The reason is the following:
- the vrfcontext is always going to set to "" for the NSX-T cloud.
- the vrfcontext will behave the same way for all other cloud types.

If in the NSX-T cloud, the vrfcontext will never transition back to a non-empty value in layer 2.

helm/ako/templates/statefulset.yaml
```diff
@@ -143,6 +143,13 @@ spec:
               configMapKeyRef:
                 name: avi-k8s-config
                 key: syncNamespace
+          {{ end }}
+           {{ if .Values.NetworkSettings.nsxt1LR }}
+          - name: NSXT_T1_LR
+            valueFrom:
+              configMapKeyRef:
+                name: avi-k8s-config
+                key: nsxt1LR
```
chauhanshubham, line 152: nsxtT1LR

akshayhavile (commented):
We need to remove vrfcontext ref from EVH vs object also present in function: AviVsBuildForEvh() and AviVsChildEvhBuild

chauhanshubham (approved):

akshayhavile (approved):

sudswasavi:
build ako

PR #472, opened 2021-06-14
add child node checksum to vs checksum calculation

chauhanshubham:
add UT corresponding to it that creates separate routes (reproducible
in case of ingresses too), with same host but differrent paths, and
then deletes one of the routes forcing a SNI pool delete. AKO does not
let graph nodes proceed to rest layer, unless the graph checksum changes,
(see saveAviModel) so for any changes whatsoever, the graph checksum must
change. This would lead to irrelevant PUT calls on the parent VS.
Fixes AV-116457


akshayhavile (approved):
LGTM

monotosh-avi (approved):

PR #474, opened 2021-06-18
Retry for vsvip and vrf Get calls, replace with wrapper functions

chauhanshubham:
A few Get calls were not being retried in case of REST API failures,
that affected the sync of some Avi objects in case of timed out controller
session (which returns a HTTP 401 status). This commit replaces those Get
calls with a AviGetRaw wrapper function that retries the API call in case
of failures.
This commit also removes unused utils cache code.


monotosh-avi (approved):

sudswasavi:
build ako

pkg/utils/avi_obj_cache.go
```diff
@@ -66,600 +59,3 @@ func (o *CtrlPropCache) GetAllCtrlProp() map[string]string {
...
-	PgCache         *AviCache
-	HTTPCache       *AviCache
-	SSLKeyCache     *AviCache
-	PkiProfileCache *AviCache
-	CloudKeyCache   *AviCache
-	PoolCache       *AviCache
-	SvcToPoolCache  *AviMultiCache
-}
-
-func NewAviObjCache() *AviObjCache {
```
sudswasavi, line 81: where did these functions move to? Are they not required any more?

sudswasavi (approved):

PR #477, opened 2021-06-22
sync CRD files from ako helm directory to ako-operator

chauhanshubham:
This also adds a pre-build sync-crd-files target in the Makefile
that should run before the build target and sync the CRD files to
the ako-operator helm directory.

Note: The expectation is that whenever someone changes the crd schema in AKO, 
the `build`/`build-local` targets should be executed which would automatically sync the 
changed CRD schemas to ako-operator. The source of truth would be the schemas 
defined in AKO and not AKO-operator, hence the new `pre-build` target is for AKO only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,14 +47,21 @@ endif
 .PHONY: all
 all: build docker
 
+.PHONY: sync-crd-files
+sync-crd-files:
+		cp ./helm/ako/crds/* ./ako-operator/helm/ako-operator/crds/
+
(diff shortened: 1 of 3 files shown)


chauhanshubham:
build ako

sudswasavi (approved):

sudswasavi:
Very nice work! Thanks @chauhanshubham

hemantavi (approved):

PR #482, opened 2021-06-29
Remove cache entry only when rest method is DELETE

akshayhavile:
This PR works on:
1. Fixing : Delete cache entry only when Rest Method is DELETE. In case of non-retryable exception, AKO delete cache entry even if rest method is POST /PUT. This PR fixes that. 
2. Few log printing related issues.

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -631,7 +631,7 @@ func (rest *RestOperations) PopulateOneCache(rest_op *utils.RestOp, aviObjKey av
 			rest.AviVsVipCacheAdd(rest_op, aviObjKey, key)
 		}
 
-	} else {
+	} else if rest_op.Err == nil && rest_op.Method == utils.RestDelete {
 		utils.AviLog.Infof("key: %s, msg: deleting %s cache", key, rest_op.Model)
 		if rest_op.Model == "PKIprofile" {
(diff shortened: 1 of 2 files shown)


akshayhavile:
build ako

internal/rest/dequeue_nodes.go
```diff
@@ -631,7 +631,7 @@ func (rest *RestOperations) PopulateOneCache(rest_op *utils.RestOp, aviObjKey av
 			rest.AviVsVipCacheAdd(rest_op, aviObjKey, key)
 		}
 
-	} else {
+	} else if rest_op.Method == utils.RestDelete {
```
monotosh-avi, line 634: I think we can add a error check here too 
else if (rest_op.Err == nil) && (rest_op.Method == utils.RestDelete)

akshayhavile: Addressed. Thanks.

monotosh-avi (dismissed):

akshayhavile:
build ako

akshayhavile:
build ako

monotosh-avi:
I have tried solving some UT issues in this PR. https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/484
Lets rebase when that gets merged.

monotosh-avi (approved):

sudswasavi (approved):

PR #487, opened 2021-07-02
EVH: Fix route transition issue from secure to insecure and vice versa

akshayhavile:
This PR fixes issue related with
1. Secure to insecure route transition.
2. Insecure to secure route transition,
3. Passthrough route deletion

### DIFF
--- a/internal/nodes/avi_model_routeingr_hostname_shard.go
+++ b/internal/nodes/avi_model_routeingr_hostname_shard.go
@@ -93,13 +93,13 @@ func HostNameShardAndPublish(objType, objname, namespace, key string, fullsync b
 	hostsMap := make(map[string]*objects.RouteIngrhost)
 
 	if lib.IsEvhEnabled() {
+		// delete stale data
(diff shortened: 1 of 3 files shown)


akshayhavile:
build ako

internal/nodes/avi_model_routeingr_hostname_shard.go
```diff
@@ -93,13 +93,13 @@ func HostNameShardAndPublish(objType, objname, namespace, key string, fullsync b
 	hostsMap := make(map[string]*objects.RouteIngrhost)
 
 	if lib.IsEvhEnabled() {
+		// delete stale data
+		DeleteStaleDataForEvh(routeIgrObj, key, &modelList, Storedhosts, hostsMap)
```
monotosh-avi, line 97: Calling DeleteStaleDataForEvh before processing would mean deletion of all objects from model, every time we get an update and then building the model again. 
What was the the issue if DeleteStaleDataForEvh is called after generating the models ?

akshayhavile:
All issue are being handled in different PR. so closing.

PR #488, opened 2021-07-03
Scale Test Changes for Jenkins integration

RiyaGaniga-avi:
This PR has changes for Jenkins integration. 

- Network Name type changes.
- Reboot of remote VMs (Kubernetes node and Controller) is now through golang's native method (golang.org/x/crypto/ssh) instead of sshpass.
- DNS A Record verification method has been modified (as the api response of dns-vs doesn't include DNS A Records anymore).
- Command line arguments are read using flag instead of indexes.


tests/scaletest/lib/utils.go
```diff
@@ -214,42 +214,49 @@ func FetchPools(t *testing.T, AviClient *clients.AviClient) []models.Pool {
...
-}
-
-func FetchDNSARecordsFQDN(t *testing.T, dnsVsUuid string, AviClient *clients.AviClient) []string {
-	ipamDNSRecords := FetchDnsVsFqdn(t, dnsVsUuid, AviClient)
-	var FQDNList []string
-	for _, ipamRecord := range ipamDNSRecords {
-		if *ipamRecord.Type == "DNS_RECORD_A" {
-			for j := 0; j < len(ipamRecord.Fqdn); j++ {
-				FQDNList = append(FQDNList, ipamRecord.Fqdn[j])
+	for result.Next != "" {
```
monotosh-avi, line 240: I think it might be better to have a recursive function, e.g. https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/internal/cache/controller_obj_cache.go#L330. 
The for loop won't have to be written twice  - _for _, elem := range elems {_

tests/scaletest/scale_test.go
```diff
@@ -189,18 +183,47 @@ func SetupForTesting(t *testing.T) {
...
 }
 
-/* Used for Controller and Node reboot */
-func Reboot(t *testing.T, wg *sync.WaitGroup, nodeType string, controllerIP string, username string, password string) {
-	t.Logf("Rebooting %s ... ", nodeType)
-	loginID := username + "@" + controllerIP
-	cmd := exec.Command("sshpass", "-p", password, "ssh", "-t", loginID, " `echo ", password, " |  sudo -S shutdown --reboot 0 && exit `")
-	_, err := cmd.Output()
+func RemoteReboot(user string, addr string, password string, cmd string) (string, error) {
+	_, _ = ssh.ParsePrivateKey([]byte(password))
```
monotosh-avi, line 187: we can skip the assignment if the return value are not being used.

tests/scaletest/lib/utils.go
RiyaGaniga-avi: Thanks for the idea. Changed it to recursive function

tests/scaletest/scale_test.go
RiyaGaniga-avi: Done

monotosh-avi (approved):

chauhanshubham:
LGTM. could you please run `go mod vendor` once @RiyaGaniga-avi

RiyaGaniga-avi:
> LGTM. could you please run `go mod vendor` once @RiyaGaniga-avi

I already ran it @chauhanshubham.

hemantavi (dismissed):

RiyaGaniga-avi:
@chauhanshubham Thanks for pointing that out. Some files were left out. Its updated now

chauhanshubham (approved):

monotosh-avi (approved):

PR #493, opened 2021-07-09
Add public ip support for public cloud

aaha97:
Tests done:
create ingress/ lb service with public ip enabled
disabling public ip from aviinfrasetting deletes public ip
enabling public ip on existing aviinfrasetting adds public ip
removing annotation from lb service removes public ip

### DIFF
--- a/internal/nodes/avi_model_evh_nodes.go
+++ b/internal/nodes/avi_model_evh_nodes.go
@@ -1600,8 +1600,13 @@ func buildWithInfraSettingForEvh(key string, vs *AviEvhVsNode, vsvip *AviVSVIPNo
 		} else {
 			vsvip.VipNetworks = lib.GetVipNetworkList()
 		}
+		enablePublicIP := false
+		if infraSetting.Spec.Network.EnablePublicIP != nil {
+			enablePublicIP = *infraSetting.Spec.Network.EnablePublicIP
+		}
+		vsvip.EnablePublicIP = enablePublicIP
+		utils.AviLog.Debugf("key: %s, msg: Applied AviInfraSetting configuration over VSNode %s", key, vs.Name)
 	}
-	utils.AviLog.Debugf("key: %s, msg: Applied AviInfraSetting configuration over VSNode %s", key, vs.Name)
 }
 
 func manipulateEvhNodeForSSL(vsNode *AviEvhVsNode, evhNode *AviEvhVsNode) {
--- a/internal/nodes/avi_model_evh_nodes.go
+++ b/internal/nodes/avi_model_evh_nodes.go
@@ -1601,7 +1601,7 @@ func buildWithInfraSettingForEvh(key string, vs *AviEvhVsNode, vsvip *AviVSVIPNo
 			vsvip.VipNetworks = lib.GetVipNetworkList()
 		}
 		enablePublicIP := false
-		if infraSetting.Spec.Network.EnablePublicIP != nil {
+		if infraSetting.Spec.Network.EnablePublicIP != nil && lib.IsPublicCloud() {
 			enablePublicIP = *infraSetting.Spec.Network.EnablePublicIP
 		}
 		vsvip.EnablePublicIP = enablePublicIP
--- a/internal/nodes/avi_model_evh_nodes.go
+++ b/internal/nodes/avi_model_evh_nodes.go
@@ -1600,11 +1600,9 @@ func buildWithInfraSettingForEvh(key string, vs *AviEvhVsNode, vsvip *AviVSVIPNo
 		} else {
 			vsvip.VipNetworks = lib.GetVipNetworkList()
 		}
-		enablePublicIP := false
 		if infraSetting.Spec.Network.EnablePublicIP != nil && lib.IsPublicCloud() {
-			enablePublicIP = *infraSetting.Spec.Network.EnablePublicIP
+			vsvip.EnablePublicIP = infraSetting.Spec.Network.EnablePublicIP
 		}
-		vsvip.EnablePublicIP = enablePublicIP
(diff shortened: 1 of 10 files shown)


internal/nodes/avi_model_nodes.go
```diff
@@ -915,6 +915,7 @@ type AviVSVIPNode struct {
 	SubnetIP                string
 	SubnetPrefix            int32
 	NetworkNames            []string
+	EnablePublicIP          *bool
```
chauhanshubham, line 918: this will need to be part of the checksum calculation. Also could you please add this field's checks in the existing AviInfraSetting UTs too.

internal/rest/avi_obj_vsvip.go
```diff
@@ -124,8 +125,9 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, vsCach
 	} else {
 		var vips []*avimodels.Vip
 		vip := avimodels.Vip{
-			VipID:          &vipId,
-			AutoAllocateIP: &autoAllocate,
+			VipID:                  &vipId,
+			AutoAllocateIP:         &autoAllocate,
+			AutoAllocateFloatingIP: &vsvip_meta.EnablePublicIP,
```
sudswasavi, line 130: does this API through an error if you are doing this on the vCenter cloud? Or it just assumes that the feature isn't supported?

sudswasavi (commented):
Couple of points:
- It looks like we are assuming that this feature will work as is in all types of cloud. Is that true? Or we need explicit qualification? If this is only targeted for AWS then we should restrict the feature to AWS cloud for now.
- Have you verified this feature for layer 4 LB? The aviinfrasetting would be annotation on the service in that case.

internal/nodes/avi_model_nodes.go
aaha97: updated

internal/rest/avi_obj_vsvip.go
```diff
@@ -77,8 +77,9 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, vsCach
 
 		// handling static IP and networkName (infraSetting) updates.
 		vip := &avimodels.Vip{
-			VipID:          &vipId,
-			AutoAllocateIP: &autoAllocate,
+			VipID:                  &vipId,
+			AutoAllocateIP:         &autoAllocate,
+			AutoAllocateFloatingIP: &vsvip_meta.EnablePublicIP,
```
apalsule, line 82: Can you make sure this works with Essentials license, without the public IP (for regressions)

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -1624,8 +1624,13 @@ func buildWithInfraSettingForEvh(key string, vs *AviEvhVsNode, vsvip *AviVSVIPNo
 				vsvip.VipNetworks = vipNetworks
 			}
 		}
+		enablePublicIP := false
+		if infraSetting.Spec.Network.EnablePublicIP != nil && lib.IsPublicCloud() {
```
apalsule, line 1628: Can you add a cloud check for this ? public ip can only be enabled with public clouds and Openstack

apalsule:
@aaha97 Do you handle the case where public ip is removed from the config, so AKO has to delete it. Does this field need to be included for checksum calculations ?

internal/rest/avi_obj_vsvip.go
```diff
@@ -523,14 +525,15 @@ func (rest *RestOperations) AviVsVipCacheDel(rest_op *utils.RestOp, vsKey avicac
...
 	var vipList []*avimodels.Vip
 	autoAllocate := true
 	for vipIDInt, vipNetwork := range vipNetworks {
 		vipID := strconv.Itoa(vipIDInt + 1)
 		newVip := &avimodels.Vip{
-			VipID:          &vipID,
-			AutoAllocateIP: &autoAllocate,
+			VipID:                  &vipID,
+			AutoAllocateIP:         &autoAllocate,
+			AutoAllocateFloatingIP: &enablePublicIP,
```
chauhanshubham, line 536: Is this a network specific field? @aaha97 
Do we see this getting enabled for some networks specifically in multivip scenarios? I think in that case we'll have to have this as follows
in values.yaml
```
vipNetwokrList:
  - networkName: net1
     enablePublicIP: true
```

and in infrasetting 
```
spec:
  network:
    vipNetworks:
    - networkName: net123
       enablePublicIP: true  
```

The new schema just got in (https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/commit/6a13b776a52e458084b2b1696b787e5a91642c3b) so you might want to rebase.

aaha97: i believe we won't be having the enablePublicIP flag as part of values.yaml, only via aviinfrasetting
all networks will have public ip enabled in case of multi vip

internal/rest/avi_obj_vsvip.go
aaha97: done, works fine

aaha97:
build ako

internal/nodes/avi_model_nodes.go
```diff
@@ -704,6 +704,9 @@ func (v *AviVsNode) CalculateCheckSum() {
 	if v.EnableRhi != nil {
 		checksum += utils.Hash(utils.Stringify(*v.EnableRhi))
 	}
+	for _, vsvipref := range v.VSVIPRefs {
```
chauhanshubham, line 707: @aaha97 could you please rebase with the current master. It now has changes related to graph checksum computation, so including this in the Vs Checksum calculation should not be required anymore.
cc: @sudswasavi

aaha97: done

chauhanshubham (approved):

sudswasavi (approved):

PR #494, opened 2021-07-14
adding additional logs for raw avi api queries outside rest layer

chauhanshubham:
This also adds a few checks around executing CRD clientset calls
if the CRD is not installed on the cluster.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -2503,13 +2503,15 @@ func validateAndConfigureSeGroup(client *clients.AviClient) bool {
 		return false
 	}
 
-	infraSettings, err := lib.GetCRDClientset().AkoV1alpha1().AviInfraSettings().List(context.TODO(), metav1.ListOptions{})
-	if err != nil {
-		utils.AviLog.Warnf("Unable to list AviInfraSettings %s", err.Error())
-	}
 	seGroupSet := make(map[string]bool)
-	for _, setting := range infraSettings.Items {
-		seGroupSet[setting.Spec.SeGroup.Name] = true
+	if lib.GetAviInfraSettingEnabled() {
+		infraSettingList, err := lib.GetCRDClientset().AkoV1alpha1().AviInfraSettings().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			utils.AviLog.Warnf("Unable to list AviInfraSettings %s", err.Error())
+		}
+		for _, setting := range infraSettingList.Items {
+			seGroupSet[setting.Spec.SeGroup.Name] = true
+		}
 	}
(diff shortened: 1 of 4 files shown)


internal/cache/controller_obj_cache.go
```diff
@@ -2503,13 +2503,15 @@ func validateAndConfigureSeGroup(client *clients.AviClient) bool {
...
-	if err != nil {
-		utils.AviLog.Warnf("Unable to list AviInfraSettings %s", err.Error())
-	}
 	seGroupSet := make(map[string]bool)
-	for _, setting := range infraSettings.Items {
-		seGroupSet[setting.Spec.SeGroup.Name] = true
+	if lib.GetAviInfraSettingEnabled() {
+		infraSettingList, err := lib.GetCRDClientset().AkoV1alpha1().AviInfraSettings().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			utils.AviLog.Warnf("Unable to list AviInfraSettings %s", err.Error())
```
monotosh-avi, line 2510: I think we should skip in case of an error.

chauhanshubham: the items slice is empty in case of errors, so it's pretty much thee same thing as skipping the operation.

monotosh-avi: In that case it is fine. Though there can be multiple types of errors which we probably can't test and if in any case the infraSettingList is null we might face a problem. So I thought  I though we can do an error check explicitly.

monotosh-avi (approved):

internal/k8s/crdcontroller.go
```diff
@@ -83,187 +83,194 @@ func (c *AviController) SetupAKOCRDEventHandlers(numWorkers uint32) {
...
-			if c.DisableSync {
-				return
-			}
-			hostrule, ok := obj.(*akov1alpha1.HostRule)
-			if !ok {
-				tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
-				if !ok {
-					utils.AviLog.Errorf("couldn't get object from tombstone %#v", obj)
+			},
+			UpdateFunc: func(old, new interface{}) {
```
akshayhavile, line 102: We should not allow update if sync is disabled. (c.DisableSync -> true)

akshayhavile: Same comment for remaining controller update function.

chauhanshubham:
build ako

chauhanshubham:
triggered rebuild due to error while initializing fake api server during unit tests.
```
09:51:24 2021-07-19T04:21:14.919Z	[34mINFO[0m	api/api_fake.go:45	Fake API server now running on port 54321
09:51:24 fatal error: concurrent map writes
```

akshayhavile (approved):

PR #495, opened 2021-07-14
Scale Test - Test case to check for any redundant/unwanted API calls to AVI on AKO reboot

RiyaGaniga-avi:
Adding test case to Scale test to check for any unwanted/redundant API calls to AVI controller on AKO reboot. 
An unwanted API call is one where only the Last Modified or Cloud Config Checksum fields of objects are updated without an update of any other fields
Config update event logs are fetched from the controller after AKO reboots. These logs are scanned for any unwanted API calls.

The second commit disables swap on a rebooted Kubernetes node. Without this, Kubelet fails to start after a node is rebooted.

### DIFF
--- a/tests/scaletest/lib/ingress.go
+++ b/tests/scaletest/lib/ingress.go
@@ -19,7 +19,6 @@ import (
 	"flag"
 	"strconv"
 	"testing"
-	"time"
 
 	appsV1 "k8s.io/api/apps/v1"
 	coreV1 "k8s.io/api/core/v1"
@@ -458,13 +457,13 @@ func DeletePod(podName string, namespace string) error {
 	return nil
 }
 
-func WaitForAKOPodReboot(t *testing.T, akoPodName string) {
+func WaitForAKOPodReboot(t *testing.T, akoPodName string) bool {
 	t.Logf("Waiting for AKO pod...")
 	pod, _ := coreV1Client.Pods(AVISYSTEM).Get(ctx, akoPodName, metaV1.GetOptions{})
 	if pod.Status.Phase != coreV1.PodRunning {
-		time.Sleep(5 * time.Second)
-		WaitForAKOPodReboot(t, akoPodName)
+		return false
 	}
+	return true
 }
 
 func KubeInit(kubeconfig string) {
--- a/tests/scaletest/lib/ingress.go
+++ b/tests/scaletest/lib/ingress.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"strconv"
 	"testing"
+	"time"
 
 	appsV1 "k8s.io/api/apps/v1"
 	coreV1 "k8s.io/api/core/v1"
@@ -45,6 +46,7 @@ const SUBDOMAIN = ".avi.internal"
 const SECRETNAME = "ingress-host-tls"
 const INGRESSAPIVERSION = "networking.k8s.io/v1"
 const PATHTYPE = "Prefix"
+const AVISYSTEM = "avi-system"
 
 func CreateApp(appName string, namespace string, replica int) error {
 	deploymentSpec := &appsV1.Deployment{
@@ -454,6 +456,15 @@ func DeletePod(podName string, namespace string) error {
 		return err
 	}
 	return nil
+}
+
+func WaitForAKOPodReboot(t *testing.T, akoPodName string) {
+	t.Logf("Waiting for AKO pod...")
+	pod, _ := coreV1Client.Pods(AVISYSTEM).Get(ctx, akoPodName, metaV1.GetOptions{})
+	if pod.Status.Phase != coreV1.PodRunning {
+		time.Sleep(5 * time.Second)
+		WaitForAKOPodReboot(t, akoPodName)
+	}
 }
 
 func KubeInit(kubeconfig string) {
--- a/tests/scaletest/lib/utils.go
+++ b/tests/scaletest/lib/utils.go
@@ -17,7 +17,9 @@ package lib
 import (
 	"encoding/json"
 	"os"
+	"reflect"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 
@@ -244,7 +246,7 @@ func FetchVirtualServiceOperStatus(t *testing.T, AviClient *clients.AviClient) [
 	}
 	for result.Next != "" {
 		page_num = page_num + 1
-		uri := "/api/virtualservice-inventory??page=" + strconv.Itoa(page_num)
+		uri := "/api/virtualservice-inventory?page=" + strconv.Itoa(page_num)
 		result, err = AviClient.AviSession.GetCollectionRaw(uri)
 		if err != nil {
 			t.Errorf("Get uri %v returned err for VS %v", uri, err)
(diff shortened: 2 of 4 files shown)


RiyaGaniga-avi:
build ako

RiyaGaniga-avi:
build ako

tests/scaletest/lib/utils.go
```diff
@@ -281,3 +283,187 @@ func FetchOPERDownVirtualService(t *testing.T, AviClient *clients.AviClient) []V
...
+	data = strings.Replace(data, "\\", "", -1)
+	data = strings.Replace(data, " ", "", -1)
+	return data
+}
+
+func CompareVirtualServiceResources(t *testing.T, eventLog models.EventLog) bool {
+	var new, old models.VirtualService
+	err := json.Unmarshal([]byte(CleanResourceData(*eventLog.EventDetails.ConfigUpdateDetails.NewResourceData)), &new)
+	if err != nil {
+		t.Logf("Error unmarshalling data into VS. Error : %v", err)
```
monotosh-avi, line 297: I think we should return here.

tests/scaletest/lib/ingress.go
```diff
@@ -456,6 +458,15 @@ func DeletePod(podName string, namespace string) error {
 	return nil
 }
 
+func WaitForAKOPodReboot(t *testing.T, akoPodName string) {
+	t.Logf("Waiting for AKO pod...")
+	pod, _ := coreV1Client.Pods(AVISYSTEM).Get(ctx, akoPodName, metaV1.GetOptions{})
+	if pod.Status.Phase != coreV1.PodRunning {
+		time.Sleep(5 * time.Second)
```
monotosh-avi, line 465: If WaitForAKOPodReboot is already waiting for the Pod, then we don't need the sleep here.

RiyaGaniga-avi: As this is a recursive function, sleep has been added here to limit the number of recursive calls.

tests/scaletest/lib/utils.go
RiyaGaniga-avi: Thanks for pointing that out. Will change it to Fatalf

tests/scaletest/lib/ingress.go
monotosh-avi: I missed that. In that case may be we can use g.Eventually() like other test cases here.

tests/scaletest/lib/ingress.go
```diff
@@ -456,6 +458,15 @@ func DeletePod(podName string, namespace string) error {
...
+		WaitForAKOPodReboot(t, akoPodName)
```
hemantavi, line 466: What if because of some reason, the pod never starts? Do we timeout in such instances because the recursion can make the stack overflow in such cases. If not already checking for such timeouts, you can simply use the `context` package to cancel the function as soon as a deadline has passed. See here: https://pkg.go.dev/context#WithCancel

tests/scaletest/lib/utils.go
```diff
@@ -281,3 +283,187 @@ func FetchOPERDownVirtualService(t *testing.T, AviClient *clients.AviClient) []V
...
+	}
+	err = json.Unmarshal([]byte(CleanResourceData(*eventLog.EventDetails.ConfigUpdateDetails.OldResourceData)), &old)
+	if err != nil {
+		t.Logf("Error unmarshalling data into VS. Error : %v", err)
```
hemantavi: We should return or do `Fatalf` here too.

tests/scaletest/lib/utils.go
```diff
@@ -281,3 +283,187 @@ func FetchOPERDownVirtualService(t *testing.T, AviClient *clients.AviClient) []V
...
+	}
+	// Genuine update of VS object
+	return true
+}
+
+func ComparePoolResources(t *testing.T, eventLog models.EventLog) bool {
+	var new, old models.Pool
+	err := json.Unmarshal([]byte(CleanResourceData(*eventLog.EventDetails.ConfigUpdateDetails.NewResourceData)), &new)
+	if err != nil {
+		t.Logf("Error unmarshalling data into Pool. Error : %v", err)
```
hemantavi: `Fatalf` or `return` for all such `err != nil` conditions here and below too.

tests/scaletest/scale_test.go
```diff
@@ -262,6 +277,13 @@ func CheckReboot(t *testing.T, wg *sync.WaitGroup) {
 		}
 		json.Unmarshal(byteValue, &testbedParams)
 		go Reboot(t, wg, KUBENODE, testbedParams.AkoParam.Clusters[0].KubeNodes[0].IP, testbedParams.AkoParam.Clusters[0].KubeNodes[0].UserName, testbedParams.AkoParam.Clusters[0].KubeNodes[0].Password, 0)
+		// Disable swap on the rebooted node
+		g := gomega.NewGomegaWithT(t)
+		time.Sleep(20 * time.Second)
```
hemantavi, line 282: Any reason we have a 20s sleep (before doing remote execute) here when we already have the 20s interval for `Eventually` too?

tests/scaletest/scale_test.go
```diff
@@ -1097,3 +1119,21 @@ func TestServiceTypeLB(t *testing.T) {
...
+	g := gomega.NewGomegaWithT(t)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	// Fetch the controller time before AKO reboot
+	// Use this time as sometimes the time on the controller and test client might be out of sync
+	startTime := FetchControllerTime(t)
+	RebootAko(t, &wg)
+	lib.WaitForAKOPodReboot(t, akoPodName)
+	// Wait for 2 min for AKO to sync
+	time.Sleep(2 * time.Minute)
```
hemantavi, line 1134: These unconditional sleeps add to the overall run time. Can we at least make this sleep conditional? The reason is, AKO may not have synced or may not have booted up during that time frame.

vendor/github.com/avinetworks/sdk/go/models/event_log.go
```diff
@@ -56,7 +56,7 @@ type EventLog struct {
 
 	// Number of report_timestamp.
 	// Required: true
-	ReportTimestamp *int64 `json:"report_timestamp"`
+	ReportTimestamp *string `json:"report_timestamp"`
```
hemantavi, line 59: Is this something that was added as part of `go mod vendor` or manually edited?

RiyaGaniga-avi: This is a manual change. The issue has been reported to concerned team.

sudswasavi (commented):
Can we also add:
- SSLKeyCertificate
- httppolicysets 

into the check for changed objects?

Otherwise, it's a great start.

RiyaGaniga-avi:
> Can we also add:
> 
> * SSLKeyCertificate
> * httppolicysets
> 
> into the check for changed objects?
> 
> Otherwise, it's a great start.

Done!

hemantavi (approved):
LGTM though I have one concern regarding the vendor changes, since they were manually edited. This is another change that we have to keep track of.

monotosh-avi (approved):

sudswasavi (approved):

sudswasavi:
Thank you @RiyaGaniga-avi

PR #497, opened 2021-07-14
EVH: Remove namespace field from child VS name

akshayhavile:
This PR
1. Removes namespace from EVH Child VS name.
2. Handles upgrade scenario. Delete existing child EVH vs and objects related with it. This existing EVH Vs will have name with namespace included into it.
3. EVH IT modifications.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -133,9 +133,10 @@ func (c *AviObjCache) AviObjCachePopulate(client *clients.AviClient, version str
 	return vsCacheCopy, allVsKeys, nil
 }
 
-//This function list EVH VS to be deleted which contain namespace in its name
-func (c *AviObjCache) listEVHChildrenToDelete(vs_cache_obj *AviVsCache, childUuids []string) (map[NamespaceName]bool, []string) {
-	childNSNameToDelete := make(map[NamespaceName]bool)
+//TODO: Deperecate this function in future release.
+//This function list EVH child VS to be deleted which contain namespace in its un-encoded name.
+func (c *AviObjCache) listEVHChildrenToDelete(vs_cache_obj *AviVsCache, childUuids []string) ([]NamespaceName, []string) {
+	var childNSNameToDelete []NamespaceName
 	var childUuidToDelete []string
 	for _, childUuid := range childUuids {
 		childKey, childFound := c.VsCacheLocal.AviCacheGetKeyByUuid(childUuid)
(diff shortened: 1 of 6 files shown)


internal/cache/controller_obj_cache.go
```diff
@@ -148,17 +174,25 @@ func (c *AviObjCache) PopulateVsMetaCache() {
 			vs_cache_obj, foundvs := vsObj.(*AviVsCache)
 			if foundvs {
 				vs_cache_obj.ReplaceSNIChildCollection(sniChildUuids)
+
+				if isEVHEnabled {
+					childNSNameToDelete, childUuidToDelete = c.listEVHChildrenToDelete(vs_cache_obj, sniChildUuids)
```
monotosh-avi, line 179: For multiple parentsVSes, childUuidToDelete would contain children of last parent VS only. This would be a problem. 
Also, do we need to handle the upgrade case, as evh was in tech preview mode only ? @sudswasavi

sudswasavi (approved):

internal/cache/controller_obj_cache.go
```diff
@@ -133,11 +133,37 @@ func (c *AviObjCache) AviObjCachePopulate(client *clients.AviClient, version str
 	return vsCacheCopy, allVsKeys, nil
 }
 
+//This function list EVH VS to be deleted which contain namespace in its name
+func (c *AviObjCache) listEVHChildrenToDelete(vs_cache_obj *AviVsCache, childUuids []string) (map[NamespaceName]bool, []string) {
```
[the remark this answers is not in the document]
akshayhavile: Will add comment.

internal/cache/controller_obj_cache.go
akshayhavile: @monotosh-avi : We can deprecate this function after 1-2 releases. But if we do not delete existing non-encoded EVH Vs then it would not allow to create new encoded VS (As encoded VS will have same host, path). 
Regarding comment of childUuidToDelete, children VSes will be associated with Shared Vs. These child VS will have name without encoding. So we will delete them all with this. I am not getting problem here.

monotosh-avi: I was thinking of a case where there are 8 parents VSes  (vs-1 to vs-8 ) each with 10 child VSes .. so we have to delete all 80 children.
As we start going through all the parents, we would add the children .. so after going though vs-1 we would add 8 children to childUuidToDelete. But once we go through vs-2 we would populate childUuidToDelete with the children of vs-2 and replace children of vs-1. I think we need to append the children.

internal/cache/controller_obj_cache.go
```diff
@@ -133,11 +133,38 @@ func (c *AviObjCache) AviObjCachePopulate(client *clients.AviClient, version str
 	return vsCacheCopy, allVsKeys, nil
 }
 
+//TODO: Deperecate this function in future release.
+//This function list EVH child VS to be deleted which contain namespace in its un-encoded name.
+func (c *AviObjCache) listEVHChildrenToDelete(vs_cache_obj *AviVsCache, childUuids []string) ([]NamespaceName, []string) {
```
akshayhavile, line 138: @sudswasavi @monotosh-avi @chauhanshubham : Can you please check this function from GC point of view? will this objects or objects declared in parent function will occupy mem even after function call exited?

monotosh-avi: i think these variables are local to the function .. so should get cleaned up. 
Though @chauhanshubham had observed that there are some memory optimisations that has to be done. So we need to take a look at that in general.

monotosh-avi (approved):

sudswasavi (approved):

PR #498, opened 2021-07-17
Fix EVH L4 service deletion

akshayhavile:
This PR handles encoding of service names/model names
1.  For L4 Virtual Service deletion case
2. For Adv L4 service 
3. For Service transition from loadbalancer type to cluster type.

### DIFF
--- a/internal/nodes/dequeue_ingestion.go
+++ b/internal/nodes/dequeue_ingestion.go
@@ -89,7 +89,7 @@ func DequeueIngestion(key string, fullsync bool) {
 		if found {
 			objects.SharedlbLister().Delete(namespace + "/" + name)
 			utils.AviLog.Infof("key: %s, msg: service transitioned from type loadbalancer to ClusterIP or NodePort, will delete model", name)
-			model_name := lib.GetModelName(lib.GetTenant(), lib.GetNamePrefix()+namespace+"-"+name)
+			model_name := lib.GetModelName(lib.GetTenant(), lib.Encode(lib.GetNamePrefix()+namespace+"-"+name, lib.L4VS))
 			objects.SharedAviGraphLister().Save(model_name, nil)
 			if !fullsync {
 				PublishKeyToRestLayer(model_name, key, sharedQueue)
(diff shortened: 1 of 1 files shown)


chauhanshubham (approved):
lgtm @akshayhavile thanks. I was just seeing this locally :)

internal/nodes/dequeue_ingestion.go
```diff
@@ -89,7 +89,7 @@ func DequeueIngestion(key string, fullsync bool) {
 		if found {
 			objects.SharedlbLister().Delete(namespace + "/" + name)
 			utils.AviLog.Infof("key: %s, msg: service transitioned from type loadbalancer to ClusterIP or NodePort, will delete model", name)
-			model_name := lib.GetModelName(lib.GetTenant(), lib.GetNamePrefix()+namespace+"-"+name)
+			model_name := lib.GetModelName(lib.GetTenant(), lib.Encode(lib.GetNamePrefix()+namespace+"-"+name, lib.Service))
```
chauhanshubham, line 92: Not sure about this. If the service is found in the LB lister then the model that is marked nil will be the L4 service model (as the comment also mentions). So I guess the objectType should be `lib.L4VS` instead of `lib.Service`. This is a l4 VS deletion case.
cc: @monotosh-avi ptal.

sudswasavi:
@akshayhavile I am a bit confused with the title of this PR. It says EVH L4?
L4 VSes are independent of EVH model right? EVH Is only for L7.
Also touching advancedL4 code will have implications on customers who are already using this feature? cc @chauhanshubham

internal/nodes/dequeue_ingestion.go
monotosh-avi: Yes looks like it should be lib.L4VS

chauhanshubham:
> Also touching advancedL4 code will have implications on customers who are already using this feature? cc @chauhanshubham

We have incorporated object encoding for service APIs (in case of advancedL4 too). The implications will be similar to regular AKO customers using non-encoded conventions. @sudswasavi

akshayhavile:
build ako

chauhanshubham (approved):

monotosh-avi (approved):

PR #499, opened 2021-07-19
Add Marker support to Avi Objects

akshayhavile:
This PR handles
1. Adding Markers to avi objects
2. Removing markers from shared objects

Note: Passthrough objects are shared so it will not be marked

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -732,7 +732,7 @@ func ObjectLabelChecksum(objectLabels []*models.RoleFilterMatchLabel) uint32 {
 	markersStr := make([]string, len(objectLabels))
 	var j int
 	var combinedString string
-	//Add all markers values to form cksum except for clusternamelabel for bkward compatibility
+	//For shared objects, checksum will be of only cluster label
 	for _, label := range objectLabels {
 		if *label.Key == clusterKey {
 			if label.Values != nil && len(label.Values) > 0 && label.Values[0] == clusterValue {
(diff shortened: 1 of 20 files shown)


sudswasavi:
build ako

akshayhavile:
build ako

sudswasavi (approved):

akshayhavile:
@sudswasavi @monotosh-avi @chauhanshubham : Please review PR. PR has been modified to label shared objects.

internal/lib/lib.go
```diff
@@ -671,20 +671,59 @@ func SetClusterLabelChecksum() {
...
+			value := vals.Field(i).Interface().(string)
+			markersStr[j] = value
+			j = j + 1
+		}
+	}
+	sort.Strings(markersStr)
+	for _, ele := range markersStr {
+		combinedString += ele
+	}
+	if len(combinedString) != 0 {
```
sudswasavi, line 695: why are you doing a length check on a string variable?

akshayhavile: @sudswasavi : GetMarkersChecksum function is common for shared objects and individual objects. For Shared objects "combinedString" will be empty length. This empty length string produces non zero checksum value. This will result in PUT call for existing shared objects even though there is no change in markers to those objects. (On previous release, these objects are marked with only Clustername label.)
To avoid un-necessary PUT call, this check is there.

monotosh-avi (dismissed):

sudswasavi:
hi @akshayhavile can you pls rebase this PR?

sudswasavi (approved):

monotosh-avi (approved):

PR #506, opened 2021-07-21
Configure static routes inside T1 VRFs

sudswasavi:
This commit introduces support for configuring the static routes
to the corresponding VRF contexts for the T1 router inside the controller.

The logic relies on the supplied T1Lr path which is used to figure
out the vrfconext name by talking to the Avi controller.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -2707,7 +2707,7 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
 		return false
 	}
 	// If an NSX-T cloud is configured without a T1LR param, we will disable sync.
-	if vType == "CLOUD_NSXT" {
+	if vType == lib.CLOUD_NSXT {
 		if lib.GetT1LRPath() == "" {
 			utils.AviLog.Errorf("Cloud is configured as NSX-T but the T1 LR mapping is not provided")
 			return false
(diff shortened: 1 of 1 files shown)


sudswasavi:
build ako

monotosh-avi (approved):
We may have to increase the timeout of the failing test case

sudswasavi:
This PR is not needed because in NSX-T with NCP, the PODs would be routable from the T1 by default. So the SEs don't need a static route.

chauhanshubham (dismissed):

sudswasavi:
build ako

internal/cache/controller_obj_cache.go
```diff
@@ -2842,11 +2842,57 @@ func checkAndSetVRFFromNetwork(client *clients.AviClient) bool {
...
-	utils.AviLog.Infof("Setting VRF %s found from network %s", vrfName, networkName)
-	lib.SetVrf(vrfName)
+			vrfName := *vrf.Name
+			if vrf.Attrs != nil {
+				for _, v := range vrf.Attrs {
+					if *v.Key == "tier1path" && *v.Value == lib.GetT1LRPath() {
+						lib.SetVrf(vrfName)
+						utils.AviLog.Infof("Setting VRF %s found that matches the T1Lr %s", vrfName, lib.GetT1LRPath())
+						foundVrf = true
+						break
```
akshayhavile, line 2877: should we exit from outer loop once we get first match? or we are planning to iterate through all items?

internal/cache/controller_obj_cache.go
```diff
@@ -2842,11 +2842,57 @@ func checkAndSetVRFFromNetwork(client *clients.AviClient) bool {
 		utils.AviLog.Infof("Using global VRF for NodePort mode")
 		return true
 	}
+	if lib.GetCloudType() == lib.CLOUD_NSXT && lib.GetServiceType() == "ClusterIP" && lib.GetCNIPlugin() != lib.NCP_CNI {
+		// Here we need to determine the right VRF for this T1LR
+		// The logic is: Get all the VRF context objects from the controller, figure out the VRF that matches the T1LR
+		// Current pagination size is set to 100, this may have to increased if we have more than 100 T1 routers.
+		uri := "/api/vrfcontext?" + "&include_name=true&cloud_ref.name=" + utils.CloudName + "&page_size=100"
```
akshayhavile, line 2849: Should we have separate function to iterate through all items through pagination logic? (even if number of items more than 100, repeatedly calling same function with different page num should serve our purpose)

sudswasavi: 100 VRFs would mean 100 T1s, I agree we should do pagination, as the comment reads - I have kept it for future. But I can include it in this PR too.

sudswasavi: Assuming 100 T1s, it won't be relevant unless there's a huge cluster. But programatically we should deal with it.

akshayhavile: We can take care of that in future PR. We can go ahead with this.

akshayhavile (dismissed):

internal/cache/controller_obj_cache.go
sudswasavi: Makes sense, I will make the change.

akshayhavile (approved):

chauhanshubham (approved):

PR #508, opened 2021-07-22
Allow Marker based selection of SEGroup and VIPNetwork

chauhanshubham:
This commit helps AKO fetch an SE group and vip Network candidate
by checking all available se groups and usable networks configured in
the Cloud's IPAM, for Markers. The Marker must be of the type
`markers:  [{'key': 'clustername', 'values': ['<clustername>']}]`
In case AKO does not find any configured markers, it defaults back to
the regular behaviour. User input is given priority in both cases.
    
This also increases Avi's maxVersion compatibility to be 20.1.7. As a result
has the new SDK changes too.


sudswasavi:
build ako

chauhanshubham:
It'll be better to not squash and merge this one, keeping the feature commit and alb-sdk inclusion commits separate.

internal/cache/controller_obj_cache.go
```diff
@@ -2523,23 +2527,6 @@ func checkRequiredValuesYaml() bool {
...
-	if segName == "" {
-		utils.AviLog.Warnf("Service Engine Group: serviceEngineGroupName not set in values.yaml, skipping sync.")
-		return false
-	}
-
-	// List all service engine groups, for every SEG check for AviInfraSettings,
-	// if AviInfraSetting NOT found, remove label if exists,
-	// if AviInfraSetting found, configure label if doesn't exist.
-	// This takes care of syncing SeGroup label settings during reboots.
 	uri := "/api/serviceenginegroup/?include_name&page_size=100&cloud_ref.name=" + utils.CloudName
```
sudswasavi, line 2530: A note on pagination here too would be good.

internal/cache/controller_obj_cache.go
```diff
@@ -2554,6 +2541,51 @@ func validateAndConfigureSeGroup(client *clients.AviClient) bool {
...
+		}
+	}
+	// 3. Default-SEGroup
+	if seGroupToUse == "" {
+		seGroupToUse = lib.DEFAULT_SE_GROUP
+	}
+	lib.SetSEGName(seGroupToUse)
+
+	// Not applicable for NodePort mode / disable route is set as True
+	if lib.GetDisableStaticRoute() {
```
sudswasavi, line 2580: If you move this downwards, then for NodePort mode, you would still run through the above code logic, which is not necessary.

internal/cache/controller_obj_cache.go
```diff
@@ -2554,6 +2541,51 @@ func validateAndConfigureSeGroup(client *clients.AviClient) bool {
 		return false
 	}
 
+	// Indulge in Marker based SEGroup searching only when user input is empty.
+	// SE Group selection priority
+	// 1. User input
+	seGroupToUse := lib.GetSEGNameEnv()
+
+	// 2. Marker based
+	if seGroupToUse == "" {
```
sudswasavi, line 2550: It looks like this feature is only relevant for `advancedL4`, in such a case having a top level filter on advancedL4 will optimise the code.

internal/cache/controller_obj_cache.go
```diff
@@ -2554,6 +2541,51 @@ func validateAndConfigureSeGroup(client *clients.AviClient) bool {
...
+				utils.AviLog.Warnf("Failed to unmarshal data, err: %v", err)
+				continue
+			}
+
+			clusterName := lib.GetClusterName()
+			if lib.GetAdvancedL4() {
+				clusterName = lib.GetClusterID()
+			}
+
+			if len(seg.Markers) == 1 &&
```
sudswasavi, line 2564: You are altering the behaviour of regular AKO flows here. It looks like you are assuming that if the se group is labeled with cluster name in the `marker` interface, we will take that SE group into consideration. This is in conflict to the labelling done on the SE group which is a route table label instead of an object label.

internal/cache/controller_obj_cache.go
```diff
@@ -2763,14 +2802,73 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
+	// 4. vipNetworkList can be empty only in WCP usecases, for all others, mark invalid configuration.
+
+	// 1. User input
+	if vipList, _ := lib.GetVipNetworkListEnv(); len(vipList) > 0 {
+		lib.SetVipNetworkList(vipList)
+		return true
+	}
+
+	// 2. Marker based, not valid for public clouds.
+	if ipamRefUri != nil {
```
sudswasavi, line 2820: Same as above, is there a reason you want to expose this behavior outside of the `advancedL4` tag?

internal/cache/controller_obj_cache.go
```diff
@@ -2763,14 +2802,73 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
+
+		ipam := models.IPAMDNSProviderProfile{}
+		ipamRef := strings.SplitAfter(*ipamRefUri, "/api/")
+		if err := lib.AviGet(client, "/api/"+ipamRef[1], &ipam); err != nil {
+			utils.AviLog.Errorf("Get uri %v returned err %v", ipamRef, err)
+			return false
+		}
+
+		markerNetworkFound := ""
+		for _, usableNetwork := range ipam.InternalProfile.UsableNetworks {
```
sudswasavi, line 2834: It looks like if you loop through all the usable networks to identify which one has the right marker, you will end up making too many GET calls to the controller on bootup.

How about you optimise this code, where you make a single call to fetch all the networks from the controller. 

Then do a in memory calculation of figuring out the corresponding usable network from that and check for the marker API?

internal/cache/controller_obj_cache.go
```diff
@@ -2763,14 +2802,73 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
+			if err := lib.AviGet(client, "/api/"+networkRef[1], &network); err != nil {
+				utils.AviLog.Errorf("Get uri %v returned err %v", networkRef, err)
+				return false
+			}
+
+			if len(network.Markers) == 1 &&
+				*network.Markers[0].Key == lib.ClusterNameLabelKey &&
+				len(network.Markers[0].Values) == 1 &&
+				network.Markers[0].Values[0] == clusterName {
+				markerNetworkFound = *network.Name
```
sudswasavi, line 2846: A info log here would be good.

internal/cache/controller_obj_cache.go
```diff
@@ -2763,14 +2802,73 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
+			}
+		}
+
+		if markerNetworkFound != "" {
+			lib.SetVipNetworkList([]akov1alpha1.AviInfraSettingVipNetwork{{
+				NetworkName: markerNetworkFound,
+			}})
```
sudswasavi, line 2853: Why wouldn't you return here? It looks like you will re-execute the code block in line 2858 even if you found a marker labelled network?

chauhanshubham:
All comments have been addressed @sudswasavi. I had initially written it assuming that we could use it for advancedL4 so that has been changed. Other optimisations have also been take care of. The api has been changed to have a query parametter `name.in` that filters networks based on a list of names.

chauhanshubham:
build ako

chauhanshubham:
build ako

internal/cache/controller_obj_cache.go
```diff
@@ -2523,37 +2527,69 @@ func checkRequiredValuesYaml() bool {
...
 // and configures labels on the SeGroup if not present already
 func validateAndConfigureSeGroup(client *clients.AviClient) bool {
-	// Not applicable for NodePort mode / disable route is set as True
+	// Indulge in Marker based SEGroup searching only when user input is empty.
+	// SE Group selection priority
+	// 1. User input
+	seGroupToUse := lib.GetSEGNameEnv()
+
+	// TODO: pagination
+	uri := "/api/serviceenginegroup/?include_name&page_size=100&cloud_ref.name=" + utils.CloudName
```
sudswasavi, line 2536: Now that I think about it, you must pick this pagination up in the next PR before release. I suspect the use case we are trying to achieve here will drive 200+ SE Groups per controller.

internal/cache/controller_obj_cache.go
```diff
@@ -2523,37 +2527,69 @@ func checkRequiredValuesYaml() bool {
...
+		for _, elem := range elems {
+			seg := models.ServiceEngineGroup{}
+			err = json.Unmarshal(elem, &seg)
+			if err != nil {
+				utils.AviLog.Warnf("Failed to unmarshal data, err: %v", err)
+				continue
+			}
+
+			clusterName := lib.GetClusterName()
+			if lib.GetAdvancedL4() {
```
sudswasavi, line 2563: this check is not needed, since there's already a top level check.

internal/cache/controller_obj_cache.go
```diff
@@ -2763,14 +2805,102 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
+			usableNetworkNames = append(usableNetworkNames, networkRefName[1])
+		}
+
+		if len(usableNetworkNames) == 0 {
+			utils.AviLog.Errorf("No usable network configured in configured cloud's ipam profile.")
+			return false
+		}
+
+		// TODO: pagination
+		uri := "/api/network/?include_name&page_size=100&name.in=" + strings.Join(usableNetworkNames, ",")
```
sudswasavi, line 2849: Again a note on pagination, this has to be picked up.

sudswasavi (approved):
Couple of nits, but I am fine if you pick them up in the next PR. Approved.

internal/cache/controller_obj_cache.go
```diff
@@ -2763,14 +2802,99 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
+	}
+
+	// 2. Marker based (only advancedL4)
+	markerNetworkFound := ""
+	if lib.GetAdvancedL4() && ipamRefUri != nil {
+		// Using clusterID for advl4.
+		clusterName := lib.GetClusterID()
+		ipam := models.IPAMDNSProviderProfile{}
+		ipamRef := strings.SplitAfter(*ipamRefUri, "/api/")
+		if err := lib.AviGet(client, "/api/"+ipamRef[1], &ipam); err != nil {
```
monotosh-avi, line 2826: do we need a length check for ipamref here ?

internal/cache/controller_obj_cache.go
```diff
@@ -2763,14 +2802,99 @@ func checkAndSetCloudType(client *clients.AviClient) bool {
...
+
+	}
+
+	// 3. Empty VipNetworkList (only valid for WCP)
+	if markerNetworkFound == "" {
+		lib.SetVipNetworkList([]akov1alpha1.AviInfraSettingVipNetwork{})
+		return true
+	}
+
+	return false
```
monotosh-avi, line 2890: Would we ever reach here ? 
if markerNetworkFound is not empty then it would be handled in the if block 
if lib.GetAdvancedL4() && ipamRefUri != nil {
otherwise it would be handled in 
if markerNetworkFound == "" {

internal/cache/controller_obj_cache.go
```diff
@@ -2523,37 +2527,66 @@ func checkRequiredValuesYaml() bool {
...
+	// SE Group selection priority
+	// 1. User input
+	seGroupToUse := lib.GetSEGNameEnv()
+
+	// TODO: pagination
+	uri := "/api/serviceenginegroup/?include_name&page_size=100&cloud_ref.name=" + utils.CloudName
+	elems := []json.RawMessage{}
+
+	// 2. Marker based (only advancedL4)
+	if seGroupToUse == "" && lib.GetAdvancedL4() {
```
monotosh-avi, line 2540: If we are checking seGroupToUse == "", then we can return before this if seGroupToUse != ""

sudswasavi (approved):

monotosh-avi (approved):

PR #509, opened 2021-07-23
Fix multihost ingress status update during bootup

chauhanshubham:


chauhanshubham:
**RCA**: `ParseOptionsFromMetadata` is now made to accumulate status update options (hostnames) based on Ingresses/Routes. Multiple entries of StatusUpdate options for same Ingress, were not being accounted for during status syncing while booting up. These multiple entries came up when there were multiple hosts defined in the Ingress (secure/insecure does not matter).

```
This roughly depicts the change in ParseOptionsFromMetadata parsing
Before
    {ingress: default/ing1, hostnames: [foo.com]}, {ingress: default/ing1, hostnames: [bar.com]}
After
    {ingress: default/ing1, hostnames: [foo.com, bar.com]}
```

sudswasavi (commented):
Is this condition detectable via a UT?

internal/status/route_status.go
```diff
@@ -48,17 +47,41 @@ func ParseOptionsFromMetadata(options []UpdateOptions, bulk bool) ([]string, []U
...
+				updateIngressOptions[ingress] = option
+			}
 		}
 	}
-	return objectsToUpdate, updateIngressOptions
+
+	ingressesToUpdate := make([]string, len(updateIngressOptions))
+	i := 0
+	for k := range updateIngressOptions {
+		ingressesToUpdate[i] = k
```
monotosh-avi, line 81: An append would also work here, and we would not need the index i. This is also fine though.

monotosh-avi (approved):

sudswasavi (approved):

PR #512, opened 2021-07-23
Allow specifying hostname per gateway listener

chauhanshubham:
The hostname can be provided using the `HostName` field in gateway listener spec.
If not provided, AKO carries out the autoFQDN routine, based on the
`autoFQDN` field provided during installation. Hostnames are eventually
updated with IPs in respective Services.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -171,6 +171,11 @@ func GetshardSize() uint32 {
 }
 
 func GetL4FqdnFormat() int32 {
+	if GetAdvancedL4() {
+		// disable for advancedL4
+		return 3
+	}
+
 	fqdnFormat := os.Getenv("AUTO_L4_FQDN")
 	enumVal, ok := fqdnEnum[fqdnFormat]
 	if ok {
--- a/internal/nodes/avi_model_advl4_translator.go
+++ b/internal/nodes/avi_model_advl4_translator.go
@@ -164,7 +164,7 @@ func (o *AviObjectGraph) ConstructSvcApiL4VsNode(gatewayName, namespace, key str
 
 		var fqdns []string
 		for _, listener := range gw.Spec.Listeners {
-			autoFQDN := false
+			autoFQDN := true
 			// Honour the hostname if specified corresponding to the listener.
 			if listener.Hostname != nil && string(*listener.Hostname) != "" {
 				fqdns = append(fqdns, string(*listener.Hostname))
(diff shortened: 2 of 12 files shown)


internal/lib/lib.go
```diff
@@ -171,6 +171,11 @@ func GetshardSize() uint32 {
 }
 
 func GetL4FqdnFormat() int32 {
+	if GetAdvancedL4() {
```
sudswasavi, line 174: I think this is fine for now, but we may want to support DNS for advancedL4 too based on some asks.

chauhanshubham:
build ako

chauhanshubham:
git error

chauhanshubham:
build ako

internal/nodes/avi_model_advl4_translator.go
```diff
@@ -147,11 +147,37 @@ func (o *AviObjectGraph) ConstructSvcApiL4VsNode(gatewayName, namespace, key str
...
+			}
+		}
+
+		var fqdns []string
+		for _, listener := range gw.Spec.Listeners {
+			autoFQDN := false
+			// Honour the hostname if specified corresponding to the listener.
+			if listener.Hostname != nil && string(*listener.Hostname) != "" {
+				fqdns = append(fqdns, string(*listener.Hostname))
+				autoFQDN = false
```
sudswasavi, line 171: is there a point of setting this variable to false again? It is already false at Line 167 right?

internal/nodes/avi_model_advl4_translator.go
```diff
@@ -147,11 +147,37 @@ func (o *AviObjectGraph) ConstructSvcApiL4VsNode(gatewayName, namespace, key str
...
+			}
+
+			subDomains := GetDefaultSubDomain()
+			services := listenerSvcMapping[fmt.Sprintf("%s/%d", listener.Protocol, listener.Port)]
+			if subDomains != nil && autoFQDN {
```
sudswasavi, line 176: in what condition will autoFQDN get set to `true`?

internal/lib/lib.go
sudswasavi: It looks like this check is really moot. You could always check for the `GetAdvancedL4()` method and be done with the validation, however if you want to keep it this way - it's okay.

sudswasavi (commented):
Minor nits

internal/lib/lib.go
chauhanshubham: this one is required to not disable autofqdn for advancedL4 since the pool node construction mechanism is common for both serviceAPI and advancedL4.

internal/nodes/avi_model_advl4_translator.go
chauhanshubham: thanks. Changed the initialization to `true`.

sudswasavi (approved):

sudswasavi:
Thanks for taking care of the comments. I suspect, if the code block was passing the tests, then we may need a test case for it which would have detected this condition.

monotosh-avi (approved):

PR #513, opened 2021-07-26
Disable static route syncs for NSX-T cloud

sudswasavi:
NCP ensures that the pod network is routable on the T1 which is used
by the Avi SE for backend connectivity to servers. This means that
the SEs do not need a static route to reach the pods in case of NSX-T
cloud even in clusterIP mode.


sudswasavi:
build ako

chauhanshubham (approved):

hemantavi (approved):

PR #517, opened 2021-07-27
Regression issue fix for multi host ingresses with different IPs.

chauhanshubham:
Introduced in #509, the parsing method aggregated hostnames based on ingresses,
and as result attached multiple hostnames to a single VIP, which is not always
the case. This commit fixes the parsing logic to account for different VIPs
in an ingress too.

### DIFF
--- a/internal/status/route_status.go
+++ b/internal/status/route_status.go
@@ -33,6 +33,9 @@ import (
 )
 
 func ParseOptionsFromMetadata(options []UpdateOptions, bulk bool) ([]string, map[string]UpdateOptions) {
+	// updateIngressOptions holds as its key ingressNS/ingressName/vsIP and tries aggregating multiiple hostnames
+	// from various service metadatas. This ensures that a particular ingress hosting a particular IP,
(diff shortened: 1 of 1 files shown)


sudswasavi:
@chauhanshubham can you pls rebase the branch?

monotosh-avi (approved):

internal/status/route_status.go
```diff
@@ -45,40 +48,42 @@ func ParseOptionsFromMetadata(options []UpdateOptions, bulk bool) ([]string, map
...
 			}
 		}
 	}
 
 	ingressesToUpdate := make([]string, len(updateIngressOptions))
 	i := 0
 	for k := range updateIngressOptions {
-		ingressesToUpdate[i] = k
+		kSlice := strings.Split(k, "/")
+		ingressesToUpdate[i] = kSlice[0] + "/" + kSlice[1]
```
sudswasavi, line 86: This is guaranteed to always yield 2 values? If not, we may through a nil pointer.

sudswasavi (approved):

PR #519, opened 2021-07-28
Regression issue fix for multi host ingresses with different IPs. (cherry-pick)

chauhanshubham:
Introduced in #509, the parsing method aggregated hostnames based on ingresses,
and as result attached multiple hostnames to a single VIP, which is not always
the case. This commit fixes the parsing logic to account for different VIPs
in an ingress too.


chauhanshubham:
build ako

akshayhavile (approved):

sudswasavi (approved):

PR #521, opened 2021-07-30
Support for multi-port services with port number

sudswasavi:
This commit removes the limitation in the current code that required
a portName to be specified in the Ingress object for Services of type
ClusterIP with multiple ports.

Now, we do the following:

- User specifies a port information with a numeric value on the ingress.
- We query the service and find the corresponding targetPort for this port.
- We use the targetPort to find a match with the endpoint list. (Like single port Service)
- Populate the servers using endpoint IPs (No change in behavior)


sudswasavi:
build ako

sudswasavi:
build ako

monotosh-avi (approved):

chauhanshubham (approved):

PR #522, opened 2021-07-30
Ignore 404 error for DELETE operations

monotosh-avi:
In case we get 404 error for DELETE operations, we can ignore them and
execure next set of operations. This would prevent a case where we
skip deletion of few objects due to 404 error on other objects.

### DIFF
--- a/internal/rest/rest_operation.go
+++ b/internal/rest/rest_operation.go
@@ -277,6 +277,9 @@ func AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp) error {
 				utils.AviLog.Warnf("Error in rest operation is not of type AviError, err: %v, %T", op.Err, op.Err)
 			} else if op.Model == "VsVip" && op.Method == utils.RestPut {
 				utils.AviLog.Debugf("Error in rest operation for VsVip Put request.")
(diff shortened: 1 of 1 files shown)


chauhanshubham (approved):

internal/rest/rest_operation.go
```diff
@@ -277,6 +277,9 @@ func AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp) error {
 				utils.AviLog.Warnf("Error in rest operation is not of type AviError, err: %v, %T", op.Err, op.Err)
 			} else if op.Model == "VsVip" && op.Method == utils.RestPut {
 				utils.AviLog.Debugf("Error in rest operation for VsVip Put request.")
+			} else if aviErr.HttpStatusCode == 404 && op.Method == utils.RestDelete {
+				utils.AviLog.Warnf("Error in rest operation, object to be deleted not found, this error would be ignored: %v", op.Err)
```
sudswasavi, line 281: can this log be bettered to:

"Error during rest operation: %s, object of type %s not found in the controller. Ignoring... ", op.Model, op.Err

monotosh-avi: Updated

sudswasavi (approved):

chauhanshubham (approved):

PR #523, opened 2021-07-30
NSX-T VLAN Transport zone support

sudswasavi:
This commit supports the NSX-T VLAN transport zone.
When NSX-T is used with VLAN transport zone, the AKO code should
behave as the vCenter write access code.

This commit, removes the mandatory checks for the NSX-T cloud and
imposes them only if the NSX-T transport zone set to OVERLAY which
can be obtained from the cloud object.


sudswasavi:
build ako

chauhanshubham (approved):

PR #525, opened 2021-08-03
VIP per namespace for EVH in VCF

monotosh-avi:
In VCF, we would create VIP per namespace, adding support for that in this PR.
We would look for an environment variable USE_VCF to enable this feature.

In this case, a mapping between namespace and hostname is saved to find out if
a hostname is present in multiple namespaces.

The parent VS name is modified in this case to include namespace instead of shard.

To do in next PRs
- UTs, Helm chart modification

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -1319,3 +1319,11 @@ func GetK8sMinSupportedVersion() string {
 func GetK8sMaxSupportedVersion() string {
 	return k8sMaxVersion
 }
+
+func IsVCFCluster() bool {
+	vcfCluster := os.Getenv(VCF_CLUSTER)
+	if vcfCluster == "true" {
+		return true
+	}
+	return false
+}
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -1322,8 +1322,5 @@ func GetK8sMaxSupportedVersion() string {
 
 func IsVCFCluster() bool {
 	vcfCluster := os.Getenv(VCF_CLUSTER)
-	if vcfCluster == "true" {
-		return true
-	}
-	return false
+	return vcfCluster == "true"
 }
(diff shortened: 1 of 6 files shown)


monotosh-avi:
build ako

monotosh-avi:
Error in running tests: /dev/stdout: resource temporarily unavailable

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -1402,7 +1413,6 @@ func (o *AviObjectGraph) DeletePoolForHostnameForEvh(vsName, hostname string, ro
 		// Replace the ingress map for this host.
 		keyToRemove := namespace + "/" + ingName
 		delete(ingressHostMap.HostNameMap, keyToRemove)
-		SharedHostNameLister().Save(hostname, ingressHostMap)
```
chauhanshubham, line 1405: How come this is not required anymore?

monotosh-avi: Good catch, deleted by mistake, udpated

monotosh-avi:
error : Observed a panic: runtime.boundsError{x:8, y:8, signed:true, code:0x0} (runtime error: index out of range [8] with length 8)

monotosh-avi:
build ako

internal/lib/lib.go
```diff
@@ -1319,3 +1319,11 @@ func GetK8sMinSupportedVersion() string {
 func GetK8sMaxSupportedVersion() string {
 	return k8sMaxVersion
 }
+
+func IsVCFCluster() bool {
+	vcfCluster := os.Getenv(VCF_CLUSTER)
```
akshayhavile, line 1324: How will this value populated to AKO POD? will by default in VCF each node have this?

internal/lib/lib.go
```diff
@@ -1319,3 +1319,11 @@ func GetK8sMinSupportedVersion() string {
...
+	if vcfCluster == "true" {
```
akshayhavile, line 1325: we can optimise it as --> return vcfCluster == "true" .

akshayhavile (approved):

chauhanshubham (approved):

PR #537, opened 2021-08-10
remove encoding for passthrough VS name

monotosh-avi:
Fixes AV-121039

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -1004,8 +1004,8 @@ func GetPassthroughShardVSName(s string, key string) string {
 	shardVsPrefix := GetClusterName() + "--" + PassthroughPrefix
 	vsNum = utils.Bkt(s, shardSize)
 	vsName := shardVsPrefix + strconv.Itoa(int(vsNum))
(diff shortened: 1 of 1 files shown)


internal/lib/lib.go
```diff
@@ -1004,8 +1004,8 @@ func GetPassthroughShardVSName(s string, key string) string {
 	shardVsPrefix := GetClusterName() + "--" + PassthroughPrefix
 	vsNum = utils.Bkt(s, shardSize)
 	vsName := shardVsPrefix + strconv.Itoa(int(vsNum))
```
akshayhavile, line 1006: We need to "EVH" in vsname . For example, "cluster1--Shared-Passthrough-EVH-0" For SNI, it will be "cluster1--Shared-Passthrough-0" like that.

monotosh-avi: Passthrough VSes don't depend on EVH, so I thought the naming can be same .. similar to L4 VS.

akshayhavile (approved):

chauhanshubham (approved):

PR #542, opened 2021-08-11
Calculate checksum based on VS type

monotosh-avi:
We should not check if EVH is enabled for checksum calculation
fixes AV-121140

### DIFF
--- a/internal/nodes/avi_model_nodes.go
+++ b/internal/nodes/avi_model_nodes.go
@@ -131,16 +131,12 @@ func (v *AviObjectGraph) CalculateCheckSum() {
 	defer v.Lock.Unlock()
 	v.GraphChecksum = 0
 	for _, model := range v.modelNodes {
-		if lib.IsEvhEnabled() {
-			if modelVsNode, ok := model.(*AviEvhVsNode); ok {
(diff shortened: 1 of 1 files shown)


internal/nodes/avi_model_nodes.go
```diff
@@ -131,16 +131,13 @@ func (v *AviObjectGraph) CalculateCheckSum() {
...
-		} else {
-			if modelVsNode, ok := model.(*AviVsNode); ok {
-				v.GraphChecksum = v.GraphChecksum + modelVsNode.CalculateForGraphChecksum()
-				continue
-			}
+		if modelVsNode, ok := model.(*AviEvhVsNode); ok {
+			v.GraphChecksum = v.GraphChecksum + modelVsNode.CalculateForGraphChecksum()
+			continue
+		}
+		if modelVsNode, ok := model.(*AviVsNode); ok {
```
akshayhavile, line 138: Should we have "else if" here?

monotosh-avi: Added.

akshayhavile (approved):

chauhanshubham (approved):
thanks

PR #549, opened 2021-08-12
infinite retry for all failures in rest layer

monotosh-avi:
- Infinite retry for all failures after retry counter is exhausted
- Increased retry counter to 100 from 10

In case rest request times out due to some problem, AKO would retry
all requests. But some of those requests might have been successful in
the controller. In that case AKO can get a 409 error if it tries to recreate
an objects on retry.

In case we get the timeout error object cache is not updated, so we would get
409 error on multiple objects.

(cherry picked from commit 00868d08c769a1d425e306b5a3992a75ac898270)

### DIFF
--- a/internal/nodes/avi_model_nodes.go
+++ b/internal/nodes/avi_model_nodes.go
@@ -108,7 +108,7 @@ func (v *AviObjectGraph) SetRetryCounter(num ...int) {
 	if len(num) > 0 {
 		v.RetryCount = num[0]
 	} else {
-		v.RetryCount = 10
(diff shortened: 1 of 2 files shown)


monotosh-avi:
build ako

monotosh-avi:
error : tenant_write /dev/stdout: resource temporarily unavailable

monotosh-avi:
build ako

sudswasavi (approved):

chauhanshubham (approved):

PR #551, opened 2021-08-14
EVH: Fix route transition secure to insecure issue

akshayhavile:


akshayhavile:
build ako

sudswasavi (approved):
LGTM. @monotosh-avi for further eval.

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -1245,15 +1246,21 @@ func DeleteStaleDataForEvh(routeIgrObj RouteIngressModel, key string, modelList
 			}
 			utils.AviLog.Infof("key: %s, host: %s, currentData: %v", key, host, currentData)
 			removeFqdn = false
+			if routeIgrObj.GetType() == utils.OshiftRoute {
+				diff := lib.GetDiffPath(hostData.PathSvc, currentData.PathSvc)
+				if len(diff) == 0 {
+					removeRouteIngData = false
```
monotosh-avi, line 1252: If a route is changed from secure to insecure and edited to change its path as part of the same operation, then removeRouteIngData be true. Would that be a problem ?

sudswasavi: Great catch.I think so, in that case, the condition would not remove the cert?

akshayhavile: @monotosh-avi , @sudswasavi : It will not create problem as removeRouteIngData is true, it will delete pool and poolgroup associated with older path. Also certificate will removed from VS while building BuildModelGraphForInsecureEVH

monotosh-avi (approved):

PR #554, opened 2021-08-16
Fix EVH: Add sslkeyref to parent EVH VS

akshayhavile:


monotosh-avi (dismissed):

internal/rest/avi_evh_obj.go
```diff
@@ -333,6 +333,9 @@ func (rest *RestOperations) AviVsBuildForEvh(vs_meta *nodes.AviEvhVsNode, rest_m
 					vs.SslKeyAndCertificateRefs = append(vs.SslKeyAndCertificateRefs, certName)
 				}
 			}
+			if vs_meta.SSLProfileRef != "" {
```
monotosh-avi, line 336: Isn't adding sslprofileref in evh child enough ? 
cc @chauhanshubham

chauhanshubham: EVH child nodes do not allow TLS configurations so sslKeyCert and sslProfile needs to go in the parent.

chauhanshubham (approved):

internal/rest/avi_evh_obj.go
akshayhavile: @monotosh-avi : In EVH deployment, as each connection is getting terminated at Parent so SSLKeyCertRef and SSLRef is associated with Parent VS.  At child VS, we have code to add ssl ref and sslkeycertref  but these fields will be empty so it will no op.

monotosh-avi: But our hostrules are per host, won't that create a conflict ?

akshayhavile: Yes. As currently, only single sslprofile reference is allowed, it will create problem for it. SSLkeyandcert can be added multiple to parent VS. So for it, it will not be a problem.

monotosh-avi: Yeah, for sslkeycert its fine, but for ssl profile looks like we can't use the feature for evh mode. Better to document it / ignore it with logs ?

akshayhavile: We will document it

chauhanshubham: I think we should still allow configuring it. It is likely that users can create a common sslProfile for all their applications, which could be a valid configuration. @akshayhavile if it is already set to a different sslProfile by a two distinct HostRules can we add a log for it, and overwrite?

monotosh-avi: @chauhanshubham common ssl profile should be defined at the level of configmap or aviinfrasetting CRD. Properties which are at the shared VS level can not be controlled by Hostrule. 

For now we can print a warning and then - ignore or overwrite - either one is fine. cc @sudswasavi

akshayhavile: Added warning message if we are overwriting older sslprofile value for specific parent vs node.

akshayhavile:
Failure in TestSecureRouteMultiNamespaceInNodePort.

akshayhavile:
build ako

akshayhavile:
build ako

monotosh-avi (approved):

sudswasavi (approved):

PR #581, opened 2021-08-19
Check if value of object in cache is nil before further processing

monotosh-avi:
We were trying to match uuid of cache object without checking if the object is nil.
This can cause segmentation fault. Adding a check to prevent this.

Also fixed another problem where we were saving pointer to an object which was
already a pointer. Fixed this, although we are most likely not hitting this path.

Fixes AV-121646

### DIFF
--- a/internal/cache/cache_utils.go
+++ b/internal/cache/cache_utils.go
@@ -15,6 +15,7 @@ package cache
 
 import (
 	"encoding/json"
+	"reflect"
 	"sync"
 
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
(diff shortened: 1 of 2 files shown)


chauhanshubham (approved):

internal/rest/rest_operation.go
```diff
@@ -268,7 +268,7 @@ func AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp) error {
 			op.Err = fmt.Errorf("Unknown RestOp %v", op.Method)
 		}
 		if op.Err != nil {
-			utils.AviLog.Warnf(`RestOp method %v path %v tenant %v Obj %s returned err %s with response %s`,
+			utils.AviLog.Infof(`RestOp method %v path %v tenant %v Obj %s returned err %s with response %s`,
```
akshayhavile, line 271: @monotosh-avi : This code path is for handling error condition. Will that be ok to change Log level from "Warn" to "Info"? It might get ignored during debugging issue.

monotosh-avi: added by mistake, reverted.

akshayhavile (approved):

internal/rest/rest_operation.go
sudswasavi: Keeping warnf is probably ideal?

sudswasavi (approved):

PR #582, opened 2021-08-19
Remove encoding from passthrough PG name

monotosh-avi:
In case of passthrough with EVH we can’t encode the PG name becasue in l4 data script there is no encode function to use.


akshayhavile (approved):
LGTM

chauhanshubham (approved):

PR #585, opened 2021-08-23
Check if value of object in cache is nil before further processing

monotosh-avi:
We were trying to match uuid of cache object without checking if the object is nil.
This can cause segmentation fault. Adding a check to prevent this.

Also fixed another problem where we were saving pointer to an object which was
already a pointer. Fixed this, although we are most likely not hitting this path.

(cherry picked from commit ffa08e28386b706c8723fbc2230d536ed1956b43)

### DIFF
--- a/internal/cache/cache_utils.go
+++ b/internal/cache/cache_utils.go
@@ -15,6 +15,7 @@ package cache
 
 import (
 	"encoding/json"
+	"reflect"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


monotosh-avi:
build ako

chauhanshubham (approved):

hemantavi (approved):

PR #598, opened 2021-08-27
Introduce ability to record Events from AKO

chauhanshubham:
This adds a wrapper function to record events and initialize
Broadcaster with schemes for which Events can be generated.
This includes non-native k8s objects like
Routes/Gateways/Hostrules/HTTPRules.


chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

monotosh-avi (dismissed):
Looks good to me Shubham. Thanks for making this change, should help in debugging.

internal/lib/control_config.go
```diff
@@ -0,0 +1,236 @@
...
+	hostRuleEnabled bool
+	// httpRuleEnabled is set to true if the cluster has
+	// HTTPRule CRD installed.
+	httpRuleEnabled bool
+}
+
+var akoControlConfigInstance *akoControlConfig
+
+func AKOControlConfig() *akoControlConfig {
+	if akoControlConfigInstance == nil {
```
akshayhavile, line 101: I think it will be good if we create singleton object of akoControlConfig  and use it over all place instead of calling lib.AKOControlConfig each time.

chauhanshubham: that is what it is @akshayhavile. `lib.AKOControlConfig()` is a function to access the singleton. I initialize it only if it is nil.

akshayhavile: This code doesn't seem to be thread safe. If multiple threads(in case) try to access this code, same variable might get initialised twice. But i think impact will be less.

chauhanshubham: I could add a initialize Once block for it, but this gets initilaized in the AKO main function, so getting initialized in multiple goroutines should not happen. Also since it is a global struct, we donot expect it to be initialized from different goroutines.

sudswasavi (approved):

akshayhavile (approved):
LGTM

PR #607, opened 2021-09-01
Handling Bootstrap CRD for NCP to AKO communication

monotosh-avi:
Adding support for boostrap CRD for NCP communication through dynamic client.
In VCF environment, NCP would communicate with AKO using a bootrap CR.
AKO would check for this CR while booting up, if not present,
AKO would wait for this CR to be created before booting up.

The second commit deals with validating the contents of the CR and avi-secret creation.

Communicating back to NCP by deleting the secret if the authtoken is not valid, is handled in the third commit.

fixes AV-120274

### DIFF
--- a/internal/k8s/vcf_controller.go
+++ b/internal/k8s/vcf_controller.go
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2020-2021 VMware, Inc.
+ * All Rights Reserved.
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*   http://www.apache.org/licenses/LICENSE-2.0
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package k8s
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
+
+	"github.com/vmware/alb-sdk/go/clients"
+	"github.com/vmware/alb-sdk/go/session"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+)
+
+func (c *AviController) AddNCPBootstrapEventHandler(k8sinfo K8sinformers, stopCh <-chan struct{}, startSyncCh chan struct{}) {
+	cs := k8sinfo.Cs
+	utils.AviLog.Debugf("Creating event broadcaster for NCP Bootstrap CRD")
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(utils.AviLog.Debugf)
+	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: cs.CoreV1().Events("")})
+	NCPBootstrapHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			utils.AviLog.Infof("NCP Bootstrap ADD Event")
+			if c.ValidBSData() && startSyncCh != nil {
+				c.UpdateAviSecret()
+				startSyncCh <- struct{}{}
+				startSyncCh = nil
+			}
+		},
+		UpdateFunc: func(old, obj interface{}) {
+			utils.AviLog.Infof("NCP Bootstrap Update Event")
+			if c.ValidBSData() && startSyncCh != nil {
+				c.UpdateAviSecret()
+				startSyncCh <- struct{}{}
+				startSyncCh = nil
+			}
+		},
+	}
+	c.dynamicInformers.NCPBootstrapInformer.Informer().AddEventHandler(NCPBootstrapHandler)
+
+	go c.dynamicInformers.NCPBootstrapInformer.Informer().Run(stopCh)
+	if !cache.WaitForCacheSync(stopCh, c.dynamicInformers.NCPBootstrapInformer.Informer().HasSynced) {
+		runtime.HandleError(fmt.Errorf("Timed out waiting for caches to sync"))
+	} else {
+		utils.AviLog.Info("Caches synced for NCP Bootstrap informer")
+	}
+}
+
+// HandleVCF checks if avi secret used by AKO is already present. If found, the it would try to connect to
+// AVI Controller. If there is any failure, we would look at Bootstrap CR used by NCP to communicate with AKO.
+// If Bootstrap CR is not found, AKO would wait for it to be created. If the authtoken from Bootstrap CR
+// can be used to connect to the AVI Controller, then avi-secret would be created with that token.
+func (c *AviController) HandleVCF(informers K8sinformers, stopCh <-chan struct{}, ctrlCh chan struct{}) {
+	cs := c.informers.ClientSet
+	aviSecret, err := cs.CoreV1().Secrets(utils.GetAKONamespace()).Get(context.TODO(), lib.AviSecret, metav1.GetOptions{})
+	if err == nil {
+		ctrlIP := os.Getenv(utils.ENV_CTRL_IPADDRESS)
+		authToken := aviSecret.Data["authToken"]
+		username := aviSecret.Data["username"]
+		var transport *http.Transport
+		_, err = clients.NewAviClient(
+			ctrlIP, string(username), session.SetAuthToken(string(authToken)),
+			session.SetRefreshAuthTokenCallbackV2(utils.GetAuthtokenFromCache),
+			session.SetNoControllerStatusCheck, session.SetTransport(transport),
+			session.SetInsecure,
+		)
+		if err == nil {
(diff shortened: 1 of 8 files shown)


monotosh-avi:
build ako

chauhanshubham (approved):

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	_, err = clients.NewAviClient(
+		ctrlIP, username, session.SetAuthToken(string(authToken)),
+		session.SetNoControllerStatusCheck, session.SetTransport(transport),
+		session.SetInsecure,
+	)
+	if err != nil {
+		utils.AviLog.Infof("Failed to connect to AVI controller using existing AKO secret, err: %v", err)
+		return false
+	}
+	utils.AviLog.Infof("Successfully connected to AVI controller using existing AKO secret")
```
akshayhavile, line 179: I think message needs to be "Successfully connected to Avi controller using NCP secret"

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+// HandleVCF checks if avi secret used by AKO is already present. If found, the it would try to connect to
+// AVI Controller. If there is any failure, we would look at Bootstrap CR used by NCP to communicate with AKO.
+// If Bootstrap CR is not found, AKO would wait for it to be created. If the authtoken from Bootstrap CR
+// can be used to connect to the AVI Controller, then avi-secret would be created with that token.
+func (c *AviController) HandleVCF(informers K8sinformers, stopCh <-chan struct{}, ctrlCh chan struct{}) {
+	cs := c.informers.ClientSet
+	aviSecret, err := cs.CoreV1().Secrets(utils.GetAKONamespace()).Get(context.TODO(), lib.AviSecret, metav1.GetOptions{})
+	if err == nil {
+		ctrlIP := os.Getenv(utils.ENV_CTRL_IPADDRESS)
+		authToken := aviSecret.Data["authToken"]
```
akshayhavile, line 81: Avi secret can contain username and password combination / username and token combination. Are we assuming that avi secret will alway contain authToken? Will this affect upgrade scenario from previous ako with password to this one?

monotosh-avi: In VCF we would always use token to connect to the controller.

internal/k8s/vcf_controller.go
monotosh-avi: Yeah, the msg was wrong, updated

akshayhavile (approved):

internal/lib/dynamic_client.go
```diff
@@ -109,6 +120,34 @@ func GetDynamicInformers() *DynamicInformers {
...
+
+	crdClient := dynamicClient.Resource(BootstrapGVR)
+	crdList, err := crdClient.List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		utils.AviLog.Errorf("Error getting CRD %v", err)
+		return false
+	}
+
+	if len(crdList.Items) > 1 {
+		utils.AviLog.Errorf("Expected ony one object for NCP bootstrap but found: %d", len(crdList.Items))
```
sudswasavi, line 133: typo: only one

internal/lib/dynamic_client.go
```diff
@@ -109,6 +120,34 @@ func GetDynamicInformers() *DynamicInformers {
 	return dynamicInformerInstance
 }
 
+var bootstrapCRFound bool
+
+func BootstrapCRPresent() bool {
+	if bootstrapCRFound {
```
sudswasavi: In what condition does this `if` block return `true`? It seems as though the `BootstrapCRPresent` method is only getting called once during bootup?

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+		UpdateFunc: func(old, obj interface{}) {
+			utils.AviLog.Infof("NCP Bootstrap Update Event")
+			if c.ValidBSData() && startSyncCh != nil {
+				c.UpdateAviSecret()
+				startSyncCh <- struct{}{}
+				startSyncCh = nil
+			}
+		},
+	}
```
sudswasavi, line 99: Is there a reason you are missing the `DeleteFunc`? If the bootstrap CR is deleted, wouldn't you take some action?

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+			ctrlIP, string(username), session.SetAuthToken(string(authToken)),
+			session.SetRefreshAuthTokenCallbackV2(utils.GetAuthtokenFromCache),
+			session.SetNoControllerStatusCheck, session.SetTransport(transport),
+			session.SetInsecure,
+		)
+		if err == nil {
+			utils.AviLog.Infof("Successfully connected to AVI controller using existing AKO secret")
+			return
+		} else {
+			utils.AviLog.Error("AVI controller initialization failed with err: %v", err)
```
sudswasavi, line 131: If you fail to connect using the information provided by the bootstrap CR, what is the action you need to take? Are you going to simply wait for a Secret Update event and then re-run this method? If so, should you return some error at line 95?

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+	}
+
+	utils.AviLog.Infof("Got error while fetching avi-secret: %v", err)
```
sudswasavi, line 135: This log will be printed even if you got the secret but failed to connect to the controller.

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+		}
+	}
+	utils.AviLog.Infof("NCP Bootstrap CR found, continuing AKO initialization")
+	c.UpdateAviSecret()
+}
+
+func (c *AviController) UpdateAviSecret() error {
+	secretName, ns, username := lib.GetBootstrapCRData()
+	cs := c.informers.KubeClientIntf.ClientSet
```
sudswasavi, line 157: is this variable really the clientset or is it the result of the `Get` call?

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+			case <-ctrlCh:
+				return
+			}
+		}
+	}
+	utils.AviLog.Infof("NCP Bootstrap CR found, continuing AKO initialization")
+	c.UpdateAviSecret()
+}
+
+func (c *AviController) UpdateAviSecret() error {
```
sudswasavi, line 155: Re-naming this method as `CreateOrUpdateSecret` is better since it seems to be creating it too.

sudswasavi: Also you are returning an error from this method but the error is not being handled in the caller.

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		utils.AviLog.Warnf("Failed to update avi-secret, err: %v", err)
+		return err
+	}
+
+	return nil
+}
+
+func (c *AviController) ValidBSData() bool {
+	utils.AviLog.Infof("Validating BS data")
```
sudswasavi, line 193: BS just sounds absurd :) Can we call it AKO bootstrap or something?

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	_, err = cs.CoreV1().Secrets(utils.GetAKONamespace()).Update(context.TODO(), &aviSecret, metav1.UpdateOptions{})
+	if err != nil {
+		utils.AviLog.Warnf("Failed to update avi-secret, err: %v", err)
+		return err
+	}
+
+	return nil
+}
+
+func (c *AviController) ValidBSData() bool {
```
sudswasavi, line 192: Is there a reason to keep 2 methods 1. HandleVCF and this one to do the same function of initialisation of session to the Avi Controller or should we converge it into 1?

internal/k8s/vcf_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	utils.AviLog.Infof("Validating BS data")
+	cs := c.informers.ClientSet
+	secretName, ns, username := lib.GetBootstrapCRData()
```
sudswasavi, line 195: There is a possibility that the `GetBootstrapCRData` method will return empty values for any of these 3 return variables, should we handle them appropriately?

internal/k8s/vcf_controller.go
monotosh-avi: Currently we dont have any action If the bootstrap CR is deleted. If the avi-secret is valid, AKO would continue to work.

internal/k8s/vcf_controller.go
monotosh-avi: In case of an error, we would wait for secret/ bootstrap CR update. 
In line no: 95 we are updating the channel after validating that the secret has valid data.

internal/k8s/vcf_controller.go
monotosh-avi: Thanks, I'll update the log in the else condition.

internal/k8s/vcf_controller.go
monotosh-avi: updated to handle the error and shut down AKO in case of an error.

internal/k8s/vcf_controller.go
monotosh-avi: Updated :)

internal/k8s/vcf_controller.go
monotosh-avi: Yes, the initialisation part by calling NewAviClient is same, but rest of the functions are different. HandleVCF internally calls ValidBSData, but ValidBSData also gets called from other functions. 
I could have put the init part in a separate function, but then it looked too small.

internal/k8s/vcf_controller.go
monotosh-avi: Good catch. Updated.

internal/lib/dynamic_client.go
monotosh-avi: updated

internal/k8s/vcf_controller.go
monotosh-avi: This is clientset, but KubeClientIntf is anonymous field - so ClientSet can be directly accessed, updated.

sudswasavi (approved):

sudswasavi:
build ako

akshayhavile (approved):

PR #608, opened 2021-09-03
Ensure AKO reboot on secret Add/Delete

chauhanshubham:


internal/k8s/controller.go
```diff
@@ -1034,6 +1034,9 @@ func validateAviConfigMap(obj interface{}) (*corev1.ConfigMap, bool) {
 
 func validateAviSecret(secret *corev1.Secret) bool {
 	if secret.Namespace == utils.GetAKONamespace() && secret.Name == lib.AviSecret {
+		// if the secret is updated or deleted we shutdown API server
```
monotosh-avi, line 1037: Though currently we are calling validateAviSecret() when secret gets added/updated, one can potentially call function from anywhere, is this the right place to trigger a shutdown ?

chauhanshubham: This is on delete/update secret, specifically for `avi-secret`. do you see any issues around calling this from the event handler itself?

monotosh-avi: Yeah, we can trigger a shutdown for the event handler.

monotosh-avi (approved):

hemantavi (approved):

hemantavi:
build ako

PR #609, opened 2021-09-03
AV-122809: EVH: Hostrule SSLcertkeyref fixes for secure/insecure ingress, routes

akshayhavile:
This PR fixes issue with EVH deployment where 
1. For secure, insecure ingress/ route with hostrule CRDs with TLS configs and Ingress secrets are mixed, then sslkeycertref on parent was getting overwritten by later applied definition.

### DIFF
--- a/internal/nodes/avi_model_evh_nodes.go
+++ b/internal/nodes/avi_model_evh_nodes.go
@@ -363,7 +363,7 @@ func (o *AviEvhVsNode) DeleteCACertRefInEVHNode(cacertNodeName, key string) {
 	for i, cacert := range o.CACertRefs {
 		if cacert.Name == cacertNodeName {
 			o.CACertRefs = append(o.CACertRefs[:i], o.CACertRefs[i+1:]...)
-			utils.AviLog.Infof("key: %s, msg: deleted cacert for evh in model: %s Pool name: %s", key, o.Name, cacert.Name)
+			utils.AviLog.Debugf("key: %s, msg: deleted cacert for evh in model: %s Pool name: %s", key, o.Name, cacert.Name)
 			return
 		}
 	}
--- a/internal/nodes/avi_model_evh_nodes.go
+++ b/internal/nodes/avi_model_evh_nodes.go
@@ -363,7 +363,7 @@ func (o *AviEvhVsNode) DeleteCACertRefInEVHNode(cacertNodeName, key string) {
 	for i, cacert := range o.CACertRefs {
 		if cacert.Name == cacertNodeName {
 			o.CACertRefs = append(o.CACertRefs[:i], o.CACertRefs[i+1:]...)
-			utils.AviLog.Infof("key: %s, msg: replaced cacert for evh in model: %s Pool name: %s", key, o.Name, cacert.Name)
+			utils.AviLog.Infof("key: %s, msg: deleted cacert for evh in model: %s Pool name: %s", key, o.Name, cacert.Name)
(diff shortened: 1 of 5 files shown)


internal/nodes/avi_model_evh_nodes.go
```diff
@@ -1059,7 +1062,18 @@ func (o *AviObjectGraph) BuildModelGraphForSecureEVH(routeIgrObj RouteIngressMod
 		evhNode.ServiceMetadata.Namespace = namespace
 		evhNode.ServiceMetadata.HostNames = hosts
 		if evhNode.SSLKeyCertAviRef != "" {
-			certsBuilt = true
+			//Logic to check host crd with sslcert field present for given host or not.
```
monotosh-avi, line 1065: Matching with dummy secret name (line no - 1037) would verify if the host is referred in a hostrule, so we may not need this check. 
cc @chauhanshubham

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -1083,7 +1097,18 @@ func (o *AviObjectGraph) BuildModelGraphForSecureEVH(routeIgrObj RouteIngressMod
 	}
 	if !certsBuilt {
 		certsBuilt = o.BuildTlsCertNodeForEvh(routeIgrObj.GetSvcLister(), vsNode[0], namespace, tlssetting, key, infraSettingName, host)
+	} else {
+		//Delete sslcertref object if host crd sslcertref (sslcertAviRef) is present for given host
+		secretName := tlssetting.SecretName
+		if strings.HasPrefix(secretName, lib.RouteSecretsPrefix) {
+			// Openshift case
+			vsNode[0].DeleteCACertRefInEVHNode(lib.GetCACertNodeName(infraSettingName, host), key)
```
monotosh-avi, line 1105: In case of openshift route, cacert would be referred by a ssl key cert. I think we would have to remove both.

akshayhavile: Yes. Thanks. Added deletion of CA for openshift and deletion of sslkeycert for both openshift and k8.

monotosh-avi (approved):

monotosh-avi (approved):

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -363,7 +363,7 @@ func (o *AviEvhVsNode) DeleteCACertRefInEVHNode(cacertNodeName, key string) {
 	for i, cacert := range o.CACertRefs {
 		if cacert.Name == cacertNodeName {
 			o.CACertRefs = append(o.CACertRefs[:i], o.CACertRefs[i+1:]...)
-			utils.AviLog.Infof("key: %s, msg: replaced cacert for evh in model: %s Pool name: %s", key, o.Name, cacert.Name)
+			utils.AviLog.Infof("key: %s, msg: deleted cacert for evh in model: %s Pool name: %s", key, o.Name, cacert.Name)
```
sudswasavi, line 366: Can we change this to debug log? Looks like this would be printed for every run.

sudswasavi (commented):
one minor nit.

internal/nodes/avi_model_evh_nodes.go
akshayhavile: Done.

sudswasavi (approved):

chauhanshubham (approved):

PR #622, opened 2021-09-17
Add AKO-Infra and reorganise files

monotosh-avi:
Introducing ako-infra which would be used to create a new container.
Handling NCP Boostrap CRD is moved to AKO-Infra.

### DIFF
--- a/cmd/infra-main/main.go
+++ b/cmd/infra-main/main.go
@@ -1,3 +1,17 @@
+/*
+ * Copyright 2021 VMware, Inc.
+ * All Rights Reserved.
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*   http://www.apache.org/licenses/LICENSE-2.0
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
 package main
 
 import (
@@ -9,7 +23,6 @@ import (
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
 
-	oshiftclient "github.com/openshift/client-go/route/clientset/versioned"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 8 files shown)


cmd/infra-main/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if !lib.IsVCFCluster() {
+		utils.AviLog.Fatalf("Not running in vcf cluster, shutting down")
+	}
+
+	var err error
+	kubeCluster := false
+	utils.AviLog.Info("AKO-Infra is running with version: ", version)
+	// Check if we are running inside kubernetes. Hence try authenticating with service token
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
```
sudswasavi, line 38: is this block of code really necessary? It looks. like we will always run inside the kubernetes cluster?

cmd/infra-main/main.go
```diff
@@ -0,0 +1,108 @@
...
+	} else {
+		serverVersion := fmt.Sprintf("%s.%s", serverVersionInfo.Major, serverVersionInfo.Minor)
+		utils.AviLog.Infof("Kubernetes cluster apiserver version %s", serverVersion)
+		if lib.CompareVersions(serverVersion, ">", lib.GetK8sMaxSupportedVersion()) ||
+			lib.CompareVersions(serverVersion, "<", lib.GetK8sMinSupportedVersion()) {
+			utils.AviLog.Fatalf("Unsupported kubernetes apiserver version detected. Please check the supportability guide.")
+		}
+	}
+
+	oshiftClient, err := oshiftclient.NewForConfig(cfg)
```
sudswasavi, line 74: this code seems unnecessary. OpenShift will not be a part of the VCF stack.

cmd/infra-main/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		utils.AviLog.Warnf("Error in creating openshift clientset")
+	}
+
+	registeredInformers, err := lib.InformersToRegister(oshiftClient, kubeClient)
```
sudswasavi, line 79: Same here, you can remove the openshift refs throughout the code.

internal/k8s/ako_init.go
```diff
@@ -216,6 +216,74 @@ func (c *AviController) HandleConfigMap(k8sinfo K8sinformers, ctrlCh chan struct
 	return nil
 }
 
+func (c *AviController) AddBootupSecretEventHandler(k8sinfo K8sinformers, stopCh <-chan struct{}, startSyncCh chan struct{}) {
```
sudswasavi, line 219: Can we move this out to a different package? It looks like this will clutter the main AKO code base and this code is only used from the infra package?

internal/k8s/ako_init.go
```diff
@@ -216,6 +216,74 @@ func (c *AviController) HandleConfigMap(k8sinfo K8sinformers, ctrlCh chan struct
...
+
+	go c.informers.SecretInformer.Informer().Run(stopCh)
+	if !cache.WaitForCacheSync(stopCh, c.informers.SecretInformer.Informer().HasSynced) {
+		runtime.HandleError(fmt.Errorf("Timed out waiting for caches to sync"))
+	} else {
+		utils.AviLog.Info("Caches synced for Avi Secret informer")
+	}
+}
+
+func (c *AviController) ValidAviSecret() bool {
```
sudswasavi, line 255: Same for this method. Ideally we should create a controller class within the infra package.

sudswasavi (commented):
In general a bit of code re-org is needed. In order to keep things clean, it would be better to create an infra package and dump all the VCF bootstrap code within it to keep the actual AKO code base clean.

Also consider controller-runtime and see if implementing the controller using that has values over using informers.

cmd/infra-main/main.go
```diff
@@ -0,0 +1,108 @@
...
+	version    = "dev"
+)
+
+func main() {
+	InitializeAKOInfra()
+}
+
+func InitializeAKOInfra() {
+	if !lib.IsVCFCluster() {
+		utils.AviLog.Fatalf("Not running in vcf cluster, shutting down")
```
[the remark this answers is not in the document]
monotosh-avi: Yes. Should't that be fine ?

sudswasavi (dismissed):

monotosh-avi:
build ako

sudswasavi (approved):

chauhanshubham (approved):

PR #625, opened 2021-09-21
Make AKO bootups faster

sudswasavi:
This commit introduces a change that works as follows:

    - During bootup of AKO, we perform a full sync. This results
    into the layer 2 models getting built before activating layer 3.
    - The informer event handlers are then used to enqueue updates to
    the layer 2 queues as well for the same updates which were processed
    as a part of the full sync update.

    This current fix optimizes the second point by caching the resource
    version of the k8s objects during fullsync.

    Then during the regular event updates, it compares the cached resource
    versions in AKO with that of the event updated objects.

    If the resource versions are same, then AKO does not enqueue the same
    key again in layer 2, thus reducing the burden of layer 2 significantly
    during bootup.

    This fix potentially solves a starvation problem in AKO when during a
    heavily loaded system, if AKO is rebooted, and some new routes/ingresses
    are created, they are starved for these already pre-processed updates.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -24,6 +24,9 @@ import (
 	"sync"
 	"time"
 
+	routev1 "github.com/openshift/api/route/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+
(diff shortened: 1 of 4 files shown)


sudswasavi:
build ako

sudswasavi:
build ako

PR #627, opened 2021-09-23
Status update should check for the AKO router name

sudswasavi:
AKO's openshift status populates a unique router name to denote
the vip that was populated via AVI. It's possible that HA Proxy
also runs on the same cluster and updates the Route object as well.

AKO's status update code wasn't checking for the routerName while
update was called on the Route object. This will remove the HAProxy
status as well and end up leading to unnecessary status calls during
AKO bootup.

This commit fixes this issue.

### DIFF
--- a/internal/status/route_status.go
+++ b/internal/status/route_status.go
@@ -133,7 +133,6 @@ func UpdateRouteStatus(options []UpdateOptions, bulk bool) {
 			skipDelete[option.IngSvc] = true
 		}
 	}
-
 	if bulk {
 		for routeNSName, route := range routeMap {
 			if val, ok := skipDelete[routeNSName]; ok && val {
@@ -147,7 +146,6 @@ func UpdateRouteStatus(options []UpdateOptions, bulk bool) {
 			}}, true, lib.SyncStatusKey)
 		}
 	}
-
 	return
 }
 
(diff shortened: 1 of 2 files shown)


internal/status/route_status.go
```diff
@@ -301,7 +299,7 @@ func updateRouteObject(mRoute *routev1.Route, updateOption UpdateOptions, retryN
 
 	// If we find a hostname in the present update, let's first remove it from the existing status.
 	for i := len(mRoute.Status.Ingress) - 1; i >= 0; i-- {
-		if utils.HasElem(hostnames, mRoute.Status.Ingress[i].Host) {
+		if utils.HasElem(hostnames, mRoute.Status.Ingress[i].Host) && mRoute.Status.Ingress[i].RouterName == lib.AKOUser {
```
akshayhavile, line 302: @sudswasavi @monotosh-avi : Will it be good to add "RouterName" check in getRoutes() function before "routes" getting added to routemap? I think this will save few cycles.

monotosh-avi: As discussed, we would need to fetch all routes for now.

monotosh-avi (approved):

hemantavi (approved):

PR #630, opened 2021-09-27
SNI: Http policyset optimization

akshayhavile:


akshayhavile:
build ako

akshayhavile:
Failed due to write /dev/stdout: resource temporarily unavailable

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

anish-avi:
build ako

internal/nodes/avi_model_nodes.go
```diff
@@ -666,12 +676,21 @@ func (o *AviVsNode) ReplaceSniSSLRefInSNINode(newSslNode *AviTLSKeyCertNode, key
...
+		if http.Name == httpPolName {
+			for _, hppMap := range o.HttpPolicyRefs[i].HppMap {
+				if hppMap.Name == hppMapName {
+					hppMapFound = true
+					if http.GetCheckSum() == checksum {
+						return false
+					}
+				}
+				//checksum is not same so need to modify
+				if hppMapFound {
```
monotosh-avi, line 691: I think we dont need the variable hppMapFound, we can add this in an else block after line no: 688

internal/rest/avi_obj_httpps.go
```diff
@@ -37,23 +39,48 @@ func (rest *RestOperations) AviHttpPSBuild(hps_meta *nodes.AviHttpPolicySetNode,
...
+		if hppmap.Path != nil {
+			hppmapWithPath = append(hppmapWithPath, hppmap)
+			httpPresentPaths = append(httpPresentPaths, hppmap.Path[0])
+		} else {
+			hppmapWithoutPath = append(hppmapWithoutPath, hppmap)
+		}
+		httpPresentIng.Insert(hppmap.IngName)
+	}
+	sort.Slice(hppmapWithPath, func(i, j int) bool {
+		return len(hppmapWithPath[i].Path[0]) > len(hppmapWithPath[j].Path[0])
```
monotosh-avi, line 66: Should this be a dictionary sort instead of length ? (e.g. between '/foobar' and '/index' )

internal/nodes/avi_model_nodes.go
```diff
@@ -875,11 +894,21 @@ func (v *AviHttpPolicySetNode) CalculateCheckSum() {
...
 		checksum = checksum + utils.Hash(utils.Stringify(v.HeaderReWrite))
 	}
+
 	if lib.GetGRBACSupport() {
 		checksum += lib.GetMarkersChecksum(v.AviMarkers)
 	}
+
 	v.CloudConfigCksum = checksum
 }
+func (v *AviHostPathPortPoolPG) CalculateCheckSum() {
```
monotosh-avi, line 904: Is this being used ?

internal/lib/lib.go
```diff
@@ -725,10 +732,19 @@ func GetMarkersChecksum(markers utils.AviObjectMarkers) uint32 {
...
 	var combinedString string
 	numMarkerFields := vals.NumField()
+	typeOfVals := vals.Type()
 	markersStr := make([]string, numMarkerFields)
 	for i := 0; i < numMarkerFields; i++ {
 		if vals.Field(i).Interface() != "" {
-			value := vals.Field(i).Interface().(string)
+			field := typeOfVals.Field(i).Name
+			var value string
+			if field == "Path" || field == "IngressName" {
```
sudswasavi, line 741: Unable to understand this code block. But I am assuming something is expected to have pathsArr for either the `Path` or the `IngressName` field.

internal/nodes/avi_model_nodes.go
```diff
@@ -612,21 +619,24 @@ func (o *AviVsNode) ReplaceSniPGInSNINode(newPGNode *AviPoolGroupNode, key strin
...
+func (o *AviVsNode) ReplaceSniHTTPRefInSNINode(httpPGPath AviHostPathPortPoolPG, httpPolName, key string) {
 	for i, http := range o.HttpPolicyRefs {
-		if http.Name == newHttpNode.Name {
-			o.HttpPolicyRefs = append(o.HttpPolicyRefs[:i], o.HttpPolicyRefs[i+1:]...)
-			o.HttpPolicyRefs = append(o.HttpPolicyRefs, newHttpNode)
-			utils.AviLog.Infof("key: %s, msg: replaced sni http in model: %s Pool name: %s", key, o.Name, http.Name)
-			return
+		if http.Name == httpPolName {
+			for j, hppMap := range o.HttpPolicyRefs[i].HppMap {
+				if hppMap.Name == httpPGPath.Name {
```
sudswasavi, line 628: Can you explain what is going on here? It seems we are deleting the element from the slice twice?

internal/lib/lib.go
```diff
@@ -301,12 +301,19 @@ func GetSniPoolName(ingName, namespace, host, path, infrasetting string, args ..
 	return poolName
 }
 
-func GetSniHttpPolName(ingName, namespace, host, path, infrasetting string) string {
+func GetSniHttpPolName(namespace, host, infrasetting string) string {
+
+	if infrasetting != "" {
+		return Encode(NamePrefix+infrasetting+"-"+namespace+"-"+host, HTTPPS)
+	}
+	return Encode(NamePrefix+namespace+"-"+host, HTTPPS)
```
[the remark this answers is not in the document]
akshayhavile: For different paths of one ingress, we are creating single HTTP policy set. Inside that policyset, we will have different hppmap for each path.
So function ' GetSniHttpPolName' gives naming convention for HTTP PS.
function 'GetSniHppMapName' gives naming convention for each HPP map inside httpps.

internal/lib/lib.go
```diff
@@ -301,12 +301,19 @@ func GetSniPoolName(ingName, namespace, host, path, infrasetting string, args ..
...
+}
+func GetSniHppMapName(ingName, namespace, host, path, infrasetting string) string {
```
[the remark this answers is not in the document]
akshayhavile: Addressed above.

internal/lib/lib.go
akshayhavile: Right. With this PR, for HTTP PS markers , ingressName key and path key will have multiple values. To take into consideration all that values, above code logic is used.

internal/nodes/avi_model_nodes.go
akshayhavile: Each HTTP PS will have multiple entries of HPP map, corresponding to path.  We are replacing old hpp map entry with latest one with this code base.

akshayhavile: To explain HTTP PS structure for host foo.com with /foo and /bar in default ns ...
VS will have single HTTP PS with naming convention `<k8clustername>--default-foo.com.`
Inside this HTTPS, we will have HPP map `<K8clustername>'--default-foo.com_foo-ing1` and `<K8clustername>'--default-foo.com_bar-ing1`

akshayhavile: As we have multiple path and ingresses with this HTTP PS, markers will also be an array

internal/nodes/avi_model_nodes.go
akshayhavile: Yes.  it is used in l7_translator as well as in evh_node file to compute checksum associated with hpp map. Sample call is: httpPGPath.CalculateCheckSum()

internal/rest/avi_obj_httpps.go
akshayhavile: Dictionary sort will create problem for overlapping path. For paths like `/lion` and `/lionbear`, dictionary sort will sort it as `/lion, /lionbear`. Due to that, all request to `/lionbear` will be handled by `/lion` backend.

akshayhavile:
build ako

monotosh-avi (approved):

sudswasavi (approved):

PR #631, opened 2021-09-27
Broadcast AKO Pod and Ingress/ServiceLB/Route events

chauhanshubham:
This adds events for AKO Pod, for bootup checkpoints, errors
during user input validation etc. This also covers broadcasting
events related to virtual service creation correponding to
k8s ingress/serviceLB oshift route objects.


sudswasavi:
There's a nil pointer exception going on in the tests. Pls check.

PR #633, opened 2021-09-27
Add autoFQDN for shared VSes

sudswasavi:
For infoblox DNS, we need to retain a FQDN in the shared VS for
it to not release the IP address in the backed.

This commit introduces the feature in the following way:

- Generate the shared VS default FQDN based on the format suggested
by the autoFQDN flag.
- If it's flat then it will generated it of the form:
 vsName + "-" + tenant + "." + <subdomain> else,
 it will be vsName + "." + tenant + "." + <subdomain>

- This FQDN won't be added if autoFQDN is set to disabled

In 1.5.2, we will keep this setting still under the L4Settings
because moving it to a generic AKOSettings now will have impact
on the AKO operator. But in 1.6.1, we would want to move this
autoFQDN flag under AKOSettings


sudswasavi:
Closing this PR because we want to just make this to AKO 1.6.1

PR #634, opened 2021-09-28
Add autoFQDN for shared VSes

sudswasavi:
For infoblox DNS, we need to retain a FQDN in the shared VS for
it to not release the IP address in the backed.

This commit introduces the feature in the following way:

- Generate the shared VS default FQDN based on the format suggested
by the autoFQDN flag.
- If it's flat then it will generated it of the form:
 vsName + "-" + tenant + "." + <subdomain> else,
 it will be vsName + "." + tenant + "." + <subdomain>

- This FQDN won't be added if autoFQDN is set to disabled

In 1.5.2, we will keep this setting still under the L4Settings
because moving it to a generic AKOSettings now will have impact
on the AKO operator. But in 1.6.1, we would want to move this
autoFQDN flag under AKOSettings


sudswasavi:
build ako

chauhanshubham:
Is the requirement for EVH too? cc: @akshayhavile

monotosh-avi (approved):

internal/nodes/avi_model_l7_translator.go
```diff
@@ -318,6 +318,35 @@ func (o *AviObjectGraph) ConstructAviL7VsNode(vsName string, key string) *AviVsN
...
+		} else if lib.GetL4FqdnFormat() == lib.AutoFQDNFlat {
+			// Generate the FQDN based on the logic: <svc_name>-<namespace>.<sub-domain>
+			fqdn = vsName + "-" + lib.GetTenant() + "." + subdomain
+		}
+		utils.AviLog.Infof("key: %s, msg: Configured the shared VS with default fqdn as: %s", fqdn)
+		fqdns = append(fqdns, fqdn)
+	}
 
 	subDomains := GetDefaultSubDomain()
 	if subDomains != nil {
```
akshayhavile, line 352: It seems we have overlapping code here(line number 352-359). so fqdns might be duplicate. Can you please check this?

akshayhavile (commented):
> Is the requirement for EVH too? cc: @akshayhavile

Yes this will be required for EVH also. We have function ConstructAviL7SharedVsNodeForEvh() that will require changes.

internal/nodes/avi_model_l7_translator.go
sudswasavi: It appears that the branch was old and the snapshot you are pointing to you is from the older branch. This code does not exist in the current `master` branch.

Either way, I will update the branch and repost.

sudswasavi: I just rebased the branch, you won't see the code now :)

sudswasavi:
build ako

akshayhavile (approved):
LGTM. As you said, in next PR, we need to move flag "defaultDomain" and "autoFQDN" from L4Settings section to AKOSetting section.

sudswasavi:
> why do we need a distinction between `flat` and `default` for Shared VS fqdns? Are there any issues in always adding the fqdn?

The reason behind doing the `flat` and `default` schemes were DNS related issues with infoblox. The same would apply here as well.

chauhanshubham (approved):

PR #636, opened 2021-09-28
Publish the CRD if the status changes during bootup

sudswasavi:
Consider the scenario:

- AKO boots up. During  bootup, a HostRule CRD is created.
- The HostRule CRD does not perform a validation logic during full
sync.
- We cache the resource version during fullsync.
- When the event handlers are switched on, we find the same resource
version, so we don't do the validation.
- The HostRule is never applied.

This fix will force a re-publish of the HostRule/HTTPRule/AviInfra
CRD if they are found to have a status change during the validation.

### DIFF
--- a/internal/k8s/crdcontroller.go
+++ b/internal/k8s/crdcontroller.go
@@ -113,11 +113,14 @@ func (c *AviController) SetupAKOCRDEventHandlers(numWorkers uint32) {
 				hostrule := obj.(*akov1alpha1.HostRule)
 				namespace, _, _ := cache.SplitMetaNamespaceKey(utils.ObjKey(hostrule))
 				key := lib.HostRule + "/" + utils.ObjKey(hostrule)
+				statusBefore := hostrule.Status.Status
 				if err := validateHostRuleObj(key, hostrule); err != nil {
 					utils.AviLog.Warnf("Error retrieved during validation of HostRule: %v", err)
 				}
+				// It's a reference, so pointer should be able to access the updated status.
+				statusAfter := hostrule.Status.Status
 				ok, resVer := objects.SharedResourceVerInstanceLister().Get(key)
-				if ok && resVer.(string) == hostrule.ResourceVersion {
+				if ok && resVer.(string) == hostrule.ResourceVersion && statusBefore == statusAfter {
 					utils.AviLog.Debugf("key: %s, msg: Same resource version returning", key)
 					return
 				}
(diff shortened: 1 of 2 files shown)


internal/k8s/crdcontroller.go
```diff
@@ -250,11 +256,14 @@ func (c *AviController) SetupAKOCRDEventHandlers(numWorkers uint32) {
...
 				key := lib.AviInfraSetting + "/" + utils.ObjKey(aviinfra)
+				statusBefore := aviinfra.Status.Status
 				if err := validateAviInfraSetting(key, aviinfra); err != nil {
 					utils.AviLog.Warnf("Error retrieved during validation of AviInfraSetting: %v", err)
 				}
+				// It's a reference, so pointer should be able to access the updated status.
+				statusAfter := aviinfra.Status.Status
 				ok, resVer := objects.SharedResourceVerInstanceLister().Get(key)
-				if ok && resVer.(string) == aviinfra.ResourceVersion {
+				if ok && resVer.(string) == aviinfra.ResourceVersion && statusAfter == statusBefore {
```
akshayhavile, line 266: Should we do similar changes to Istio CRDs?

internal/nodes/avi_model_l7_translator.go
```diff
@@ -218,7 +215,7 @@ func (o *AviObjectGraph) BuildTlsCertNode(svcLister *objects.SvcLister, tlsNode
 			return false
 		}
 	} else {
-		secretObj, err := mClient.CoreV1().Secrets(secretNS).Get(context.TODO(), secretName, metav1.GetOptions{})
+		secretObj, err := utils.GetInformers().SecretInformer.Lister().Secrets(namespace).Get(secretName)
```
akshayhavile, line 218: Should we use "secretNS" instead of "namespace" ?

sudswasavi: Done.

internal/k8s/crdcontroller.go
sudswasavi: Istio CRDs may not have this problem. We can tackle it in `master` during Istio implementation. Istio's code base in layer 2 won't look for an `Accepted` or `Rejected` decision to apply the CRD, hence I don't think the problem should arise.

sudswasavi:
build ako

sudswasavi:
build ako

monotosh-avi (approved):

akshayhavile (approved):

PR #637, opened 2021-09-29
Added a change to convert the secret from clientset to informer

sudswasavi:
A simple one liner change to use informers instead of clients for secret data.


sudswasavi:
build ako

monotosh-avi (approved):

PR #638, opened 2021-09-29
Correct problem with status checks on CRD

sudswasavi:
This commit addresses 2 problems:

- The CRD status verification in layer 1 for optimization is now
   removed.

- The secret informer is now switched on for all namespaces.
This helps with bootup speed because we don't use the clientsets
anymore. As a result of this change, the pod resource sizing
of AKO is altered in the values.yaml

Fixes: AV-125902

### DIFF
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -77,11 +77,11 @@ nodePortSelector: # Only applicable if serviceType is NodePort
 
 resources:
   limits:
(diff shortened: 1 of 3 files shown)


monotosh-avi (approved):

sudswasavi:
build ako

sudswasavi:
build ako

PR #639, opened 2021-09-29
Fix status update issue during WCP full sync

chauhanshubham:
The issue was introduced as part of
https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/commit/46e8d1da1496b4bf66220158150d704574a58aee
The changes got left out for advl4 usecases.

### DIFF
--- a/internal/rest/k8s_utils.go
+++ b/internal/rest/k8s_utils.go
@@ -73,15 +73,6 @@ func (rest *RestOperations) SyncObjectStatuses() {
 						VirtualServiceUUID: vsCacheObj.Uuid,
 					})
 			}
-		} else if len(vsSvcMetadataObj.NamespaceServiceName) > 0 {
-			// serviceLB
-			allServiceLBUpdateOptions = append(allServiceLBUpdateOptions,
-				status.UpdateOptions{
-					Vip:                vsCacheObj.Vip,
-					ServiceMetadata:    vsSvcMetadataObj,
-					Key:                lib.SyncStatusKey,
-					VirtualServiceUUID: vsCacheObj.Uuid,
-				})
 		} else {
 			// insecure VSes handler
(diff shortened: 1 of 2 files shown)


chauhanshubham:
build ako

internal/rest/k8s_utils.go
```diff
@@ -104,9 +95,28 @@ func (rest *RestOperations) SyncObjectStatuses() {
 							Key:                lib.SyncStatusKey,
 							VirtualServiceUUID: vsCacheObj.Uuid,
 						})
+				} else if len(poolCacheObj.ServiceMetadataObj.NamespaceServiceName) > 0 {
```
monotosh-avi, line 98: We are updating allServiceLBUpdateOptions again in line no 110, would that add a duplicate ?

chauhanshubham: that is corresponding to vs cache svc metadata. This one is when corresponding to pool metadata. Had made an enhancement in pool svc metadata sometime back for `servicesAPI: true` so this takes care of that. The other one is for dedicated L4 VSes.

internal/status/advl4_status.go
```diff
@@ -90,8 +81,6 @@ func UpdateGatewayStatusAddress(options []UpdateOptions, bulk bool) {
 			}, lib.SyncStatusKey)
 		}
 	}
-
-	UpdateL4LBStatus(updateServiceOptions, bulk)
```
sudswasavi, line 94: are we going to not update the service of type loadbalancers now?

monotosh-avi (approved):

sudswasavi (approved):

sudswasavi:
There are currently no UT jobs for this PR. I am purely merging this based on local testing.

PR #643, opened 2021-10-01
Revert to using locks for all objects

sudswasavi:
This commit reverts all the changes which were made earlier w.r.t
removing locks from the objects that didn't require them.

If we keep the two models, it is possibly error prone for the future
hence removing those using this commit to keep the store consistent
to using locks always.


sudswasavi:
build ako

monotosh-avi (approved):

PR #651, opened 2021-10-07
Use NS bound informer for Secrets in Openshift

monotosh-avi:
There are many secrets installed by default in Openshift cluster which have to be handled if NS bound informer is not used.
Fixes AV-126820


monotosh-avi:
build ako

PR #654, opened 2021-10-08
Add logical segment and tier1 router in nsxt cloud from networkinfo

monotosh-avi:
In VCF environment, Read network info CRD to find out all LR to LS mapping,
and update them in the data network of NSXT cloud.

During boot up, do a full sync to correct any mismatch in LS-LR mapping.
After that, Add/Delete segments based on events for Networkinfo CRD.

### DIFF
--- a/ako-infra/avirest/handle_netinfo.go
+++ b/ako-infra/avirest/handle_netinfo.go
@@ -44,6 +44,15 @@ func SyncLSLR() {
 		utils.AviLog.Warnf("Failed to get Cloud data from cache")
 		return
 	}
+
+	if len(cloudModel.NsxtConfiguration.DataNetworkConfig.VlanSegments) != 0 {
+		utils.AviLog.Infof("NSX-T cloud is using Vlan Segments, LS-LR mapping won't be updated")
+		return
+	}
+	if cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual == nil {
+		utils.AviLog.Warnf("Tier1SegmentConfig is nil in NSX-T cloud, LS-LR mapping won't be updated")
+		return
+	}
 	matched := matchSegmentInCloud(cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual.Tier1Lrs, lslrmap)
 	if !matched {
 		cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual.Tier1Lrs = constructLsLrInCloud(lslrmap)
@@ -57,8 +66,8 @@ func SyncLSLR() {
 }
 
 func AddSegment(obj interface{}) {
-	objKey := "Netinfo" + utils.ObjKey(obj)
-	utils.AviLog.Debugf("key: %s, Network Info DELETE Event", objKey)
+	objKey := "Netinfo" + "/" + utils.ObjKey(obj)
+	utils.AviLog.Debugf("key: %s, Network Info ADD Event", objKey)
 	crd := obj.(*unstructured.Unstructured)
 
 	specJSON, found, err := unstructured.NestedStringMap(crd.UnstructuredContent(), "topology")
@@ -72,7 +81,7 @@ func AddSegment(obj interface{}) {
 	client := avicache.SharedAVIClients().AviClient[0]
 	found, cloudModel := getAviCloudFromCache(client, utils.CloudName)
 	if !found {
-		utils.AviLog.Warnf("key: %s, Failed to get Cloud data from cache", objKey)
+		utils.AviLog.Fatalf("key: %s, Failed to get Cloud data from cache", objKey)
 		return
 	}
 	updateRequired, lslrList := addSegmentInCloud(cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual.Tier1Lrs, lr, ls)
@@ -119,12 +128,15 @@ func DeleteSegment(obj interface{}) {
 }
 
 func addSegmentInCloud(lslrList []*models.Tier1LogicalRouterInfo, lr, ls string) (bool, []*models.Tier1LogicalRouterInfo) {
-	for i := range lslrList {
+	listCopy := make([]*models.Tier1LogicalRouterInfo, len(lslrList))
+	copy(listCopy, lslrList)
+	for i := range listCopy {
 		if *lslrList[i].SegmentID == ls {
 			if *lslrList[i].Tier1LrID == lr {
 				return false, lslrList
 			}
 			lslrList = append(lslrList[:i], lslrList[i+1:]...)
+			break
 		}
 	}
 	lrls := models.Tier1LogicalRouterInfo{
@@ -230,9 +242,6 @@ func executeRestOp(key string, client *clients.AviClient, restOp *utils.RestOp, 
 }
 
 func checkAndRetry(key string, err error) bool {
-	if err == nil {
-		return false
-	}
 	if webSyncErr, ok := err.(*utils.WebSyncError); ok {
 		if aviError, ok := webSyncErr.GetWebAPIError().(session.AviError); ok {
 			switch aviError.HttpStatusCode {
--- a/internal/lib/dynamic_client.go
+++ b/internal/lib/dynamic_client.go
@@ -216,8 +216,16 @@ func GetNetinfoCRData() map[string]string {
 	for _, obj := range crList.Items {
 		spec, found, err := unstructured.NestedStringMap(obj.UnstructuredContent(), "topology")
 		if found && err == nil {
-			lr := spec["gatewayPath"]
-			ls := spec["aviSegmentPath"]
+			lr, found := spec["gatewayPath"]
+			if !found {
+				utils.AviLog.Warnf("gatewayPath not found in NetworkInfo object")
+				continue
+			}
+			ls, found := spec["aviSegmentPath"]
+			if !found {
+				utils.AviLog.Warnf("aviSegmentPath not found in NetworkInfo object")
+				continue
+			}
 			lslrMap[ls] = lr
 		} else {
 			utils.AviLog.Warnf("NetworkInfo topology not found: %v", err)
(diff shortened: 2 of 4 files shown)


ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,250 @@
...
+		restOp := utils.RestOp{ObjName: utils.CloudName, Path: path, Method: utils.RestPut, Obj: &cloudModel,
+			Tenant: "admin", Model: "cloud", Version: utils.CtrlVersion}
+		executeRestOp("fullsync", client, &restOp)
+	} else {
+		utils.AviLog.Infof("LS LR update not required in cloud: %s", utils.CloudName)
+	}
+}
+
+func AddSegment(obj interface{}) {
+	objKey := "Netinfo" + utils.ObjKey(obj)
```
sudswasavi, line 60: are you missing a `/`

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,250 @@
...
+	lslrmap := lib.GetNetinfoCRData()
+	utils.AviLog.Infof("Got data LS LR Map: %v, from NetworkInfo CR", lslrmap)
+
+	client := avicache.SharedAVIClients().AviClient[0]
+	found, cloudModel := getAviCloudFromCache(client, utils.CloudName)
+	if !found {
+		utils.AviLog.Warnf("Failed to get Cloud data from cache")
+		return
+	}
+	matched := matchSegmentInCloud(cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual.Tier1Lrs, lslrmap)
```
sudswasavi, line 47: This may through a `nil` pointer if the cloud is configured with `VLAN` segment I think. It is good to check if this variable cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual.Tier1Lrs is true in all cases. 

We could optionally just have a check where unless the cloud is of type `Overlay` we don't process anything.

internal/lib/dynamic_client.go
```diff
@@ -202,6 +203,30 @@ func GetBootstrapCRData() (string, string, string) {
...
+	crList, err := crdClient.List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		utils.AviLog.Errorf("Error getting Networkinfo CR %v", err)
+		return lslrMap
+	}
+
+	for _, obj := range crList.Items {
+		spec, found, err := unstructured.NestedStringMap(obj.UnstructuredContent(), "topology")
+		if found && err == nil {
+			lr := spec["gatewayPath"]
```
sudswasavi, line 219: Is it possible that the keys don't exist for a particular instance of this CRD? Let's say an older version?

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,250 @@
...
+			Tenant: "admin", Model: "cloud", Version: utils.CtrlVersion}
+		executeRestOp("fullsync", client, &restOp)
+	} else {
+		utils.AviLog.Infof("LS LR update not required in cloud: %s", utils.CloudName)
+	}
+}
+
+func AddSegment(obj interface{}) {
+	objKey := "Netinfo" + utils.ObjKey(obj)
+	utils.AviLog.Debugf("key: %s, Network Info DELETE Event", objKey)
```
sudswasavi, line 61: Why is it a `DELETE` event?

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,250 @@
...
+		utils.AviLog.Warnf("NetworkInfo topology not found: %+v", err)
+		return
+	}
+	lr := specJSON["gatewayPath"]
+	ls := specJSON["aviSegmentPath"]
+	utils.AviLog.Infof("key: %s, Adding LR %s, LS %s from networkinfo CR", objKey, lr, ls)
+	client := avicache.SharedAVIClients().AviClient[0]
+	found, cloudModel := getAviCloudFromCache(client, utils.CloudName)
+	if !found {
+		utils.AviLog.Warnf("key: %s, Failed to get Cloud data from cache", objKey)
```
sudswasavi, line 75: is it just okay to return of the cloud object is not found or should there be some action? It seems like:
 - either we can fetch the cloud cache.
 - cloud itself does not exist in that case it's a `Fatalf`

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,250 @@
...
+
+	specJSON, found, err := unstructured.NestedStringMap(crd.UnstructuredContent(), "topology")
+	if err != nil || !found {
+		utils.AviLog.Warnf("key: %s, NetworkInfo topology not found: %v", objKey, err)
+		return
+	}
+	lr := specJSON["gatewayPath"]
+	ls := specJSON["aviSegmentPath"]
+	utils.AviLog.Infof("key: %s, Network Info CR deleted, removing LR %s, LS %s from cloud", objKey, lr, ls)
+	client := avicache.SharedAVIClients().AviClient[0]
```
sudswasavi, line 103: It seems when the WCP cluster is being teared down, we will see lot of these `DeleteSegment` calls and because we are using the same AviClient, all calls would be blocked into 1 session. Can we do something to make these parallel, without necessary using a queue?

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+}
+
+func addSegmentInCloud(lslrList []*models.Tier1LogicalRouterInfo, lr, ls string) (bool, []*models.Tier1LogicalRouterInfo) {
+	for i := range lslrList {
+		if *lslrList[i].SegmentID == ls {
+			if *lslrList[i].Tier1LrID == lr {
+				return false, lslrList
+			}
+			lslrList = append(lslrList[:i], lslrList[i+1:]...)
```
sudswasavi, line 127: if we don't hit the `if` block then we will remove the entry from the `lslrlist`? Why would we do that?

sudswasavi (commented):
Do we need some UTs for these?

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,261 @@
...
+	utils.AviLog.Infof("Got data LS LR Map: %v, from NetworkInfo CR", lslrmap)
+
+	client := avicache.SharedAVIClients().AviClient[0]
+	found, cloudModel := getAviCloudFromCache(client, utils.CloudName)
+	if !found {
+		utils.AviLog.Warnf("Failed to get Cloud data from cache")
+		return
+	}
+
+	if len(cloudModel.NsxtConfiguration.DataNetworkConfig.VlanSegments) != 0 {
```
akshayhavile, line 48: @monotosh-avi @sudswasavi : I was going through model definition of cloud. Few fields are supported in Basic, Enterprise and Few in Enterprise and Essential. For which license version, are we claiming support for nsx-t cloud.

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,261 @@
...
+		utils.AviLog.Infof("NSX-T cloud is using Vlan Segments, LS-LR mapping won't be updated")
+		return
+	}
+	if cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual == nil {
```
akshayhavile, line 52: I think `cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig` will be empty in `VLAN segment`. Will then referring `cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual` cause exception?

ako-infra/avirest/handle_netinfo.go
akshayhavile: As we take out entry from lslrList should we "break" for loop? This will save few cycles. I assume there will not be multiple entries for same SegID.

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,261 @@
...
+		if checkAndRetry(key, err) {
+			executeRestOp(key, client, restOp, retry+1)
+		}
+	}
+	AviCloudCachePopulate(client, utils.CloudName)
+	utils.AviLog.Infof("key: %s, Successfully executed rest operation to sync object in cloud: %v", key, *restOp)
+}
+
+func checkAndRetry(key string, err error) bool {
+	if err == nil {
```
akshayhavile, line 244: checkAndRetry will be call only when err != nil. so i feel this "if" is redundant.

ako-infra/avirest/handle_netinfo.go
```diff
@@ -0,0 +1,261 @@
...
+		retry = retryNum[0]
+		if retry >= 3 {
+			utils.AviLog.Errorf("key %s, rest request execution aborted after reqtrying 3 times", key)
+			return
+		}
+	}
+	err := rest.AviRestOperateWrapper(client, []*utils.RestOp{restOp})
+	if err != nil {
+		utils.AviLog.Infof("key %s, Got error in executing rest request: %v", key, err)
+		if checkAndRetry(key, err) {
```
akshayhavile, line 235: if `checkAndRetry` failed due to `Invalid Credentials(401)` and returns `false`, we will not retry. We will populateCache and Print rest operation successful message even though rest op is failed.  I think this flow seems to be wrong.

ako-infra/avirest/handle_netinfo.go
monotosh-avi: For VLAN, we would exit in line no: 50 above.

ako-infra/avirest/handle_netinfo.go
monotosh-avi: Invalid credential is a genuine error which won't be fixed in retry, 401 error due to session timeout would be handled in retry. This is similar to - https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/internal/rest/dequeue_nodes.go#L193

ako-infra/avirest/handle_netinfo.go
monotosh-avi: This was a safety check, but we don't it for now, removed.

ako-infra/avirest/handle_netinfo.go
monotosh-avi: yeah, we can break. added.

sudswasavi (approved):

sudswasavi:
The build job is stuck. But this fix is required to proceed and hence force merging.

PR #655, opened 2021-10-08
Support for multi-port Services with port number for ServiceType NodePort and NodePortLocal

swathinsankaran:
This PR adds support for multi-port services with the port number when the ServiceType is specified as NodePort/NodePortLocal.

For example, consider the following service and ingress:

multi-port service.yaml
```
apiVersion: v1
kind: Service
metadata:
  name: ew-app
  namespace: ingress-class-ns
spec:
  ports:
  - port: 8080
    protocol: TCP
    targetPort: 8080
    name: port1
    nodePort: 32008
  - port: 8081
    protocol: TCP
    targetPort: 8080
    name: port2
    nodePort: 32009
  selector:
    app: ew-app
  sessionAffinity: None
  type: NodePort
```

ingress.yaml
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test-ingress
  namespace: ingress-class-ns
spec:
  ingressClassName: avi-lb
  rules:
  - host: test-ingclass.avi.internal
    http:
      paths:
      - path: /bar
        backend:
          serviceName: ew-app
          servicePort: 8080
      - path: /
        backend:
          serviceName: ew-app
          servicePort: port2
```

The backend service port, 8080, is an integer value in the above ingress definition, and the server list getting sent as empty to the controller when the backend service port was given as a number. This PR fixes the issue.

### DIFF
--- a/tests/npltests/npl_pod_test.go
+++ b/tests/npltests/npl_pod_test.go
@@ -1208,9 +1208,9 @@ func TestNPLSvcNodePort(t *testing.T) {
 	}
 }
 
-// TestIngressAddPodWithMultiport creates a Pod with multiple nodeportlocal.antrea.io annotations, Service with multiport and
+// TestIngressAddPodWithMultiportSvc creates a Pod with multiple nodeportlocal.antrea.io annotations, Service with multiport and
 // an Ingress which uses that Service. Port number is mentioned instead of port name as backend servicePort.
-func TestIngressAddPodWithMultiport(t *testing.T) {
+func TestIngressAddPodWithMultiportSvc(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
 
 	SetUpTestForIngress(t, defaultL7Model)


internal/nodes/avi_model_l4_translator.go
```diff
@@ -197,7 +197,7 @@ func PopulateServersForNPL(poolNode *AviPoolNode, ns string, serviceName string,
 
 	targetPorts := make(map[int]bool)
 	for _, port := range svcObj.Spec.Ports {
-		if port.Name != poolNode.PortName && len(svcObj.Spec.Ports) != 1 {
```
sudswasavi, line 200: will there be a case where the svcObj.Spec.Ports 's length is 0?

sudswasavi (commented):
Looks like a great PR. One minor nit, otherwise LGTM.

internal/nodes/validator.go
```diff
@@ -361,14 +365,31 @@ func (v *Validator) findTargetPort(serviceName, ns string, servicePort int32, ke
 	}
```
sudswasavi, line 365: Is line 358's check of `NodePort` still needed?

swathinsankaran: No harm in keep/removing the condition check as the `TargetPort` is not getting used for `NodePort`

internal/nodes/validator.go
```diff
@@ -361,14 +365,31 @@ func (v *Validator) findTargetPort(serviceName, ns string, servicePort int32, ke
...
 	for _, port := range svcObj.Spec.Ports {
 		// Iterate the ports and find the match for targetPort
-		if servicePort == port.Port {
-			utils.AviLog.Infof("key: %s, msg: Found targetPort %v for Port: %v", key, port.TargetPort.IntVal, servicePort)
+		if serviceBackendPort.Number == port.Port ||
+			serviceBackendPort.Name == port.Name {
+			utils.AviLog.Infof("key: %s, msg: Found targetPort %v for Port: %v", key, port.TargetPort.IntVal, serviceBackendPort)
```
akshayhavile, line 370: @swathinsankaran : Are we intended print whole serviceBackendPort struct or just serviceport  number

swathinsankaran: The user can give either port number or port name in the ingress config. In both cases, we will see at least one value in the log.

PR #657, opened 2021-10-11
cherry-pick: Fix status update issue during WCP full sync

chauhanshubham:
The issue was introduced as part of
https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/commit/46e8d1da1496b4bf66220158150d704574a58aee
The changes were left out for advl4 usecases.

### DIFF
--- a/internal/rest/k8s_utils.go
+++ b/internal/rest/k8s_utils.go
@@ -76,21 +76,6 @@ func (rest *RestOperations) SyncObjectStatuses() {
 						VirtualServiceUUID: vsCacheObj.Uuid,
 					})
 			}
-		} else if len(vsSvcMetadataObj.NamespaceServiceName) > 0 {
-			// serviceLB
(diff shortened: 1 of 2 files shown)


internal/rest/k8s_utils.go
```diff
@@ -76,21 +76,6 @@ func (rest *RestOperations) SyncObjectStatuses() {
 						VirtualServiceUUID: vsCacheObj.Uuid,
 					})
 			}
-		} else if len(vsSvcMetadataObj.NamespaceServiceName) > 0 {
```
akshayhavile, line 79: @chauhanshubham : Sorry, I could have missed this during original PR. I am not getting purpose of removing LBUpdateOption block from here and adding it two places. 
 I think for each L4 service, vskey will be always different and will be part of vsKeys fetched.  so i think we should not treat it with insecure vs way.
 Also all changes doesn't seems to be specific to wcp(gw).

PR #662, opened 2021-10-12
Allow adding custom Pod CIDR to Node mapping via annotation

chauhanshubham:
This commit allows users to add a comma separated list of Pod CIDRs
that are being used in the k8s Node, as an annotation. The annotation
key to be used for this is `ako.vmware.com/pod-cidrs` in the Node.
This is required in scenarios where AKO cannot figure out the CIDR to
Node mapping from any existing k8s resources, for instance, when
calico is installed on the cluster using `etcd` as the datastore.
AV-126750


chauhanshubham:
build ako

sudswasavi (approved):

internal/rest/avi_obj_vrf.go
```diff
@@ -115,7 +115,7 @@ func (rest *RestOperations) getVrfCacheObj(vrfName string) *avicache.AviVrfCache
 
 func (rest *RestOperations) AviVrfCacheAdd(restOp *utils.RestOp, vrfKey avicache.NamespaceName, key string) error {
 	if (restOp.Err != nil) || (restOp.Response == nil) {
-		utils.AviLog.Warnf("key: %s, rest_op has err or no response for vrfcontext, err: %s, response: %s", key, restOp.Err, restOp.Response)
+		utils.AviLog.Warnf("key: %s, rest_op has err or no response for vrfcontext, err: %s, response: %v", key, restOp.Err, utils.Stringify(restOp.Response))
```
monotosh-avi, line 118: Nit: Do we need %v after Stringify ?

monotosh-avi (approved):

PR #667, opened 2021-10-18
Auto-detect cloud and create SE group

sudswasavi:
The following commit introduces the following changes:

- Derives the NSX-T cloud name from the controller based on the
transport zone info.
- Caches the cloud in an AKO in memory story.
- Verifies if the cloud has an SE group template populated or not.
- If SE group template exists, then uses that to create an SE group.
- If the SE group template does not exist then it uses the `Default-Group`
settings

NOTE: This code requires a resolution on: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/654
before merge

Integration tests cannot be performed till then.

### DIFF
--- a/ako-infra/ingestion/avi_infra_controller.go
+++ b/ako-infra/ingestion/avi_infra_controller.go
@@ -118,66 +118,44 @@ func (a *AviControllerInfra) DeriveCloudNameAndSEGroupTmpl(tz string) (error, st
 
 func (a *AviControllerInfra) SetupSEGroup(tz string) bool {
 	// This method checks if the cloud in Avi has a SE Group template configured or not. If has the SEG template then it returns true, else false
-	err, _, segUuid := a.DeriveCloudNameAndSEGroupTmpl(tz)
-	if err == nil {
-		if segUuid == "" {
-			// The cloud does not have a SEG template set, use `Default-Group`
-			uri := "/api/serviceenginegroup/" + "?name=Default-Group"
-			result, err := lib.AviGetCollectionRaw(a.AviRestClients.AviClient[0], uri)
-			if err != nil {
-				utils.AviLog.Errorf("Get uri %v returned err %v", uri, err)
-				return false
-			}
-			// Construct an SE group based on parameters in the `Default-Group`
-			elems := make([]json.RawMessage, result.Count)
-			err = json.Unmarshal(result.Results, &elems)
-			if err != nil {
-				utils.AviLog.Errorf("Failed to unmarshal data, err: %v", err)
-				return false
-			}
+	err, cloudName, segUuid := a.DeriveCloudNameAndSEGroupTmpl(tz)
+	if err != nil {
+		return false
+	}
+	var uri string
+	if segUuid == "" {
+		// The cloud does not have a SEG template set, use `Default-Group`
+		uri = "/api/serviceenginegroup/?include_name&cloud_ref.name=" + cloudName + "&name=Default-Group"
+	} else {
+		// se group template exists, use the same to fetch the SE group details and use it to create the new SE group
+		// The cloud does not have a SEG template set, use `Default-Group`
+		uri = "/api/serviceenginegroup/" + segUuid
+	}
+	result, err := lib.AviGetCollectionRaw(a.AviRestClients.AviClient[0], uri)
+	if err != nil {
+		utils.AviLog.Errorf("Get uri %v returned err %v", uri, err)
+		return false
+	}
+	// Construct an SE group based on parameters in the `Default-Group`
+	elems := make([]json.RawMessage, result.Count)
+	err = json.Unmarshal(result.Results, &elems)
+	if err != nil {
+		utils.AviLog.Errorf("Failed to unmarshal data, err: %v", err)
+		return false
+	}
 
-			for _, elem := range elems {
-				seg := models.ServiceEngineGroup{}
-				if err := json.Unmarshal(elem, &seg); err != nil {
-					utils.AviLog.Warnf("Failed to unmarshal data, err: %v", err)
-					continue
-				}
+	for _, elem := range elems {
+		seg := models.ServiceEngineGroup{}
+		if err := json.Unmarshal(elem, &seg); err != nil {
+			utils.AviLog.Warnf("Failed to unmarshal data, err: %v", err)
+			continue
+		}
 
-				if !ConfigureSeGroup(a.AviRestClients.AviClient[0], &seg) {
-					return false
-				}
-			}
-
-		} else {
-			// se group template exists, use the same to fetch the SE group details and use it to create the new SE group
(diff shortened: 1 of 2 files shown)


ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -73,6 +77,137 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+			return nil, *cloud.Name, ""
+		}
+	}
+	return errors.New("Cloud not found"), "", ""
+}
+
+func (a *AviControllerInfra) SetupSEGroup(tz string) bool {
+	// This method checks if the cloud in Avi has a SE Group template configured or not. If has the SEG template then it returns true, else false
+	err, _, segUuid := a.DeriveCloudNameAndSEGroupTmpl(tz)
+	if err == nil {
```
monotosh-avi, line 122: nit: we can return if error is not nil and avoid the else

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -73,6 +77,137 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+		if segUuid == "" {
+			// The cloud does not have a SEG template set, use `Default-Group`
+			uri := "/api/serviceenginegroup/" + "?name=Default-Group"
```
monotosh-avi, line 125: Should we filter for cloud name here ? there can be different segroups belonging to different Clouds.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -73,6 +77,137 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+		}
+	}
+	return errors.New("Cloud not found"), "", ""
+}
+
+func (a *AviControllerInfra) SetupSEGroup(tz string) bool {
+	// This method checks if the cloud in Avi has a SE Group template configured or not. If has the SEG template then it returns true, else false
+	err, _, segUuid := a.DeriveCloudNameAndSEGroupTmpl(tz)
+	if err == nil {
+		if segUuid == "" {
```
akshayhavile, line 123: I think we can optimise this code as (only change uri on segUuid .. rest of code of fetching seg and configuring it is common):
if segUuid == "" {
  uri = "/api/serviceenginegroup/?name=default-group"
} else {
  uri = "/api/serviceenginegroup/" + segUuid"
}
result, err := lib.AviGetCollectionRaw(a.AviRestClients.AviClient[0], uri)
.
.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -73,6 +77,137 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+					}
+				}
+			}
+			return nil, *cloud.Name, ""
+		}
+	}
+	return errors.New("Cloud not found"), "", ""
+}
+
+func (a *AviControllerInfra) SetupSEGroup(tz string) bool {
```
akshayhavile, line 119: How "SetupSEGroup" is getting called? Couldn't figure out logical flow.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -73,6 +77,137 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+	if lib.GetTenantsPerCluster() && lib.IsCloudInAdminTenant {
+		SetAdminTenant := session.SetTenant(lib.GetAdminTenant())
+		SetTenant := session.SetTenant(lib.GetTenant())
+		SetAdminTenant(client.AviSession)
+		defer SetTenant(client.AviSession)
+	}
+
+	err := lib.AviPost(client, uri, seGroup, response)
+	if err != nil {
+		utils.AviLog.Warnf("Error during POST call to create the SE group :%v. Expected Labels: %v", *seGroup.Name, err.Error(), utils.Stringify(lib.GetLabels()))
```
akshayhavile, line 203: one %v is missing. We are trying to print 3 parameters.

internal/lib/avi_api.go
```diff
@@ -114,6 +114,28 @@ func AviPut(client *clients.AviClient, uri string, payload interface{}, response
...
+		retry = retryNum[0]
+		if retry >= 3 {
+			err := errors.New("msg: AviPut retried 3 times, aborting")
+			return err
+		}
+	}
+
+	err := client.AviSession.Post(uri, payload, &response)
+	if err != nil {
+		utils.AviLog.Warnf("msg: Unable to execute Put on uri %s %v", uri, err)
```
akshayhavile, line 129: I think message needs to be changed as we are doing "post" operation.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -73,6 +77,137 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+	response := models.ServiceEngineGroupAPIResponse{}
+	// If tenants per cluster is enabled then the X-Avi-Tenant needs to be set to admin for vrfcontext and segroup updates
+	if lib.GetTenantsPerCluster() && lib.IsCloudInAdminTenant {
+		SetAdminTenant := session.SetTenant(lib.GetAdminTenant())
+		SetTenant := session.SetTenant(lib.GetTenant())
+		SetAdminTenant(client.AviSession)
+		defer SetTenant(client.AviSession)
+	}
+
+	err := lib.AviPost(client, uri, seGroup, response)
```
akshayhavile, line 201: One basic doubt: We are either using "default SE group" or new SE Group from template. Do we need to always do post operation (as in "Default SE group" case , as group already exist, post might fail)

ako-infra/ingestion/avi_infra_controller.go
sudswasavi: done.

ako-infra/ingestion/avi_infra_controller.go
sudswasavi: Great catch done.

ako-infra/ingestion/avi_infra_controller.go
sudswasavi: It will be called post the #654

ako-infra/ingestion/avi_infra_controller.go
sudswasavi: Thanks changed.

ako-infra/ingestion/avi_infra_controller.go
sudswasavi: Good question. There's a never an update expected on the SEG, so I missed a method on bootup on this PR that will check if the SE group exists or not with the right marker. If it does, then it will no-op. I wanted to add it post the integration PR.

sudswasavi:
build ako

monotosh-avi:
build ako

PR #668, opened 2021-10-20
Annotate the system namespaces for ako-infra

sudswasavi:
This commit takes care annotating the system namespace with the cloudName
and the seGroup keys which are to be used by AKO to sync the objects.

### DIFF
--- a/ako-infra/ingestion/avi_infra_controller.go
+++ b/ako-infra/ingestion/avi_infra_controller.go
@@ -119,6 +119,10 @@ func (a *AviControllerInfra) DeriveCloudNameAndSEGroupTmpl(tz string) (error, st
 }
 
 func (a *AviControllerInfra) SetupSEGroup(tz string) bool {
+	err, seGroup := lib.FetchSEGroupWithMarkerSet(a.AviRestClients.AviClient[0])
+	if err == nil && seGroup != "" {
+		utils.AviLog.Infof("SE Group: %s already configured with the marker labels: %s", seGroup, lib.GetClusterID())
+	}
 	// This method checks if the cloud in Avi has a SE Group template configured or not. If has the SEG template then it returns true, else false
 	err, cloudName, segUuid := a.DeriveCloudNameAndSEGroupTmpl(tz)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -187,6 +189,21 @@ func ConfigureSeGroup(client *clients.AviClient, seGroup *models.ServiceEngineGr
 
 }
 
+func (a *AviControllerInfra) AnnotateSystemNamespace(seGroup string, cloudName string) {
+	nsName := "vmware-system-ako"
+	nsObj, err := a.cs.CoreV1().Namespaces().Get(context.TODO(), nsName, metav1.GetOptions{})
+	if err != nil {
+		utils.AviLog.Fatalf("Failed to update the vmware-system-ako namespace due to the following error :%v", err.Error())
```
monotosh-avi, line 196: we can update the msg a bit, and say that get call failed.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -187,6 +189,21 @@ func ConfigureSeGroup(client *clients.AviClient, seGroup *models.ServiceEngineGr
...
+	}
+	// Update the namespace with the required annotations
+	nsObj.Annotations["ako.vmware.com/wcp-se-group"] = seGroup
+	nsObj.Annotations["ako.vmware.com/wcp-cloud-name"] = cloudName
+	_, err = a.cs.CoreV1().Namespaces().Update(context.TODO(), nsObj, metav1.UpdateOptions{})
+	if err != nil {
+		utils.AviLog.Fatalf("Error occurred while Updating namespace: %v", err)
```
monotosh-avi, line 203: I think we can print a warning here, return false and let the caller of SetupSEGroup() make a fatal call if required. It can be same for all false cases. 
Later we may have to add a few retries in case of failure.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -187,6 +189,21 @@ func ConfigureSeGroup(client *clients.AviClient, seGroup *models.ServiceEngineGr
 
 }
 
+func (a *AviControllerInfra) AnnotateSystemNamespace(seGroup string, cloudName string) {
+	nsName := "vmware-system-ako"
```
akshayhavile, line 193: We can use constant VMWARE_SYSTEM_AKO present in types.go

Dhivyaaj:
build ako

Dhivyaaj:
build ako

PR #669, opened 2021-10-21
Support IPv6 vip and static routes

sudswasavi:
This commit introduces the following:

- A helm knob to specify the IP scheme.
- Based on the IP scheme, AKO requests a VIP in either IPv6 format
or IPv4 format.
- Based on the IP scheme, the static routes are programmed.

It's assumed that in ClusterIP mode, the `Node` object of Kubernetes
and associated CNI metadata would be in IPv6 scheme to program the
nextHop in Avi.

### DIFF
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -214,6 +214,11 @@ spec:
               configMapKeyRef:
                 name: avi-k8s-config
                 key: autoFQDN
+          - name: IP_SCHEME
+            valueFrom:
+              configMapKeyRef:
+                name: avi-k8s-config
+                key: ipscheme
           {{ if .Values.persistentVolumeClaim }}
           - name: USE_PVC
             value: "true"
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -277,6 +277,9 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 	var worker *utils.FullSyncThread
 	var tokenWorker *utils.FullSyncThread
 	informersArg := make(map[string]interface{})
+	if lib.GetIpScheme() == lib.IPV6 {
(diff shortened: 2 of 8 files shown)


monotosh-avi (approved):

akshayhavile (commented):
We need to introduce configmap and values.yaml changes in ako-operator code base also.

helm/ako/templates/statefulset.yaml
```diff
@@ -214,6 +214,11 @@ spec:
               configMapKeyRef:
                 name: avi-k8s-config
                 key: autoFQDN
+          - name: IP_SCHEME
+            valueFrom:
+              configMapKeyRef:
+                name: avi-k8s-config
+                key: ipscheme
```
swathinsankaran, line 221: key must be **ipScheme** instead, right?

internal/k8s/ako_init.go
```diff
@@ -277,6 +277,9 @@ func (c *AviController) InitController(informers K8sinformers, registeredInforme
 	var worker *utils.FullSyncThread
 	var tokenWorker *utils.FullSyncThread
 	informersArg := make(map[string]interface{})
+	if lib.GetIpScheme() == lib.IPV6 {
+		utils.AviLog.Infof("AKO is running with IPv6 scheme")
```
swathinsankaran, line 281: It would be better to print the scheme in both IPv4 and IPv6 cases since the lib.GetIpScheme() is returning a string.

helm/ako/templates/statefulset.yaml
sudswasavi: Thanks corrected.

internal/k8s/ako_init.go
sudswasavi: Agreed. Done.

monotosh-avi:
build ako

Dhivyaaj:
build ako

PR #671, opened 2021-10-21
Fix multivip status update with secure ingress

aaha97:


aaha97:
build ako

aaha97:
build ako

chauhanshubham:
build ako

internal/rest/avi_obj_vs.go
```diff
@@ -371,38 +371,66 @@ func AviVsHttpPSAdd(vs_meta interface{}, isEVH bool) []*avimodels.HTTPPolicies {
...
-								ObjType: utils.L4LBService,
-								Op:      lib.UpdateStatus,
-								Options: &updateOptions,
-							}
-							status.PublishToStatusQueue(updateOptions.ServiceMetadata.NamespaceServiceName[0], statusOption)
-						} else if pool_cache_obj.ServiceMetadataObj.Namespace != "" {
-							if len(IPAddrs) == 0 {
-								IPAddrs = []string{parentVsObj.Vip}
+			//in secure case, PoolKeyCollection will be empty
+			if len(vs_cache_obj.PoolKeyCollection) > 0 {
```
chauhanshubham, line 375: I think this might not be true? In case of secure VSes, the insecure hosts might land up in the pool key collection, in case of SNI model. I see that we are processing SNIChildCollection *only* when PoolKeyCollection has no elements which doesn't seem to be right.

chauhanshubham:
The fix for this has gone as part of https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/686 
@aaha97 Can we now close this PR now?

PR #678, opened 2021-10-28
Fix: Refresh the session and retry the APIs calls when controller returns 403 status code

swathinsankaran:
This PR takes care of refreshing the session and retrying the below two API calls when the controller returns 403 HTTP status code in the response.

- api/initial-data - `done`
- api/cluster/runtime - `in-progress`


swathinsankaran:
Closing the PR since the controller changes that were restricting the two APIs were reverted.

PR #679, opened 2021-10-29
Add VCF network to IPAM and use SEgroup from NS annotation

monotosh-avi:
In this commit changes for ako-infra is integrated to be consumed in the main AKO.
Changes include,
- SEgroup and cloud annotation in the vmware-systems-ns are being consumed in main ako
- VCF Network is added to IPAM
- Code reorganisation to maintain correct sequence flow.

### DIFF
--- a/ako-infra/avirest/handle_netinfo.go
+++ b/ako-infra/avirest/handle_netinfo.go
@@ -166,7 +166,7 @@ func DeleteSegment(obj interface{}) {
 	} else {
 		utils.AviLog.Infof("key: %s, LSLR update not required in cloud: %s", objKey, utils.CloudName)
 	}
-	delCIDRFromNetwork(objKey, cidrs, false)
+	delCIDRFromNetwork(objKey, cidrs)
 }
 
 func matchCidrInNetwork(subnets []*models.Subnet, cidrs map[string]struct{}) bool {
@@ -188,15 +188,20 @@ func matchCidrInNetwork(subnets []*models.Subnet, cidrs map[string]struct{}) boo
 func findAndRemoveCidrInNetwork(subnets []*models.Subnet, cidrs map[string]struct{}) ([]*models.Subnet, bool) {
 	subnetsCopy := make([]*models.Subnet, len(subnets))
 	copy(subnetsCopy, subnets)
+	cidrsLen := len(cidrs)
+	if cidrsLen == 0 {
+		return subnetsCopy, true
+	}
 	for i, subnet := range subnets {
 		addr := *subnet.Prefix.IPAddr.Addr
 		mask := *subnet.Prefix.Mask
 		cidr := fmt.Sprintf("%s/%d", addr, mask)
 		if _, found := cidrs[cidr]; found {
 			subnetsCopy = append(subnetsCopy[:i], subnetsCopy[i+1:]...)
-		}
-		if len(cidrs) == 0 {
-			return subnetsCopy, true
+			cidrsLen -= 1
+			if cidrsLen == 0 {
+				return subnetsCopy, true
+			}
 		}
 	}
 	return subnetsCopy, false
@@ -305,7 +310,7 @@ func addNetworkInIPAM() {
 	}
 }
 
-func delCIDRFromNetwork(objKey string, cidrs map[string]struct{}, replaceAll bool) {
+func delCIDRFromNetwork(objKey string, cidrs map[string]struct{}) {
 	client := avicache.SharedAVIClients().AviClient[0]
 	netName := lib.GetVCFNetworkName()
 	method := utils.RestPut
--- a/ako-infra/avirest/handle_netinfo.go
+++ b/ako-infra/avirest/handle_netinfo.go
@@ -17,7 +17,7 @@ package avirest
 import (
 	"encoding/json"
 	"fmt"
-	"strconv"
+	"net"
 	"strings"
 
 	avicache "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/cache"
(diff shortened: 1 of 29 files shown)


monotosh-avi:
write /dev/stdout: resource temporarily unavailable error

monotosh-avi:
build ako

sudswasavi:
build ako

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -195,13 +196,33 @@ func ConfigureSeGroup(client *clients.AviClient, seGroup *models.ServiceEngineGr
...
+
+	uri := "/api/serviceenginegroup/?name=" + segroupName
+	response := models.ServiceEngineGroupAPIResponse{}
+	err := lib.AviGet(client, uri, &response)
+	if err != nil {
+		utils.AviLog.Warnf("Error during Get call for the SE group :%v", err.Error())
+		return false
+	}
+	utils.AviLog.Infof("Found Service Engine Group :%v", segroupName)
+	utils.IsVCFCluster()
```
sudswasavi, line 212: what is this being used for?

monotosh-avi: this is not required, removed

pkg/utils/utils.go
```diff
@@ -413,6 +413,11 @@ func IsServiceNSValid(namespace string) bool {
 	return true
 }
 
+func IsVCFCluster() bool {
+	vcfCluster := os.Getenv(VCF_CLUSTER)
+	return vcfCluster == "true"
```
sudswasavi, line 418: @monotosh-avi not as a part of this PR, but can we modify this function to take in the `parseBool` to validate the input instead of a string match?

monotosh-avi: done

sudswasavi (dismissed):

internal/k8s/ako_init.go
```diff
@@ -121,6 +121,93 @@ func deleteConfigFromConfigmap(cs kubernetes.Interface) bool {
...
+		if cloudName == "" {
+			// If the cloud name is blank - assume it to be Default-Cloud
+			cloudName = "Default-Cloud"
+		}
+		utils.SetCloudName(cloudName)
+		return true
+	}
+
+	nsName := utils.GetAKONamespace()
+	nsObj, err := c.informers.ClientSet.CoreV1().Namespaces().Get(context.TODO(), nsName, metav1.GetOptions{})
```
chauhanshubham, line 155: can we use informer cache here?

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -195,13 +196,32 @@ func ConfigureSeGroup(client *clients.AviClient, seGroup *models.ServiceEngineGr
...
+		return false
+	}
+	utils.AviLog.Infof("Found Service Engine Group :%v", segroupName)
+	return true
+}
+
 func (a *AviControllerInfra) AnnotateSystemNamespace(seGroup string, cloudName string) bool {
-	nsName := utils.VMWARE_SYSTEM_AKO
+	nsName := utils.GetAKONamespace()
 	nsObj, err := a.cs.CoreV1().Namespaces().Get(context.TODO(), nsName, metav1.GetOptions{})
```
chauhanshubham, line 217: can we use the informer cache call here as well?

internal/k8s/ako_init.go
```diff
@@ -121,6 +121,93 @@ func deleteConfigFromConfigmap(cs kubernetes.Interface) bool {
 	return true
 }
 
+func (c *AviController) SetSEGroupCloudName() bool {
+	seGroupToUse := lib.GetSEGNameEnv()
+	client := avicache.SharedAVIClients().AviClient[0]
+	var err error
+	// 2. Marker based (only advancedL4)
+	if seGroupToUse == "" && lib.GetAdvancedL4() {
+		err, seGroupToUse = lib.FetchSEGroupWithMarkerSet(client)
```
akshayhavile, line 130: In `FetchSEGroupWithMarkerSet`, we have reference to utils.CloudName. But We are setting `Cloudname` either in ako-infra code base or at line number: 150. I think this will result in failure for advanceL4 deployment.

akshayhavile: I think similar comment is true for function `validateAndConfigureSeGroup`

monotosh-avi: This is interesting. We were never setting cloud name before. init() function gets called automatically which sets the cloud name from env variable. So for normal AKO this works fine.  
In case of VCF we don't have the cloud name in env, so it has to be set explicitly.
Removed SetCloudName for non vcf setup.

ako-infra/ingestion/avi_infra_controller.go
monotosh-avi: We don't have access to informers here, and it should not be updated frequently, so I think its fine to use client here.

internal/k8s/ako_init.go
monotosh-avi: Good idea ! changed to informers.

ako-infra/avirest/handle_netinfo.go
```diff
@@ -182,16 +185,14 @@ func matchCidrInNetwork(subnets []*models.Subnet, cidrs []string) bool {
...
 	for i, subnet := range subnets {
 		addr := *subnet.Prefix.IPAddr.Addr
 		mask := *subnet.Prefix.Mask
 		cidr := fmt.Sprintf("%s/%d", addr, mask)
-		found, cidrs = utils.FindAndRemove(cidrs, cidr)
-		if found {
+		if _, found := cidrs[cidr]; found {
 			subnetsCopy = append(subnetsCopy[:i], subnetsCopy[i+1:]...)
 		}
 		if len(cidrs) == 0 {
```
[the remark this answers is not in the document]
monotosh-avi: Good catch. There is a bug here. We should return true only if all elements of cidrs are matched and removed from subnets. Fixed it.

chauhanshubham (dismissed):

monotosh-avi:
build ako

chauhanshubham (dismissed):

chauhanshubham (approved):

akshayhavile (approved):

PR #681, opened 2021-11-02
Fix: New flag to retain shared virtual service IPs

swathinsankaran:
This PR introduces a new flag to retain the shared VS IPs.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -1463,5 +1463,9 @@ var AviSecretInitialized bool
 
 func GetRetainSharedVSIPs() bool {
 	retainSharedVSIPs := os.Getenv(RETAIN_SHARED_VS_IPS)
-	return retainSharedVSIPs == "true"
+	ok, err := strconv.ParseBool(retainSharedVSIPs)
+	if err != nil {
+		utils.AviLog.Warnf("Failed to parse the retain shared VS IP data, err: %v", err)
+	}
+	return ok
 }
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -1461,7 +1461,7 @@ func IsVCFCluster() bool {
 var VCFInitialized bool
 var AviSecretInitialized bool
 
-func GetRetainSharedVSIPs() bool {
+func IsRetainSharedVSIPsEnabled() bool {
 	retainSharedVSIPs := os.Getenv(RETAIN_SHARED_VS_IPS)
 	ok, err := strconv.ParseBool(retainSharedVSIPs)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


internal/lib/lib.go
```diff
@@ -1460,3 +1460,8 @@ func IsVCFCluster() bool {
 
 var VCFInitialized bool
 var AviSecretInitialized bool
+
+func GetRetainSharedVSIPs() bool {
+	retainSharedVSIPs := os.Getenv(RETAIN_SHARED_VS_IPS)
```
sudswasavi, line 1465: it would be good to do a `parseBool` check here instead of doing a string compare.

sudswasavi (commented):
Hi @swathinsankaran changes look broadly good.
1. You will need to update the doc here: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/docs/values.md (pls update in separate commit so that you can point the doc team before release to it, for all doc changes)
2. Parameters table here: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/docs/install/helm.md#parameters
2.

sudswasavi:
Can you also add a couple of UTs for this case?

internal/lib/lib.go
swathinsankaran: I have removed the string comparison and added the parseBool check as per the comment.

internal/rest/dequeue_nodes.go
```diff
@@ -60,7 +60,7 @@ func (rest *RestOperations) DequeueNodes(key string) {
 		utils.AviLog.Warnf("key: %s, msg: no model found for the key", key)
 		// In the case of L7 shared VS, the following condition check makes sure the
 		// VIPs persist over AKO reboot.
-		if lib.IsShardVS(key) {
+		if lib.IsShardVS(key) && lib.GetRetainSharedVSIPs() {
```
akshayhavile, line 63: @swathinsankaran : If we change condition to `lib.GetRetainSharedVSIPs() && lib.IsShardVS(key)`, it will be more faster as default value to flag is `false`. So we will avoid un-necessary computation.

akshayhavile: @sudswasavi : Are we proposing this solution for scenario if customer wants to migrate from shardVsSize "small" to "large" or other combinations of shardVsSize?

akshayhavile (commented):
@swathinsankaran: We also need to do changes in ako operator code base once this gets approved or you can do it in parallel.

swathinsankaran:
@akshayhavile I have added the ako-operator changes also. Could you please review it?

internal/rest/dequeue_nodes.go
swathinsankaran: I have addressed the comment.

swathinsankaran:
As per the discussion with @sudswasavi, @akshayhavile, a different solution is required other than adding a new flag addition for better convenience. Hence I am closing this PR as it is no longer required.

PR #688, opened 2021-11-12
Update annotation in AKO statefulset after avi object deletion through deleteConfig

monotosh-avi:
Currently status of ako statefulset is updated with condition values after objects are deleted
by setting deleteConfig to true. However, in kubernetes 1.22 this can not be done because kubernetes
automatically removes this status. To get around this problem annotation of the statefulset would be
updated instead of status.

When object deletion starts we would set the annotation as:
  annotations:
    ObjectDeletionStatus: Started

After object deletion is completed, this would be changed to
    ObjectDeletionStatus: Done

### DIFF
--- a/internal/status/statefulset_status.go
+++ b/internal/status/statefulset_status.go
@@ -16,29 +16,16 @@ package status
 
 import (
 	"context"
+	"encoding/json"
 
-	appsv1 "k8s.io/api/apps/v1"
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
 )
 
-var msgForReason = map[string]string{
-	lib.ObjectDeletionStartStatus:   "Started deleting objects",
-	lib.ObjectDeletionDoneStatus:    "Successfully deleted all objects",
-	lib.ObjectDeletionTimeoutStatus: "Error, timed out while deleting objects",
-}
-
-func msgFoundInStatus(conditions []appsv1.StatefulSetCondition, msg string) bool {
-	for _, c := range conditions {
-		if c.Type == lib.AKOConditionType && c.Message == msg {
-			return true
-		}
-	}
-	return false
-}
+const ObjectDeletionStatus = "AviObjectDeletionStatus"
 
 // ResetStatefulSetStatus removes the condition set by AKO from AKO statefulset
 func ResetStatefulSetStatus() {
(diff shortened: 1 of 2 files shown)


internal/status/statefulset_status.go
```diff
@@ -68,52 +57,69 @@ func ResetStatefulSetStatus() {
...
-		if c.Type == lib.AKOConditionType {
-			ss.Status.Conditions[i].Reason = reason
-			ss.Status.Conditions[i].Message = msg
-			ss.Status.Conditions[i].Status = statusCondition
-			ss.Status.Conditions[i].LastTransitionTime = currentTime
-			foundCondition = true
-			break
-		}
+	//Remove any status from previous versions of AKO
+	ResetStatefulSetStatus()
```
sudswasavi, line 92: is this required still?

internal/status/statefulset_status.go
```diff
@@ -16,32 +16,21 @@ package status
...
-
-func msgFoundInStatus(conditions []appsv1.StatefulSetCondition, msg string) bool {
-	for _, c := range conditions {
-		if c.Type == lib.AKOConditionType && c.Message == msg {
-			return true
-		}
-	}
-	return false
-}
+const ObjectDeletionStatus = "ObjectDeletionStatus"
```
sudswasavi, line 28: Not too particular but shall we call it `AviObjectDeletionStatus` ?

monotosh-avi: yeah, I'll change it

internal/status/statefulset_status.go
monotosh-avi: I thought that - if previous version of AKO had set the status value in statefulset, and then we move to a new version of AKO, then the status has to be cleaned up. 

Status in particular is not straightforward to cleanup using kubectl, so it may linger on. If thats fine, we can remove this call.

sudswasavi (approved):

internal/status/statefulset_status.go
akshayhavile: will it create infinite loop ? ... ResetStatefulSetStatus --> ResetStatefulSetAnnotation and ResetStatefulSetAnnotation -->ResetStatefulSetStatus

internal/status/statefulset_status.go
```diff
@@ -68,52 +57,69 @@ func ResetStatefulSetStatus() {
...
+	_, err = utils.GetInformers().ClientSet.AppsV1().StatefulSets(utils.GetAKONamespace()).Patch(context.TODO(), ss.Name, types.MergePatchType, payloadBytes, metav1.PatchOptions{})
 
-	u, err := utils.GetInformers().ClientSet.AppsV1().StatefulSets(utils.GetAKONamespace()).UpdateStatus(context.TODO(), ss, metav1.UpdateOptions{})
 	if err != nil {
-		utils.AviLog.Warnf("Error in patching ako statefulset: %v", err)
+		utils.AviLog.Warnf("Error in patching ako statefulset annotation: %v", err)
 		return
 	}
-	utils.AviLog.Debugf("Successfully updated ako statefulset: %v", u)
+	utils.AviLog.Debugf("Successfully updated annotation %s in ako statefulset: %v", ObjectDeletionStatus)
```
akshayhavile, line 124: We are missing %v value.

internal/status/statefulset_status.go
monotosh-avi: Great catch, removed ResetStatefulSetAnnotation call from ResetStatefulSetStatus

internal/status/statefulset_status.go
monotosh-avi: updated

akshayhavile (approved):

PR #689, opened 2021-11-12
Update annotation in AKO statefulset after avi object deletion through deleteConfig

monotosh-avi:
Currently status of ako statefulset is updated with condition values after objects are deleted
by setting deleteConfig to true. However, in kubernetes 1.22 this can not be done because kubernetes
automatically removes this status. To get around this problem annotation of the statefulset would be
updated instead of status.

When object deletion starts we would set the annotation as:
  annotations:
    AviObjectDeletionStatus: Started

After object deletion is completed, this would be changed to
    AviObjectDeletionStatus: Done

(cherry picked from commit 71002106456eaedbd3bd73fb5ead0593e76985a5)

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -244,7 +244,7 @@ func (c *AviController) HandleConfigMap(k8sinfo K8sinformers, ctrlCh chan struct
 			lib.SetGRBACSupport()
 			delModels := delConfigFromData(cm.Data)
 			if !delModels {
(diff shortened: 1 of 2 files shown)


sudswasavi (dismissed):

akshayhavile (dismissed):

sudswasavi:
build ako

monotosh-avi:
build ako

chauhanshubham (approved):

akshayhavile (approved):

PR #690, opened 2021-11-15
Introduce TCPSettings in Hostrule for ParentVS configuration

chauhanshubham:
This commit introduces tcpSettings section that allows for
parent VS tcp settings like custom ports to be opened as part
of the virtual service, or a static load balancer IP to be
assigned to the VS. Includes a few validaitons around the
new fields too.


chauhanshubham:
This PR does not cover AKO behaviour when a loadBalancerIP which leads to a VSVIP Post/Put error/failure is provided.

chauhanshubham:
build ako

chauhanshubham:
build ako

tests/integrationtest/l4_service_test.go
```diff
@@ -962,46 +961,3 @@ func TestInfraSettingChangeMapping(t *testing.T) {
 	TeardownAviInfraSetting(t, settingName2)
 	TearDownTestForSvcLB(t, g)
 }
-
-func TestBootupServiceLBStatusPersistence(t *testing.T) {
```
akshayhavile, line 966: Are we covering this use case somewhere else?

ako-operator/helm/ako-operator/crds/ako.vmware.com_hostrules.yaml
```diff
@@ -77,6 +77,20 @@ spec:
...
+                      listeners:
+                        items:
+                          properties:
+                            port:
+                              type: integer
+                            enableSSL:
+                              type: boolean
+                          type: object
+                        type: array
+                      loadBalancerIP:
```
akshayhavile, line 91: Should we include IP type also v4 /V6?

akshayhavile: Can user specify public ip over here for cloud deployments?

akshayhavile: How will it behave if EnablepublicIP is true and we have specified IP over here?

chauhanshubham: We don't have ipv6 support yet, later we can possibly auto detect it from the field.
Not sure about the behaviour but programmatically, handling enablePublicIP and static IP is mutually exclusive.

tests/integrationtest/l4_service_test.go
chauhanshubham: yes bootup scenarios are covered in the functional tests.

akshayhavile (approved):
LGTM.

monotosh-avi (approved):

PR #692, opened 2021-11-15
Support for enabling/disabling non-significant logs through Hostrule CRD

swathinsankaran:
This PR adds support for enabling/disabling non-significant logs through Hostrule CRD.


internal/nodes/crd_translator.go
```diff
@@ -107,6 +109,17 @@ func BuildL7HostRule(host, key string, vsNode AviVsEvhSniModel) {
 			Value:  hostrule.Namespace + "/" + hostrule.Name,
 			Status: lib.CRDActive,
 		}
+
+		if (akov1alpha1.HostRuleAnalyticsPolicy{}) != hostrule.Spec.VirtualHost.AnalyticsPolicy {
```
monotosh-avi, line 113: @chauhanshubham Can you check if this is fine in case the field AnalyticsPolicy is omitted from the hostrule ? I think you had to use a pointer here ?

chauhanshubham: I think this should be fine. Although I would prefer declaring `AnalyticsPolicy` to be of type `* HostRuleAnalyticsPolicy` and comparing here with nil value, but i have no good reason to use one or the other.
@swathinsankaran can you please check this scenario once to confirm.

swathinsankaran: Yes, the pointer will be the best approach. I have changed the code accordingly.

pkg/apis/ako/v1alpha1/hostrule.go
```diff
@@ -91,3 +94,15 @@ type HostRuleList struct {
 
 	Items []HostRule `json:"items"`
 }
+
+// HostRuleAnalyticsPolicy holds analytics policy objects
+type HostRuleAnalyticsPolicy struct {
+	FullClientLogs FullClientLogs `json:"fullClientLogs,omitempty"`
+	AllHeaders     bool           `json:"allHeaders,omitempty"`
```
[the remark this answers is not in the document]
swathinsankaran: I have renamed the field to `LogAllHeaders`.

chauhanshubham (approved):

swathinsankaran:
build ako

monotosh-avi (approved):

akshayhavile:
@swathinsankaran :  Yaml changes are missing in hostrule yaml file present ako operator. You can take it in next PR.

monotosh-avi:
@akshayhavile  I think as per the code, it would work for shared VS too. I think it should be fine. Do you see any issue if that is the case ?

akshayhavile:
> @akshayhavile I think as per the code, it would work for shared VS too. I think it should be fine. Do you see any issue if that is the case ?

@monotosh-avi : As per current changes, analytics policy will be applied to child VS only. We are missing parent vS changes at rest layer if we want to associate analyticspolicy with shared VS.

PR #702, opened 2021-11-23
Handling non admin tenant with SE_in_provider_context set

aaha97:
When any SE group related op fails with provided tenant(GET/PUT), an assumption is made that the tenant is configured with se_in_provider_context set to True on the AVI controller.
A context switch is then done for the session and the operation is retried.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -2599,12 +2599,21 @@ func validateAndConfigureSeGroup(client *clients.AviClient, returnErr *error) bo
 	}
 	seGroupSet.Insert(lib.GetSEGName())
 
+	SetAdminTenant := session.SetTenant(lib.GetAdminTenant())
+	SetTenant := session.SetTenant(lib.GetTenant())
+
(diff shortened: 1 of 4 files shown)


internal/cache/controller_obj_cache.go
```diff
@@ -2689,27 +2703,39 @@ func DeConfigureSeGroupLabels() {
...
-		SetTenant := session.SetTenant(lib.GetTenant())
 		SetAdminTenant(client.AviSession)
 		defer SetTenant(client.AviSession)
 	}
 
 	err = lib.AviPut(client, uri, seGroup, response)
 	if err != nil {
-		utils.AviLog.Warnf("Deconfiguring SE Group labels failed on %v with error %v", segName, err.Error())
-		return
+		SetAdminTenant(client.AviSession)
```
monotosh-avi, line 2715: Can we check the error type and then switch tenant ?

monotosh-avi (approved):

aaha97:
build ako

chauhanshubham (approved):

PR #703, opened 2021-11-29
kube-builder tags to access CRDs for VCF

monotosh-avi:
in VCF cluster, ako needs permission to access namespacenetworkinfos and akobootstrapconditions

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -42,9 +42,11 @@ var ctrlonce sync.Once
 
 // These tags below are only applicable in case of advanced L4 features at the moment.
 
+// +kubebuilder:rbac:groups=ncp.vmware.com,resources=akobootstrapconditions;akobootstrapconditions/status,verbs=get;list;watch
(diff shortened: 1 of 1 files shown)


sudswasavi:
@monotosh-avi the cluster roles of the AKO objects needn't be updated. You can see the example here: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/internal/k8s/controller.go#L45 
You have annotate the code with the cube-builder annotations for this to work.

monotosh-avi:
> @monotosh-avi the cluster roles of the AKO objects needn't be updated. You can see the example here: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/internal/k8s/controller.go#L45
> You have annotate the code with the cube-builder annotations for this to work.

@sudswasavi  updated kubebuilder tags

sudswasavi (approved):

sudswasavi:
force merging this since it does not have any code changes.

PR #708, opened 2021-12-07
Fix: VSes not getting deleted from AVI Controller after Route deletion from cluster

swathinsankaran:
This PR fixes the issue of VSes not getting deleted from AVI Controller after Route deletion from the cluster.

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -72,6 +72,9 @@ func (rest *RestOperations) DequeueNodes(key string) {
 			// In the case of L7 shared VS, the following condition check makes sure the
 			// VIPs persist over AKO reboot.
 			if lib.IsShardVS(key) && !lib.DisableSync {
+				// Save the model in the cache as it is required when the user
+				// sets the deleteconfig flag to `true`.
+				objects.SharedAviGraphLister().Save(key, nil)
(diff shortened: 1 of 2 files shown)


internal/rest/dequeue_nodes.go
```diff
@@ -69,6 +69,12 @@ func (rest *RestOperations) DequeueNodes(key string) {
 		}
 		if vs_cache_obj != nil {
 			utils.AviLog.Infof("key: %s, msg: nil model found, this is a vs deletion case", key)
+			// In the case of L7 shared VS, the following condition check makes sure the
+			// VIPs persist over AKO reboot.
+			if lib.IsShardVS(key) && !lib.DisableSync {
+				utils.AviLog.Infof("key: %s, VIPs needs to be retained, not performing vs deletion", key)
+				return
+			}
 			rest.deleteVSOper(vsKey, vs_cache_obj, namespace, key, false, false)
```
monotosh-avi, line 78: deleteVSOper internally tries deletes other objects like HTTP PolicySet, Poolgroup, Datascript which are referred in the VS. If we skip calling deleteVSOper, then those objects won't be deleted. This would be a problem. 
Also deleteVSOper can be called from multiple functions, so I think its better to add the check inside deleteVSOper function.

PR #710, opened 2021-12-07
Reverts the code to skip the shared empty parent VS deletion

swathinsankaran:
This PR reverts the code to skip the shared empty parent VS deletion.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -954,18 +954,6 @@ func (c *AviController) FullSyncK8s() error {
 	syncNamespace := lib.GetNamespaceToSync()
 	for _, vsCacheKey := range vsKeys {
 		modelName := vsCacheKey.Namespace + "/" + vsCacheKey.Name
-		ok, _ := objects.SharedAviGraphLister().Get(modelName)
-		if !ok {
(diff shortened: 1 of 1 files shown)


monotosh-avi (approved):

internal/k8s/ako_init.go
```diff
@@ -954,18 +954,6 @@ func (c *AviController) FullSyncK8s() error {
 	syncNamespace := lib.GetNamespaceToSync()
 	for _, vsCacheKey := range vsKeys {
 		modelName := vsCacheKey.Namespace + "/" + vsCacheKey.Name
-		ok, _ := objects.SharedAviGraphLister().Get(modelName)
```
akshayhavile, line 957: @swathinsankaran : similar set of changes is present in PR: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/708/ .
Can you please let me know reason behind raising different PR with similar set of changes.

swathinsankaran: @akshayhavile The VSes were not cleaning up properly due to the code changes that were done earlier to skip the deletion of empty Parent VS. This PR is to revert those changes and unblock system testing. The other PR is for the reintroduction of this feature which requires some more code changes and thorough testing.

akshayhavile (approved):

PR #714, opened 2021-12-10
Introduce fqdnType (Exact/Contains/Wildcard) in HostRule

chauhanshubham:
The field fqdnType determines the interpretation of the fqdn matching.
fqdnType can be one of the following:
Exact - matches the fqdn exactly
Contains - considers the value specified in `fqdn` for a substring match
Wildcard - matches based on fqdn suffix, this must contain a `*` as a prefix
(for instance: *.tenant.com)
This implementation also helps in **easily** specifying the Hostrule for parent VSes
for which AKO programs the fqdn automatically (without user input). The users
can specify a substring of the parent VS name, to attach the hostrule to that VS.

### DIFF
--- a/internal/objects/crdrules.go
+++ b/internal/objects/crdrules.go
@@ -15,7 +15,10 @@
 package objects
 
 import (
+	"strings"
 	"sync"
+
+	akov1alpha1 "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/apis/ako/v1alpha1"
 )
 
 var CRDinstance *CRDLister
@@ -32,6 +35,7 @@ func SharedCRDLister() *CRDLister {
 			FqdnToGSFQDNCache:      NewObjectMapStore(),
 			FqdnSharedVSModelCache: NewObjectMapStore(),
 			SharedVSModelFqdnCache: NewObjectMapStore(),
+			FqdnFqdnTypeCache:      NewObjectMapStore(),
 		}
 	})
 	return CRDinstance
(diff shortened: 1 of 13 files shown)


chauhanshubham:
build ako

sudswasavi (commented):
The only problem with this approach is while it servers clubbing multiple FQDNs that match a pattern, it does not help in adding two FQDNs that do not follow a pattern. Any thoughts on that?

chauhanshubham:
> The only problem with this approach is while it servers clubbing multiple FQDNs that match a pattern, it does not help in adding two FQDNs that do not follow a pattern. Any thoughts on that?

Does this equate to the following intent - "Apply configurations to all but 2 FQDNs (lets say abc.com, def.com)"?
I believe that might over complicate the intent of hostrule configurations. hostrule is applied on specified FQDNs, and is not intended to be **not** applied for specific fqdns.

sudswasavi:
> > The only problem with this approach is while it servers clubbing multiple FQDNs that match a pattern, it does not help in adding two FQDNs that do not follow a pattern. Any thoughts on that?
> 
> Does this equate to the following intent - "Apply configurations to all but 2 FQDNs (lets say abc.com, def.com)"? I believe that might over complicate the intent of hostrule configurations. hostrule is applied on specified FQDNs, and is not intended to be **not** applied for specific fqdns.

No - I meant the following - Apply the same hostrule for def.com and abc.com using the same object. But there's a way out of this, that is to create multiple host rules. So we are good.

internal/objects/crdrules.go
```diff
@@ -178,12 +239,37 @@ func (c *CRDLister) UpdateFqdnHTTPRulesMappings(fqdn, path, httprule string) {
...
+func (c *CRDLister) GetFQDNToSharedVSModelMapping(fqdn string) (bool, []string) {
+	c.NSLock.Lock()
+	defer c.NSLock.Unlock()
+	oktype, fqdnType := c.FqdnFqdnTypeCache.Get(fqdn)
+	if !oktype || fqdnType == "" {
+		fqdnType = string(akov1alpha1.Exact)
 	}
-	return true, modelName.(string)
+
+	allFqdns := c.FqdnSharedVSModelCache.GetAllKeys()
```
sudswasavi, line 250: it seems you are nesting a lock here. The GetAllKeys already has a READ lock inside of it. Is this intentional?

sudswasavi (commented):
@swathinsankaran pls review this PR w.r.t your multi-alias changes.
@chauhanshubham can you pls add UTs?

internal/objects/crdrules.go
```diff
@@ -76,6 +82,45 @@ func (c *CRDLister) GetFQDNToHostruleMapping(fqdn string) (bool, string) {
...
+				returnHostrules = append(returnHostrules, hostrule.(string))
+				break
+			}
+		} else if fqdnType == string(akov1alpha1.Contains) && strings.Contains(fqdn, mFqdn) {
+			if found, hostrule := c.FqdnHostRuleCache.Get(mFqdn); found {
+				returnHostrules = append(returnHostrules, hostrule.(string))
+				break
+			}
+		} else if fqdnType == string(akov1alpha1.Wildcard) && strings.HasPrefix(mFqdn, "*") {
+			wildcardFqdn := strings.Split(mFqdn, "*")[1]
```
sudswasavi, line 108: what if the user just specified a "*" and nothing else. Does this code fail in that case?

chauhanshubham: you're right yes, thanks.

chauhanshubham: actually no, this is fine. https://go.dev/play/p/NNRDA0cexeT

internal/objects/crdrules.go
```diff
@@ -178,12 +237,35 @@ func (c *CRDLister) UpdateFqdnHTTPRulesMappings(fqdn, path, httprule string) {
...
+	oktype, fqdnType := c.FqdnFqdnTypeCache.Get(fqdn)
+	if !oktype || fqdnType == "" {
+		fqdnType = string(akov1alpha1.Exact)
+	}
+
+	allFqdns := c.FqdnSharedVSModelCache.GetAllKeys()
+	returnModelNames := []string{}
+	for _, mFqdn := range allFqdns {
+		if fqdnType == string(akov1alpha1.Exact) && mFqdn == fqdn {
+			if found, modelName := c.FqdnSharedVSModelCache.Get(mFqdn); found {
```
monotosh-avi, line 250: if the fqdn is not found in the Cache, we would break here and then return true in line no 268 with empty list. Is that fine ?

monotosh-avi (commented):
Overall looks fine to me. 
One issue could be - if one fqdn is matched by multiple hostrules. e.g. lets say we have two hostrules with contains match foo and bar respectively. Then the fqdn - foobar would match with both hostrules. It might be a problem if the behaviour gets unpredictable in this case.

monotosh-avi (approved):

sudswasavi (approved):

PR #718, opened 2021-12-14
Retain the shared VS VIP on ako reboot

swathinsankaran:
This PR adds support to retain shared VS VIP on AKO reboots.

The following approaches have to be taken by the user to remove the shared VS VIP after this PR.

1. Toggle the `enableEVH` flag in config map and reboot the ako
2. Manually delete from the UI or shell.


monotosh-avi (approved):

monotosh-avi (commented):
As discussed, we would have to make a put call and remove the fqdns from vsvip.

swathinsankaran:
build ako

monotosh-avi (dismissed):

monotosh-avi (approved):

sudswasavi (approved):

PR #724, opened 2021-12-16
AKO CRD documentation and missing ako-operator install workflow

chauhanshubham:


ako-operator/controllers/crd.go
```diff
@@ -135,6 +153,74 @@ func createHostRuleCRD(clientset *apiextension.ApiextensionsV1Client, log logr.L
 											},
 										},
 									},
+									"analyticsPolicy": {
+										Type: "object",
```
RiyaGaniga-avi, line 157: 'analyticsPolicy' and 'tcpSettings' additions are present as a part of https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/711 too. We would need to edit either of the PRs based on which gets merged first.

sudswasavi (approved):

ako-operator/controllers/crd.go
chauhanshubham: #711 has a merge conflict and needs another update anyways, let's get this one in first then if that's ok? :D

RiyaGaniga-avi: Fine with me

RiyaGaniga-avi (approved):

PR #726, opened 2021-12-17
Patch the l4 policy rule name to be similar to the corresponding Pool name

chauhanshubham:


internal/rest/avi_obj_l4ps.go
```diff
@@ -52,7 +55,8 @@ func (rest *RestOperations) AviL4PSBuild(hps_meta *nodes.AviL4PolicyNode, cache_
 	var l4rules []*avimodels.L4Rule
 	for _, hppmap := range hps_meta.PortPool {
 		if hppmap.Port != 0 {
-			ruleName := name + strconv.Itoa(int(hppmap.Port))
+			// Keep the l4 policy rule name similar to the Pool name it corresponds to.
+			ruleName := hppmap.Pool
```
akshayhavile, line 59: @chauhanshubham: This change will impact only new l4 policies. Will that be ok?

chauhanshubham: nice catch, i've taken care of this in the checksum calculation. please check.

akshayhavile (approved):
LGTM

monotosh-avi (approved):

PR #730, opened 2021-12-20
Do not overwrite the ipam network subnet config in Vip before PUT

chauhanshubham:
This prevents AKO to overwrite the ipam network subnet configuration
in the Vip while doing a PUT VSVIP call.
Resolves AV-132243

### DIFF
--- a/internal/rest/avi_obj_vsvip.go
+++ b/internal/rest/avi_obj_vsvip.go
@@ -90,8 +90,11 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, vsCach
 		vsvip.DNSInfo = dns_info_arr
 		vsvip.VsvipCloudConfigCksum = &cksumstr
 
-		// handling static IP and networkName (infraSetting) updates.
-		if GetIPAMProviderType() != lib.IPAMProviderInfoblox {
+		noVipUpdatesAllowedForIPAMTypes := []string{
+			lib.IPAMProviderInfoblox,
+			lib.IPAMProviderCustom,
+		}
+		if !utils.HasElem(noVipUpdatesAllowedForIPAMTypes, GetIPAMProviderType()) {
 			vip := &avimodels.Vip{
 				VipID:                  &vipId,
 				AutoAllocateIP:         &autoAllocate,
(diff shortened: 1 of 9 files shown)


monotosh-avi (approved):

chauhanshubham:
build ako

internal/rest/avi_obj_vsvip.go
```diff
@@ -75,34 +75,25 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, vsCach
...
 		} else {
 			// Set the IPAM network subnet for all clouds except AWS and Azure
 			if len(vsvip_meta.VipNetworks) != 0 {
-				if vip.IPAMNetworkSubnet == nil {
-					vip.IPAMNetworkSubnet = &avimodels.IPNetworkSubnet{}
+				if len(vsvip.Vip) == 1 {
+					vip := vsvip.Vip[0]
+					networkRef := "/api/network/?name=" + vsvip_meta.VipNetworks[0].NetworkName
+					vip.IPAMNetworkSubnet.NetworkRef = &networkRef
+					vsvip.Vip = []*avimodels.Vip{vip}
```
sudswasavi, line 87: This code implies that we will put whatever networkName is provided by the user into the vip object, keeping everything else including the subnet unchanged.
Can the user not associate a different subnet with the vip object along with a network change?

sudswasavi: If really this has been the behaviour then I am afraid the vip network switch just does not work. Since each vip network would also have an associated subnet. My suggestion would be to not allow this switch at all, keep the code simple and just call it out in the doc.

chauhanshubham: one network to another would still work, 
same network different subnet won’t I believe
the current implementation is limited.

sudswasavi: But you are copying the rest of the vip object, which implies the subnet of the older network would be copied into the new network as well, is that practical?

chauhanshubham:
build ako

internal/rest/avi_obj_vsvip.go
```diff
@@ -76,44 +91,47 @@ func (rest *RestOperations) AviVsVipBuild(vsvip_meta *nodes.AviVSVIPNode, vsCach
 		vsvip.VsvipCloudConfigCksum = &cksumstr
 
 		// handling static IP and networkName (infraSetting) updates.
-		vip := &avimodels.Vip{
-			VipID:                  &vipId,
-			AutoAllocateIP:         &autoAllocate,
-			AutoAllocateFloatingIP: vsvip_meta.EnablePublicIP,
-		}
+		if GetIPAMProviderType() != lib.IPAMProviderInfoblox || GetIPAMProviderType() != lib.IPAMProviderCustom {
```
sudswasavi, line 94: A better check is to just do the utils.hasElem check here by declaring a slice variable that lists all of these exceptions in once place.

sudswasavi (approved):

sudswasavi:
build AKOs

monotosh-avi (approved):

monotosh-avi:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

chauhanshubham:
build ako

monotosh-avi (approved):

chauhanshubham:
write /dev/stdout: resource temporarily unavailable

chauhanshubham:
build ako

monotosh-avi (approved):

swathinsankaran (approved):

PR #732, opened 2021-12-20
[AKO Chart] Set helm chart to allow the use of an existing secret

EStork09:
# What changes?
- Updating helm chart to allow for existing secret to be used

# Why even make the changes?
- GitOps workflows ideally shouldn't have credentials in plain text in a values.yaml file
- Mainly around gitops, but allows for the secret to managed by something [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets)


chauhanshubham:
Instead of introducing a new field `existingSecret` I think it might be better to decide for secret creation based on existing fields. We could check whether the `username` & `password` is provided in the helm chart and then only proceed with creating the Secret object. cc @akshayhavile

EStork09:
Sorry @Useurmind I totally forgot about this until you added this.

@chauhanshubham can you review this again? I have updated to not require the new value

akshayhavile:
build ako

swathinsankaran:
build ako

swathinsankaran:
@akshayhavile, @apalsule,

This commit would cause one side effect.

**Actual behavior:** The secret created by the `helm install` contains an annotation and upon execution of `helm delete`, the helm cleans up this resource. This is applicable to all the resources created using the helm.

With this commit, we will be breaking it. The `helm delete` will not clean up the secret from the cluster since the annotation will have a different release-name.

An example of the secret annotation is show below:
```
metadata:
  annotations:
    meta.helm.sh/release-name: ci-build-1-1686921510
    meta.helm.sh/release-namespace: avi-system
```

Helm list output: 
```
aviuser@avi-dev:~$ helm list
NAME                 	NAMESPACE 	REVISION	UPDATED                                	STATUS  	CHART     	APP VERSION
ci-build-1-1686921510	avi-system	1       	2023-06-16 13:18:31.095325174 +0000 UTC	deployed	ako-0.1.0 	1.10.1-5017
aviuser@avi-dev:~$
```

The user has to manually clean up the secret during a `helm delete` and we have to document this.

Please let me know your views.

DixitAakash:
build ako

DixitAakash:
Closed the PR as the suggested changes have been merged as part of https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/1384

PR #738, opened 2021-12-22
Doc: FAQ changes for `retain VS VIP over AKO reboot`

swathinsankaran:
This PR addresses the document changes for `retain VS VIP over AKO reboot`.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -30,6 +30,9 @@ After all relevant objects gets deleted from Avi, the reason is changed to objDe
 
 To re-create the objects in Avi, the configmap has to be edited to set deleteConfig: "false". After this, the condition in ako statefulset of type ako.vmware.com/ObjectDeletionInProgress is deleted automatically.
 
+**Note**: To cleanup the shared VS VIP that is retained, the user has to:
+1. Manually delete it from the controller UI or shell.
+2. Toggle the `enableEVH` flag in configmap and reboot the AKO.
 
 #### How is the Shared VS lifecycle controlled?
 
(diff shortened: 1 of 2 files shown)


sudswasavi (approved):

docs/faq.md
```diff
@@ -30,6 +30,9 @@ After all relevant objects gets deleted from Avi, the reason is changed to objDe
 
 To re-create the objects in Avi, the configmap has to be edited to set deleteConfig: "false". After this, the condition in ako statefulset of type ako.vmware.com/ObjectDeletionInProgress is deleted automatically.
 
+**Note**: To cleanup the shared VS VIP that is retained, the user has to:
```
akshayhavile, line 33: should it be part of section "#### How is the Shared VS lifecycle controlled" with more elaboration of shared vs vip behaviour change?

docs/faq.md
```diff
@@ -30,6 +30,9 @@ After all relevant objects gets deleted from Avi, the reason is changed to objDe
...
+1. Manually delete it from the controller UI or shell.
+2. Toggle the `enableEVH` flag in configmap and reboot the AKO.
```
akshayhavile, line 35: I did not get relevance of `enableEVH` flag to clean up shared vs vip

docs/faq.md
swathinsankaran: I have moved it to the section `How is the Shared VS lifecycle controlled`.

docs/faq.md
swathinsankaran: I have removed it as the flag `enableEVH` usage, in this case, is not the right way. Thanks.

akshayhavile (approved):

swathinsankaran:
build ako

sudswasavi (approved):

PR #742, opened 2021-12-23
Always recreate IPAMNetworkSubnet while updating VSVIP

chauhanshubham:
copying the IPAMNetworkSubnet does not work, and the vip.DiscoveredNetworks gets removed.
which causes the vsvip to go down eventually after enabling/disabling the virtualservice


anish-avi:
build ako

monotosh-avi (approved):

akshayhavile (approved):

PR #751, opened 2022-01-26
Hostrule fixes relating to fqdnType Wildcard and changes between fqdnTypes

chauhanshubham:
This commit fixes the following problems in the object cache mapping
- solves a problem with Wildcard based fqdnType hostrules, that was not
returning all Ingresses in which the host the wildcard fqdn.
- solves a problem with `Contains` based fqdnType hostrules, that was not
correctly returning all Shared VS models, particularly in case of
create-delete-create workflows.
- solves problems related to fqdnType switching.

### DIFF
--- a/internal/nodes/hostname_cache.go
+++ b/internal/nodes/hostname_cache.go
@@ -15,10 +15,11 @@
 package nodes
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/objects"
-
+	akov1alpha1 "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/apis/ako/v1alpha1"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
 )
 
@@ -98,6 +99,25 @@ func (h *HostNamePathStore) GetHostPathStore(host string) (bool, map[string][]st
 	return true, obj.(map[string][]string)
 }
 
+func (h *HostNamePathStore) GetHostsFromHostPathStore(host, fqdnMatchType string) []string {
+	allHosts := h.hostNamePathStore.GetAllKeys()
+	returnHosts := []string{}
+
+	for _, mHost := range allHosts {
+		if fqdnMatchType == string(akov1alpha1.Exact) && mHost == host {
+			returnHosts = append(returnHosts, mHost)
+		} else if fqdnMatchType == string(akov1alpha1.Contains) && strings.Contains(host, mHost) {
+			returnHosts = append(returnHosts, mHost)
+		} else if fqdnMatchType == string(akov1alpha1.Wildcard) && strings.HasPrefix(host, "*") {
+			wildcardFqdn := strings.Split(host, "*")[1]
+			if strings.HasSuffix(mHost, wildcardFqdn) {
+				returnHosts = append(returnHosts, mHost)
+			}
+		}
+	}
+	return returnHosts
+}
+
 func (h *HostNamePathStore) GetHostPathStoreIngresses(host, path string) (bool, []string) {
 	ok, obj := h.hostNamePathStore.Get(host)
 	if !ok {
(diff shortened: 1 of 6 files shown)


chauhanshubham:
build ako

chauhanshubham:
build ako

internal/objects/crdrules.go
```diff
@@ -178,7 +178,7 @@ func (c *CRDLister) UpdateFQDNHostruleMapping(fqdn string, hostrule string) {
 func (c *CRDLister) GetFQDNFQDNTypeMapping(fqdn string) (bool, string) {
 	found, fqdnType := c.FqdnFqdnTypeCache.Get(fqdn)
 	if !found {
-		return false, ""
+		return true, string(akov1alpha1.Exact)
```
monotosh-avi, line 181: If we are always returning true, then probably we dont need to return it.

chauhanshubham: makes sense. done thanks

internal/nodes/ingress_model_rel.go
```diff
@@ -540,56 +541,69 @@ func parseServicesForRoute(routeSpec routev1.RouteSpec, key string) []string {
...
-		utils.AviLog.Debugf("key: %s, msg: Couldn't find hostpath info for host: %s in cache", key, fqdn)
-	} else {
-		for _, ingresses := range obj {
-			for _, ing := range ingresses {
-				if !utils.HasElem(allIngresses, ing) {
-					allIngresses = append(allIngresses, ing)
+	// in case the hostname is updated, we need to find ingresses for the old ones as well to recompute
+	if oldFound {
+		allOldFqdns := SharedHostNameLister().GetHostsFromHostPathStore(oldFqdn, oldFqdnType)
+		for _, i := range allOldFqdns {
```
akshayhavile, line 583: should we use 'fqdn'/ anyother name instead of `i`?

internal/nodes/ingress_model_rel.go
```diff
@@ -540,56 +541,69 @@ func parseServicesForRoute(routeSpec routev1.RouteSpec, key string) []string {
...
 			}
 		}
 	}
 
-	// in case the hostname is updated, we need to find ingresses for the old ones as well to recompute
-	if oldFound {
-		ok, oldobj := SharedHostNameLister().GetHostPathStore(oldFqdn)
+	// find ingresses with host==fqdn, across all namespaces
+	allFqdns := SharedHostNameLister().GetHostsFromHostPathStore(fqdn, fqdnType)
+	for _, i := range allFqdns {
```
akshayhavile, line 601: Should we use `fqdn` or anyother name instead of `i`?

internal/nodes/hostname_cache.go
```diff
@@ -98,6 +99,25 @@ func (h *HostNamePathStore) GetHostPathStore(host string) (bool, map[string][]st
...
+	returnHosts := []string{}
+
+	for _, mHost := range allHosts {
+		if fqdnMatchType == string(akov1alpha1.Exact) && mHost == host {
+			returnHosts = append(returnHosts, mHost)
+		} else if fqdnMatchType == string(akov1alpha1.Contains) && strings.Contains(host, mHost) {
+			returnHosts = append(returnHosts, mHost)
+		} else if fqdnMatchType == string(akov1alpha1.Wildcard) && strings.HasPrefix(host, "*") {
+			wildcardFqdn := strings.Split(host, "*")[1]
+			if strings.HasSuffix(host, wildcardFqdn) {
```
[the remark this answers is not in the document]
chauhanshubham: no, this is fine actually. `GetHostsFromHostPathStore` gets called from `HostRuleToIng` and the `host` here comes from the hostrule. the hostrule would have the wildcard fqdn defined in it.

internal/nodes/ingress_model_rel.go
chauhanshubham: the singular version of `allOldFqdns` -> `oldFqdn` was already taken. I thought adding another fqdn like thing might just cause confusion 😅. There are too many `*fqdn*` variables already.

internal/nodes/hostname_cache.go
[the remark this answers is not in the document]
chauhanshubham: got it, was looking at HasPrefix earlier. great catch. Thanks

chauhanshubham:
build ako

monotosh-avi (approved):

akshayhavile (approved):

PR #758, opened 2022-01-28
VCF: Deconfiguring infra configurations on namespace delete

chauhanshubham:
This commit adds deconfiguration logic when a namespace is deleted
on the supervisor cluster. The logic covers deletion of SEs and
the SEG corresponding to the supervisor cluster. This workflow
is triggered when there are no NSXNetworkConfig CRs left in the
cluster.
We also introduce the use of finalizers in the Ingresses synced via
AKO. The finalizer is added on the Ingress when there is atleast 1
host/IP status populated, and removed when no host/IP status is present.
This allows for clean deletion of Ingresses, only after VS/vip deletion
in Avi.


ako-infra/avirest/handle_deconfig.go
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright 2021 VMware, Inc.
```
monotosh-avi, line 2: the year needs to be changed to 2022

ako-infra/avirest/handle_deconfig.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+
+	if len(response.Results) == 0 {
+		utils.AviLog.Infof("No service engine groups found in cloud %s", cloudName)
+		return nil
+	}
+
+	segUuid := *response.Results[0].UUID
+	deleteSEGUri := "/api/serviceenginegroup/" + segUuid
```
monotosh-avi, line 104: Instead of deleting the first SE from the list should we query by SEGroup Name and delete the specific SEGroup ?

internal/lib/finalizer_utils.go
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2020-2021 VMware, Inc.
```
monotosh-avi, line 2: Year has to be updated.

ako-infra/avirest/handle_deconfig.go
chauhanshubham: I am fetching the SEs using the se group name reference in the GET call above.

monotosh-avi: missed it, thanks

chauhanshubham:
build ako

ako-infra/avirest/handle_deconfig.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"github.com/vmware/alb-sdk/go/models"
+
+	avicache "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/cache"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
+	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/third_party/github.com/vmware/alb-sdk/go/session"
+)
+
+func DeleteServiceEngines() error {
```
akshayhavile, line 26: should we introduce paging mechanism ?

ako-infra/avirest/handle_deconfig.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return nil
+}
+
+func DeleteServiceEngineGroup() error {
+	client := avicache.SharedAVIClients().AviClient[0]
+	segName := lib.GetClusterID()
+	cloudName := utils.CloudName
+
+	segListUri := "/api/serviceenginegroup/?name=" + segName + "cloud_ref.name=" + cloudName
```
akshayhavile, line 77: @chauhanshubham:  We are missing `&` between segName  and "cloud_ref.name=".  Similar changes are required in function `checkSeGroup()`, we are forming `uri` wrong.

internal/lib/avi_api.go
```diff
@@ -154,6 +154,31 @@ func AviPost(client *clients.AviClient, uri string, payload interface{}, respons
...
+			return err
+		}
+	}
+
+	err := client.AviSession.Delete(uri)
+	if err != nil {
+		utils.AviLog.Warnf("msg: Unable to execute Delete on uri %s %v", uri, err)
+		checkForInvalidCredentials(uri, err)
+		apimodels.RestStatus.UpdateAviApiRestStatus("", err)
+		if aviError, ok := err.(session.AviError); ok && aviError.HttpStatusCode == 403 {
```
akshayhavile, line 172: Should switch context for sE and SEG (if they are in provider context) instead of returning err?

ako-infra/avirest/handle_deconfig.go
chauhanshubham: done

monotosh-avi (approved):

akshayhavile (approved):

PR #771, opened 2022-02-09
Fix to configure appProfile, httpPolicies, datascripts for EVH/SNI parent VSes

chauhanshubham:
This commit allows AKO to configure these additional policies in the
SNI/EVH parent VSes, which were missed before. All other policies are
being set in the parent VS fine.

### DIFF
--- a/internal/rest/avi_evh_obj.go
+++ b/internal/rest/avi_evh_obj.go
@@ -277,6 +277,11 @@ func (rest *RestOperations) AviVsBuildForEvh(vs_meta *nodes.AviEvhVsNode, rest_m
 		// This is EVH Parent or dedicated VS
 		network_prof := "/api/networkprofile/?name=" + vs_meta.NetworkProfile
 		app_prof := "/api/applicationprofile/?name=" + vs_meta.ApplicationProfile
+		if vs_meta.AppProfileRef != "" {
+			// hostrule ref overrides defaults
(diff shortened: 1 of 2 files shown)


chauhanshubham:
Some configurations were missing to be programmed in the rest layer for parent VSes. Let me know if this is a potential 1.6.2 candidate too @akshayhavile @monotosh-avi

internal/rest/avi_evh_obj.go
```diff
@@ -277,6 +277,11 @@ func (rest *RestOperations) AviVsBuildForEvh(vs_meta *nodes.AviEvhVsNode, rest_m
 		// This is EVH Parent or dedicated VS
 		network_prof := "/api/networkprofile/?name=" + vs_meta.NetworkProfile
 		app_prof := "/api/applicationprofile/?name=" + vs_meta.ApplicationProfile
+		if vs_meta.AppProfileRef != "" {
+			// hostrule ref overrides defaults
+			app_prof = vs_meta.AppProfileRef
```
akshayhavile, line 282: should we cast it with proto.String()?

akshayhavile (commented):
@chauhanshubham, @monotosh-avi : I feel, we should include it as part of 1.6.2 as we are trying suggest mTLS solution to customer for EVH.

akshayhavile (approved):

monotosh-avi (approved):

PR #774, opened 2022-02-11
Fix to configure appProfile, httpPolicies, datascripts for EVH/SNI parent VSes

chauhanshubham:
This commit allows AKO to configure these additional policies in the
SNI/EVH parent VSes, which were missed before. All other policies are
being set in the parent VS fine.


monotosh-avi (approved):

akshayhavile (approved):

akshayhavile:
build ako

PR #777, opened 2022-02-14
Running multiple AKO instances in cluster

akshayhavile:
This PR introduces changes for running multiple AKO instances per cluster. (Not contain changes for stretched cluster).
Use case is to improve performance of AKO for handling scale setup.
Customer can run multiple AKO with helm command as: 
`helm install ./ako --generate-name -n avi-system --set AKOSettings.namespaceSelector.labelKey="app" --set AKOSettings.namespaceSelector.labelValue="migrate" --set AKOSettings.primaryInstance=true`

value for `AKOSettings.primaryInstance` is `true/false`
Points to consider:
1. Each AKO instance needs to be installed in different namespace.
2. For non-primary AKO, shared Vs naming convention will be `Shared-VS-Name`= `<cluster-name>--<AKO-namespace>-Shared-L7-<Shard number>`
For Primary AKO, no naming convention changes.
3. User name (with which object created at AVI side) for non-primary will be `Username` = `ako-<cluster-name>-<AKO-namespace>`

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -286,30 +286,18 @@ func GetLayer7Only() bool {
 }
 
 var AKOUser string
-var AKOID string
 
 func SetAKOUser() {
 	AKOUser = "ako-" + GetClusterName()
-	if GetAKOID() != "1" {
-		AKOUser = AKOUser + "-" + AKOID
+	isPrimaryAKO := akoControlConfigInstance.GetAKOInstanceFlag()
+	if !isPrimaryAKO {
+		AKOUser = AKOUser + "-" + os.Getenv("POD_NAMESPACE")
 	}
 	utils.AviLog.Infof("Setting AKOUser: %s for Avi Objects", AKOUser)
 }
 
 func GetAKOUser() string {
 	return AKOUser
-}
-
-func SetAKOID() {
-	AKOID = os.Getenv("AKO_ID")
-	if AKOID == "" {
-		utils.AviLog.Warn("Setting AKO ID to 1")
-		AKOID = "1"
-	}
-}
-
-func GetAKOID() string {
-	return AKOID
 }
 
 var enableCtrl2014Features bool
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -286,14 +286,30 @@ func GetLayer7Only() bool {
 }
 
 var AKOUser string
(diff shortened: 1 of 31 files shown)


akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

tests/evhtests/l7_evh_crd_test.go
```diff
@@ -80,6 +80,8 @@ func TestMain(m *testing.M) {
 	os.Setenv("POD_NAMESPACE", utils.AKO_DEFAULT_NS)
 	os.Setenv("SHARD_VS_SIZE", "LARGE")
 	os.Setenv("AUTO_L4_FQDN", "default")
+	os.Setenv("AKO_ID", "1")
+	lib.SetAKOID()
```
monotosh-avi, line 84: Can we call SetAKOID with the AKO_ID as argument in stead of setting env variable ?

internal/lib/lib.go
```diff
@@ -285,16 +285,32 @@ func GetLayer7Only() bool {
 }
 
 var AKOUser string
+var AKOID string
```
chauhanshubham, line 288: could you please move this global variable in `akoControlConfig`. Let's try moving all the global variables to the common struct.

chauhanshubham:
@akshayhavile If the AKO ID is assumed to be an integer always, can we add some validation that does not allow random strings to be passed on by users?
Also the `== 1` and `!= 1` checks at first glance look very rigid, I'm thinking we replace AKOID with a simple boolean that indicates `primary AKO` and `secondary AKOs`

akshayhavile:
build ako

akshayhavile:
build ako

chauhanshubham:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

monotosh-avi (approved):

chauhanshubham (approved):

PR #788, opened 2022-02-25
Remove overwriting controllerVersion on POST/PUT/DELETE api calls

chauhanshubham:
This commit removes the controller version that is being overwritten
on individual API calls to the controller instead of using the api
version which was used to initialise the session to the controller.
With these changes, all REST layer API calls end up using the version
which was used while initializing the session.


chauhanshubham:
build ako

akshayhavile (approved):

monotosh-avi (approved):

Issue #783, opened 2022-02-23
Got security vulnerabilities alert from github when import github.com/vmware/load-balancer-and-ingress-services-for-kubernetes


XudongLiuHarold:
### Describe the bug

I got `We found potential security vulnerabilities in your dependencies` alert when I import `github.com/vmware/load-balancer-and-ingress-services-for-kubernetes` in my go project.

![Imgur](https://i.imgur.com/JQitr7Q.png)

### Reproduction steps

```bash
run `go get github.com/vmware/load-balancer-and-ingress-services-for-kubernetes` in your repo.
```

### Expected behavior

No security vulnerabilities after run `go get github.com/vmware/load-balancer-and-ingress-services-for-kubernetes`

akshayhavile:
@XudongLiuHarold: This has been fixed in recent release. So  I am closing this.

PR #794, opened 2022-03-05
Remove stale references from go.sum and go module files

akshayhavile:
Addresses issue:  #783 
Steps:
1. Add a replace directive using `go mod edit -replace old=new@version` command to replace all old stale references with exact version references required.  or you can edit go.mod file directly and replace a module with specific version of module. 
2. Then run `go mod tidy`
3. `go mod vendor`
This will replace all references in go.sum and vendor file with specific @version mentioned in replace directive. 

TODO: Replace older references with latest stable version.


akshayhavile:
build ako

chauhanshubham (approved):

monotosh-avi (approved):

PR #813, opened 2022-03-23
AV-141276: Allow AKO to boot when  `deleteConfig` flag is set to true

akshayhavile:
`Current Behaviour`: When AKO boots up with deleteConfig flag set to true, it shutdowns api server and goes into crashloopback off.  Due to this, AKO might not able to delete all objects if it is crashed before deleting all objects or timeout happens.

`New behaviour`: If AKO boots up with deleteConfig set to `true`, it should delete objects from Controller side and does not sync up objects from K8 side.  Required message will be set in log file as well as part of event framework for AKO pod. 

User has to set `deleteConfig` to `false` to start re-sync of k8 objects.


monotosh-avi (approved):
You may want to change the PR header to Allow AKO to boot when deleteConfig flag is set to true.

akshayhavile:
build ako

internal/rest/k8s_utils.go
```diff
@@ -147,6 +147,8 @@ func (rest *RestOperations) SyncObjectStatuses() {
 			status.UpdateL4LBStatus(allServiceLBUpdateOptions, true)
 		}
 	}
-	utils.AviLog.Infof("Status syncing completed")
-	lib.AKOControlConfig().PodEventf(v1.EventTypeNormal, lib.StatusSync, "Status syncing completed")
+	if !lib.DisableSync {
```
chauhanshubham, line 150: This seems to just control the log, but the sync is still happening. I think it's fine to log these since we are syncing statuses of already synced VSes in Avi

akshayhavile: addressed

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

swathinsankaran (approved):

chauhanshubham (approved):

PR #821, opened 2022-04-01
Convergence time optimizations

swathinsankaran:
This PR optimizes the AKO's convergence time.


apalsule:
@swathinsankaran Please add some description on what optimizations were done here.

swathinsankaran:
@chauhanshubham The condition check is added after the CRD validations. The new ako does the validation and throws an error if the validation fails.

internal/k8s/ako_init.go
```diff
@@ -703,39 +703,56 @@ func (c *AviController) FullSyncK8s() error {
...
-			/*
-				Key added to Ingestion queue if
-				1. Advance L4 enabled or
-				2. Namespace is valid
-			*/
-			if !utils.IsServiceNSValid(ns[0]) {
-				continue
+
+	acceptedNamespaces := make(map[string]struct{})
+	for _, nsObj := range nsObjs {
```
akshayhavile, line 714: populateNamespaceList() already prepare a list of valid namespace. We can use that.

internal/k8s/crdcontroller.go
```diff
@@ -794,6 +794,11 @@ func validateHostRuleObj(key string, hostrule *akov1alpha1.HostRule) error {
 		return err
 	}
 
+	// During ako reboot the hostrule is validated by the old AKO, hence returning.
```
chauhanshubham, line 797: could you change the comment to better reflect updated the workflow - something like "no need to update status to accepted if it was accepted before".
Also, could you please add this logic for AviInfraSetting CRD too.

internal/k8s/ako_init.go
swathinsankaran: I have added a function to get all valid namespaces present in the filter.

internal/k8s/ako_init.go
```diff
@@ -703,39 +703,56 @@ func (c *AviController) FullSyncK8s() error {
...
+			var key string
+			svcLabel := utils.ObjKey(svcObj)
+			ns := strings.Split(svcLabel, "/")
+			if isSvcLb && !lib.GetLayer7Only() {
+				/*
+					Key added to Ingestion queue if
+					1. Advance L4 enabled or
+					2. Namespace is valid
+				*/
+				if !utils.IsServiceNSValid(ns[0]) {
```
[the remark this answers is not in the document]
swathinsankaran: yes, you're right. The function always returns true. I have removed the condition check.

chauhanshubham (approved):

akshayhavile (approved):

PR #823, opened 2022-04-11
Fixes for ako-infra in WCP with NSX-t

chauhanshubham:
Set controller version in ako-infra from fetching it from the controller.
Fix avi client usage in ako-infra, before this the container was using clients
from the AKO initialized avi shared clients.
Do not try configuring ls-lr mappings in the cloud (PUT calls) when no mappings are
found in the CRs.
Correct the usable network configuration in IPAM profile for nsxt cloud, use
UsableNetworks instead of UsableNetworkRefs (deprecated)
Allow for both create and update of serviceenginegroup while ako-infra boots up. Earlier
the infra container was programmed to just create SEG - leading to errors related to
repetetive Post API calls on every bootup.
While creating SEGroups, configure markers and NOT labels, in order for AKO to find
the relevant SE group for syncing VSes.
In nsx-t cloud's NsxtConfiguration, configure the transport zone properly in
ManagementNetworkConfig.
Earlier the infra-container used a deprecated transport zone field in the
cloud configuration.
Fix incorrect copying of secret from NCP secret to AKO secret.
Add clusterroles for namespace (to allow updating AKO namespace with annotations)
and secrets (for creating and updating AKO secret) for AKO in WCP.
Sync updated status schema of akobootstrapconditions CRD, to fetch transportZone path
details properly.
Misc cleanup for admin switch context API retries.


chauhanshubham:
Closing this and opening a separate one. https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/848

PR #824, opened 2022-04-13
Disable finalizers in non-VCF clusters

chauhanshubham:


swathinsankaran (approved):

internal/status/ing_status.go
```diff
@@ -360,18 +363,25 @@ func deleteObject(option UpdateOptions, key string, isVSDelete bool, retryNum ..
...
 			}
 		}
 	}
 
-	if len(mIngress.Status.LoadBalancer.Ingress) > 0 && mIngress.Status.LoadBalancer.Ingress[0].IP != "" {
-		lib.CheckAndSetIngressFinalizer(mIngress)
-	} else {
-		lib.RemoveIngressFinalizer(mIngress)
+	// Add/Remove Ingress finalizers only when operating in a VCF Cluster.
+	if utils.IsVCFCluster() {
```
akshayhavile, line 379: @chauhanshubham : I think we need to add `if vcfcluster` condition over here also: https://github.com/chauhanshubham/load-balancer-and-ingress-services-for-kubernetes/blob/finalizer-nonvcf/internal/nodes/route_ingress_model.go#L173

akshayhavile (approved):

PR #825, opened 2022-04-15
AV-142604: VSes are not getting created for Openshift cluster for Passthrough Redirect route with Multiple AKO

akshayhavile:
Issue was due to name convention for shared passthrough VS.
Passthrough VSes were getting created with same name in multiple AKO.
This PR addresses that.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -1337,11 +1337,19 @@ func PassthroughShardSize() uint32 {
 	}
 	return 1
 }
-
+func GetAKOID_Prefix() string {
+	var akoID string
+	isPrimaryAKO := AKOControlConfig().GetAKOInstanceFlag()
+	if !isPrimaryAKO {
+		akoID = os.Getenv("POD_NAMESPACE") + "-"
+	}
+	return akoID
+}
 func GetPassthroughShardVSName(s string, key string) string {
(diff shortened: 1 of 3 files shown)


akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

internal/lib/lib.go
```diff
@@ -1337,11 +1337,19 @@ func PassthroughShardSize() uint32 {
 	}
 	return 1
 }
-
+func GetAKOID_Prefix() string {
```
swathinsankaran, line 1340: please consider removing the `_` from the function name.

chauhanshubham, line 1340: Nit: Can we not use the underscore in function name, also lets reuse this same function in GetShardVSPrefix too.

akshayhavile: Addressed

akshayhavile: addressed

akshayhavile:
Failed due to: /dev/stdout: resource temporarily unavailable... Re-running

akshayhavile:
build ako

chauhanshubham (approved):

swathinsankaran (approved):

PR #826, opened 2022-04-18
Unblock next EVH child if current EVH child fails with error that does not impact next child

akshayhavile:
As part of PR changes, in EVH deployment, AKO processes next EVH child if current child does not fail with error

1. Error status code is between 500 to 599
2. Error status code is 408 which indicates there is an issue with controller session timeout
3. Error status code is 400 with error `no free IP available`
4. Error status code is 403 with error `configuration not allowed during controller upgrade`
5. Error status code is 401 with invalid credential error

### DIFF
--- a/internal/rest/avi_evh_obj.go
+++ b/internal/rest/avi_evh_obj.go
@@ -140,9 +140,9 @@ func (rest *RestOperations) RestOperationForEvh(vsName string, namespace string,
 		} else {
 			_, evh_rest_ops = rest.EvhNodeCU(evhNode, nil, namespace, sni_to_delete, evh_rest_ops, key)
 		}
-		if success, next_child := rest.ExecuteRestAndPopulateCache(evh_rest_ops, vsKey, avimodel, key, true); !success {
-			if !next_child {
-				utils.AviLog.Debugf("key: %s, msg: Failure in processing EVH node: %s. Not processing other child nodes.", key, evhNode.Name)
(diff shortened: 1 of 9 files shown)


internal/rest/avi_evh_obj.go
```diff
@@ -140,8 +140,11 @@ func (rest *RestOperations) RestOperationForEvh(vsName string, namespace string,
 		} else {
 			_, evh_rest_ops = rest.EvhNodeCU(evhNode, nil, namespace, sni_to_delete, evh_rest_ops, key)
 		}
-		if success := rest.ExecuteRestAndPopulateCache(evh_rest_ops, vsKey, avimodel, key, true); !success {
-			return
+		if success, next_child := rest.ExecuteRestAndPopulateCache(evh_rest_ops, vsKey, avimodel, key, true); !success {
+			if !next_child {
+				utils.AviLog.Debugf("key: %s, msg: Failure in processing EVH node: %s. Not processing other child nodes.", key, evhNode.Name)
```
chauhanshubham, line 145: I think this might be a good `Infof` candidate instead of Debug

chauhanshubham (dismissed):
looks good to me, minor nits

swathinsankaran (dismissed):

akshayhavile:
build ako

swathinsankaran (approved):

chauhanshubham (approved):

PR #829, opened 2022-04-19
Fix for `AKO VS creation fails when using HostRule with customPorts, aliases`

swathinsankaran:
This PR contains the following two changes.
1. Removes the hostname reference from HTTP policy
2. Modifies the datascript to remove the port and then
   route the traffic([reference](https://github.com/vmware/nsx-alb-datascript-samples-library/blob/master/availability/location_header_rewrite_with_nonstandard_port.md)).

TODO: FT fixes.


swathinsankaran:
build ako

internal/nodes/avi_model_nodes.go
```diff
@@ -1017,9 +1017,14 @@ func (v *AviHttpPolicySetNode) CalculateCheckSum() {
 	var checksum uint32
 	for _, hpp := range v.HppMap {
 		sort.Strings(hpp.Path)
-		sort.Strings(hpp.Host)
-		checksum = checksum + utils.Hash(utils.Stringify(hpp))
-
+		checksum += utils.Hash(utils.Stringify(hpp.Name)) +
```
chauhanshubham, line 1020: you can copy hpp into a new variable and remove Host from that.

swathinsankaran: I have made changes as per the comment.

swathinsankaran:
build ako

chauhanshubham (dismissed):

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

internal/nodes/avi_model_nodes.go
```diff
@@ -1077,6 +1067,38 @@ type AviHostPathPortPoolPG struct {
...
+func (v *AviHostPathPortPoolPG) GetCheckSum() uint32 {
+	// Calculate checksum and return
+	v.CalculateCheckSum()
+	return v.Checksum
+}
+
+func (v *AviHostPathPortPoolPG) CalculateCheckSum() {
+	var checksum uint32
+	sort.Strings(v.Path)
+	sort.Strings(v.Host)
```
akshayhavile, line 1079: @swathinsankaran: `Host` are marked nil. I think we can skip sorting it.

internal/nodes/avi_model_nodes.go
```diff
@@ -1077,6 +1067,38 @@ type AviHostPathPortPoolPG struct {
...
+	checksum = checksum + utils.Hash(utils.Stringify(v))
+	v.Checksum = checksum
+}
+
+func (v *AviHostPathPortPoolPG) GetNodeType() string {
+	// Calculate checksum and return
+	return "AviHostPathPortPoolPG"
```
akshayhavile, line 1086: We can introduce constant here.

internal/nodes/avi_model_nodes.go
```diff
@@ -1077,6 +1067,38 @@ type AviHostPathPortPoolPG struct {
...
+func (v *AviHostPathPortPoolPG) CalculateCheckSum() {
+	var checksum uint32
+	sort.Strings(v.Path)
+	sort.Strings(v.Host)
+	checksum = checksum + utils.Hash(utils.Stringify(v))
+	v.Checksum = checksum
+}
+
+func (v *AviHostPathPortPoolPG) GetNodeType() string {
+	// Calculate checksum and return
```
akshayhavile, line 1085: Comment seems to be irrelevant to code in place.

internal/nodes/avi_model_nodes.go
```diff
@@ -1077,6 +1067,38 @@ type AviHostPathPortPoolPG struct {
...
+	return "AviHostPathPortPoolPG"
+}
+
+func (v *AviHostPathPortPoolPG) CopyNode() AviModelNode {
```
akshayhavile, line 1089: `AviHostPathPortPoolPG` is inner field of the `httpPolicyset` object.  will it be required? When will it be called?

internal/nodes/avi_model_nodes.go
```diff
@@ -1302,7 +1324,9 @@ func (v *AviHTTPDataScriptNode) GetCheckSum() uint32 {
 func (v *AviHTTPDataScriptNode) CalculateCheckSum() {
 	// A sum of fields for this VS.
 	checksum := lib.DSChecksum(v.PoolGroupRefs, nil, false)
-	checksum = utils.Hash(fmt.Sprint(checksum) + utils.HTTP_DS_SCRIPT_MODIFIED)
+	for _, pgRefName := range v.PoolGroupRefs {
```
akshayhavile, line 1327: I think DS will point to single PG. Do we need this for loop.?

internal/rest/avi_datascript.go
```diff
@@ -159,7 +159,9 @@ func (rest *RestOperations) AviDSCacheAdd(rest_op *utils.RestOp, vsKey avicache.
 
 		// Datascript should not have a checksum
 		checksum := lib.DSChecksum(ds_cache_obj.PoolGroups, nil, false)
-		checksum = utils.Hash(fmt.Sprint(checksum) + utils.HTTP_DS_SCRIPT_MODIFIED)
+		for _, pgName := range ds_cache_obj.PoolGroups {
```
akshayhavile, line 162: `DS` will point to single `PG`. Do we need it?

akshayhavile (commented):
@swathinsankaran: Have we tested this solution on EVH? Datascripts are only get attached into insecure VS in SNI mode.

internal/rest/avi_datascript.go
swathinsankaran: Some UTs of panicked when I accessed the PoolGroup with index 0.

internal/nodes/avi_model_nodes.go
swathinsankaran: yes, you are right. but some UTs panicked when I accessed it with the slice with 0th index.

internal/nodes/avi_model_nodes.go
swathinsankaran: This function is getting called from other places so we cant remove the string sort of hosts.

internal/nodes/avi_model_nodes.go
swathinsankaran: This function is no longer required/called hence I have removed it.

internal/nodes/avi_model_nodes.go
swathinsankaran: This function is no longer required/called hence I have removed it.

swathinsankaran:
@akshayhavile, yes I have tested the http policy changes in both sni and evh. DS changes in sni mode. Also the upgrade scenarios as suggested by Shubham.

swathinsankaran:
build ako

internal/nodes/avi_model_nodes.go
akshayhavile: For receiver type `*AviHostPathPortPoolPG`, this will not create a problem as for each object of type `AviHostPathPortPoolPG` , AKO path should be nil as per new code changes.

internal/nodes/avi_model_nodes.go
akshayhavile: I think we should look into panic trace into depth. It could be somewhere we are missing setting up things properly.

swathinsankaran: it is failing at the place where we are calculating the checksum of DS after dissociation of PG from DS. I will add a condition check whether the number of PGs associated is 1, if yes, include the DS.

swathinsankaran:
build ako

chauhanshubham:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

chauhanshubham:
build ako

chauhanshubham (approved):

akshayhavile (approved):

PR #831, opened 2022-04-27
[Bug-fix] Fixes the VS delete issue after AKO reboot

swathinsankaran:
This PR fixes the issue of VSes getting deleted after the AKO reboot.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -717,7 +717,17 @@ func (c *AviController) FullSyncK8s() error {
 		}
 	}
 
-	acceptedNamespaces := utils.GetAllNamespacesInFilter()
+	acceptedNamespaces := make(map[string]struct{})
+	allNamespaces, err := utils.GetInformers().ClientSet.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		utils.AviLog.Errorf("Error in getting all namespaces: %v", err.Error())
+		return err
+	}
+	for _, ns := range allNamespaces.Items {
+		if utils.CheckIfNamespaceAccepted(ns.GetName(), ns.GetLabels(), false) {
+			acceptedNamespaces[ns.GetName()] = struct{}{}
+		}
+	}
 
 	for namespace := range acceptedNamespaces {
 		svcObjs, err := utils.GetInformers().ServiceInformer.Lister().Services(namespace).List(labels.Set(nil).AsSelector())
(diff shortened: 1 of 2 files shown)


swathinsankaran:
build ako

swathinsankaran:
build ako

internal/k8s/ako_init.go
```diff
@@ -717,7 +717,17 @@ func (c *AviController) FullSyncK8s() error {
...
+		utils.AviLog.Errorf("Error in getting all namespaces: %v", err.Error())
+		return err
+	}
+	for _, ns := range allNamespaces.Items {
+		if utils.CheckIfNamespaceAccepted(ns.GetName(), ns.GetLabels(), false) {
+			acceptedNamespaces[ns.GetName()] = struct{}{}
+		}
+	}
 
 	for namespace := range acceptedNamespaces {
```
chauhanshubham, line 732: on L885 in this file we seem to be again iterating through the acceptedNamespaces. This is particularly happening for Ingress objects. Doesn't seem right.

swathinsankaran: yes, What we are doing here is getting a list of accepted Namespaces (NSes associated with that AKO) and later listing the service, routes, and ingresses separately using their respective informers(in L885). The point to note here is that we are listing the objects that are part of the accepted namespaces only.

akshayhavile (approved):
LGTM

internal/k8s/ako_init.go
chauhanshubham: oh got it yes, misread because of the change in indentation down on L885. These are separate and not nested. So the iteration seems to be happening for only service and ingress objects, whereas for other objects (like pods, routes, CRDs, gateways) we are still fetching from all namespaces. Should we take this up later then?

chauhanshubham: Also lets say that between reboots (or better, upgrade) the accepted namespaces changed 
from **before: ns1, ns2**
to **after: ns2, ns3**

This would mean that post reboot we would have to remove objects from ns1. Will this be taken care of? cc: @akshayhavile

chauhanshubham (approved):

internal/k8s/ako_init.go
swathinsankaran: objects in the ns1 will be marked for deletion and objects in the ns3 will get added.
As discussed the optimization for pods will be done in the subsequent PR.
cc: @akshayhavile

PR #833, opened 2022-04-28
Fix pod panic while deleting ingress statuses

chauhanshubham:
Fixes slice bounds out of range error while deleting ingress statuses.
Fixes AV-143532

### DIFF
--- a/internal/status/ing_status.go
+++ b/internal/status/ing_status.go
@@ -359,8 +359,8 @@ func deleteObject(option UpdateOptions, key string, isVSDelete bool, retryNum ..
 	}
 
 	for _, host := range option.ServiceMetadata.HostNames {
-		for i, status := range mIngress.Status.LoadBalancer.Ingress {
-			if status.Hostname != host {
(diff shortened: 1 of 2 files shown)


internal/status/ing_status.go
```diff
@@ -359,8 +359,9 @@ func deleteObject(option UpdateOptions, key string, isVSDelete bool, retryNum ..
 	}
 
 	for _, host := range option.ServiceMetadata.HostNames {
-		for i, status := range mIngress.Status.LoadBalancer.Ingress {
-			if status.Hostname != host {
+		for i := 0; i < len(mIngress.Status.LoadBalancer.Ingress); {
```
akshayhavile, line 362: @chauhanshubham This will create problem when in status, there is same host with different IP. I think we should iterate from length to zero index

chauhanshubham:
build ako

aaha97 (approved):

akshayhavile (approved):

PR #839, opened 2022-05-09
AV-139061 Added cloud_uuid to network api call

abhishekbsingh:
This PR adds a cloud_uuid query parameter in the /api/network call which will work in case of both Default-Cloud and NonDefault-Cloud.


chauhanshubham (approved):
LGTM, let's try to move all package scoped global variables into the ako control config struct instead moving forward.

PR #854, opened 2022-05-20
Docs for ServiceLB shared-VIP and hostrule alternateCertificate

chauhanshubham:


docs/shared_vip.md
```diff
@@ -0,0 +1,100 @@
...
+
+The expected status message should have the VIP matching the preferred static IP provided in the Service spec.
+```
+NAME                      TYPE           CLUSTER-IP       EXTERNAL-IP     PORT(S)        AGE
+sharedvip-avisvc-lb1      LoadBalancer   10.108.153.227   100.64.196.75   80:31658/TCP   6d23h
+sharedvip-avisvc-lb2      LoadBalancer   10.102.147.29    100.64.196.75   80:31331/UDP   6d23h
+```
+
+
+There are a few things that must be considered while configuring the Services with the aforementioned annotation;
```
akshayhavile, line 97: Instead of `;`, it should be `:`

docs/shared_vip.md
```diff
@@ -0,0 +1,100 @@
...
+1. Make sure that LoadBalancer Services which are intended to share a VIP, must have the **same** annotation value. As shown in the example shown above, the annotation value `shared-vip-key-1` is same for both Services.
```
akshayhavile, line 98: `As shown in the example shown above` --> `As shown in the above example`

docs/shared_vip.md
```diff
@@ -0,0 +1,100 @@
+AKO provides support for sharing VIP among multiple kubernetes Services of type LoadBalancer. Generally, with LoadBalancer Services, AKO creates dedicated L4 Virtual Services in the Avi Controller, but multiple LoadBalancer Services can also be combined to share a single VIP.
+
+> **Note**: The ability to share a single VIP among multiple LoadBalancer Services is released as tech-preview as part of AKO release-1.7.1
+
+This can be achieved by providing an annotation to multiple LoadBalancer Services, where VIP sharing is intended.
+The annotation to be applied is `ako.vmware.com/enable-shared-vip` with a string value. An example of this is shown below.
```
akshayhavile, line 6: We can combine this two sentences `The annotation to be applied is `ako.vmware.com/enable-shared-vip` with a string value. An example of this is shown below.` as `The annotation to be applied is `ako.vmware.com/enable-shared-vip` with a string value as shown below`

docs/shared_vip.md
```diff
@@ -0,0 +1,100 @@
...
+sharedvip-avisvc-lb2      LoadBalancer   10.102.147.29    100.64.196.75   80:31331/UDP   6d23h
+```
+
+
+There are a few things that must be considered while configuring the Services with the aforementioned annotation;
+1. Make sure that LoadBalancer Services which are intended to share a VIP, must have the **same** annotation value. As shown in the example shown above, the annotation value `shared-vip-key-1` is same for both Services.
+2. In order to avoid any errors while configuring the Virtual Service on the Avi controller, it is required that there is no conflicting Port-Protocol pairs in the LB Services that share the Annotation value. With the example shown above both Services are exposing a unique, non-conflicting Port-Protocol for the backend application i.e. 80/TCP and 80/UDP. Explicit checks will be added around this to ensure misconfigurations in future AKO releases.
+3. The annotation must be provided only on Service of type LoadBalancers. In case the annotation is not provided on the LoadBalancer Service, a dedicated L4 Virtual Service is created corresponding to the Service.
```
akshayhavile, line 100: I think Second part of 3rd point is existing behaviour. So we should not mention it explicitly

chauhanshubham:
build ako

akshayhavile (approved):

apalsule (approved):

PR #861, opened 2022-05-29
adds regex check for RFC 1035

metalcyanide:


metalcyanide:
build ako

internal/nodes/avi_model_l4_translator.go
```diff
@@ -441,10 +441,17 @@ func getAutoFQDNForService(svcNamespace, svcName string) string {
 	if lib.GetL4FqdnFormat() == lib.AutoFQDNDefault {
 		// Generate the FQDN based on the logic: <svc_name>.<namespace>.<sub-domain>
 		fqdn = svcName + "." + svcNamespace + "." + subdomain
+
 	} else if lib.GetL4FqdnFormat() == lib.AutoFQDNFlat {
 		// Generate the FQDN based on the logic: <svc_name>-<namespace>.<sub-domain>
 		fqdn = svcName + "-" + svcNamespace + "." + subdomain
 	}
+	if !lib.CheckRFC1035(fqdn) {
```
chauhanshubham, line 449: The fqdn can be auto-generated from within AKO itself, which is not an explicit user input, in which case 
1. either this should not be enforced.
2. the auto-generated fqdn must be compliant at all times regardless of any user input.

metalcyanide: We need to follow the label standard RFC 1035. Either we can ask them to provide the svcName starting with an alphabetic character or we can add that tag "svc" when they don't. The other alternative is to always add the tag. Please suggest.

internal/nodes/avi_model_l4_translator.go
```diff
@@ -441,10 +441,17 @@ func getAutoFQDNForService(svcNamespace, svcName string) string {
...
+		if lib.CheckRFC1035("svc" + fqdn) {
+			fqdn = "svc" + fqdn
```
akshayhavile, line 451: We can add warn message over here to notify user that fqdn has been changed. After line number 451.

internal/nodes/avi_model_l4_translator.go
```diff
@@ -441,10 +441,17 @@ func getAutoFQDNForService(svcNamespace, svcName string) string {
...
+		}
+		utils.AviLog.Warnf("Provided fqdn: %s doesn't obey RFC 1035", fqdn)
```
akshayhavile, line 453: This message seems to be irrelevant as we are printing Warn message in  CheckRFC1035.

internal/lib/utils.go
```diff
@@ -200,3 +201,26 @@ func Difference(a, b []string) []string {
...
+
+func CheckRFC1035(name string) bool {
+	RFCpattern := "^[a-z]([-a-z0-9]*[a-z0-9])?$"
+	maxlength := 63
+
+	if CheckConstraintsForRFC(name, RFCpattern, maxlength) {
+		return true
+	}
+
+	utils.AviLog.Warnf("Given string doesn't follow RFC 1035 constraints")
```
akshayhavile, line 224: Warn message should be  `FQDN %s does not follow RFC 1035.`

metalcyanide:
build ako

metalcyanide:
build ako

internal/nodes/avi_model_l4_translator.go
swathinsankaran, line 451: Shall we have a hyphen between string "svc" and fqdn so that it will become more readable?

internal/lib/utils.go
```diff
@@ -200,3 +201,26 @@ func Difference(a, b []string) []string {
...
+
+	compliedRegex := regexp.MustCompile(pattern)
+	match := compliedRegex.Match([]byte(name))
+
+	return match
+}
+
+func CheckRFC1035(name string) bool {
+	RFCpattern := "^[a-z]([-a-z0-9]*[a-z0-9])?$"
+	maxlength := 63
```
swathinsankaran, line 218: is the max length 255 or 63?

internal/lib/utils.go
```diff
@@ -200,3 +201,26 @@ func Difference(a, b []string) []string {
 	}
 	return diff
 }
+
+func CheckConstraintsForRFC(name string, pattern string, maxlength int) bool {
+	if len(name) > maxlength {
+		utils.AviLog.Warnf("Given string is longer than expected. Maximum allowed length is: %d", maxlength)
```
swathinsankaran, line 207: do we need to return a false here?

metalcyanide:
build ako

metalcyanide:
build ako

swathinsankaran (approved):

internal/lib/utils.go
akshayhavile: @metalcyanide : You can change log statement as "String %s has length more than maximum allowed length %d"

metalcyanide:
build ako

metalcyanide:
build ako

swathinsankaran (approved):

akshayhavile (approved):

PR #865, opened 2022-06-15
Add infrasetting support with sharedvip using Service of type LB

chauhanshubham:
This commit adds support for using infrasetting CRD,
with Services of type LB, when sharing VIP using
the annotation. An infrasetting can be referenced from such
Services via an annotation, which must be provided on
all Services that intend to share a VIP.


akshayhavile (approved):
LGTM

swathinsankaran:
build ako

swathinsankaran (approved):

PR #871, opened 2022-06-24
Removes shortname of clustersets

swathinsankaran:
This PR removes the shortname of Clustersets which was conflicting with the shortname of ComponentStatues.


swathinsankaran:
unwanted commits are coming hence closing this PR.

PR #877, opened 2022-07-04
Fix for the issue`Predefined health monitors are not assigned to a pool`

swathinsankaran:
This PR adds the code to process the update of hostrule, httprule, and aviinfrasetting once after the ako update the status as Accepted.

This PR fixes issue #AV-143438

### DIFF
--- a/internal/k8s/crdcontroller.go
+++ b/internal/k8s/crdcontroller.go
@@ -69,19 +69,26 @@ func NewIstioCRDInformers(cs istiocrd.Interface) {
 	})
 }
 
+func isHostRuleUpdated(oldHostRule, newHostRule *akov1alpha1.HostRule) bool {
+	if oldHostRule.ResourceVersion == newHostRule.ResourceVersion {
+		return false
+	}
+
+	oldSpecHash := utils.Hash(utils.Stringify(oldHostRule.Spec) + oldHostRule.Status.Status)
+	newSpecHash := utils.Hash(utils.Stringify(newHostRule.Spec) + newHostRule.Status.Status)
+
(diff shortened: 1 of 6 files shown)


swathinsankaran:
build ako

akshayhavile (approved):

internal/k8s/crdcontroller.go
```diff
@@ -69,6 +69,22 @@ func NewIstioCRDInformers(cs istiocrd.Interface) {
 	})
 }
 
+func isHostRuleUpdated(oldHostRule, newHostRule *akov1alpha1.HostRule) bool {
+	if oldHostRule.ResourceVersion == newHostRule.ResourceVersion {
```
akshayhavile, line 73: @swathinsankaran: I forgot once scenario. I think check should be `if resource version are not same then return true`. Reason: With current logic, `status` code will not be hit until unless there is update to CRD object. As per customer logs, there were no updates to hostrule after timing issue happened. So in that scenario AKO will never do update to VS.

akshayhavile (changes requested):

internal/k8s/crdcontroller.go
swathinsankaran: We are doing extra handling when the AKO updates the status which should resolve this issue. Periodic processing of the CR is not required as we will end up in unnecessary checksum calculations every 30 seconds.

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

akshayhavile (approved):

apalsule (approved):

PR #882, opened 2022-07-14
Regression fix: Change L4 checksum as naming convention for L4 pool is changed in 1.7.1

akshayhavile:
AV-149730: This PR addresses following issue:

Naming convention of L4 pool is changed in 1.7.1.  `protocol` field has been added as part of pool name. But as checksum was not getting changed for `VirtualService` and `L4PolicySet`, new pools were not getting referenced in L4policyset and Virtual Service. Also as old pools were attached to VS and L4 Policyset, old pools also were not getting deleted.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -1091,6 +1091,14 @@ func PopulateL4VSNodeMarkers(namespace, serviceName string) utils.AviObjectMarke
 	return markers
 }
 
+func PopulateL4PolicysetMarkers(namespace, serviceName string, protocols string) utils.AviObjectMarkers {
+	var markers utils.AviObjectMarkers
+	markers.Namespace = namespace
+	markers.ServiceName = serviceName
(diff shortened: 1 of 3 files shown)


apalsule:
Please add jira ID to the PR.

swathinsankaran (approved):
LGTM

akshayhavile:
build ako

swathinsankaran:
build ako

apalsule (approved):

akshayhavile:
@apalsule : Testing:
1. Upgrade AKO from 1.6.x to 1.7.2-beta with L4 services: L4 services with new pool should be created without any error.
2. Applying shared VIP annotation to L4 service and check transition. Transition should not leave any stale object associated with dedicated L4 Vs.
3. Remove shared VIP annotation from L4 service and check transition. Transition should not leave any stale object associated with shared vip

PR #884, opened 2022-07-20
AV-149930 Fix named ports with Ingress

aaha97:
Fixes the servers not being added to pool for ingress with named ports

testing done:
on AKO with serviceType as NodePortLocal, created an ingress with named ports. 

result:
servers got added to pool

### DIFF
--- a/internal/nodes/avi_model_evh_nodes.go
+++ b/internal/nodes/avi_model_evh_nodes.go
@@ -30,7 +30,6 @@ import (
 	avimodels "github.com/vmware/alb-sdk/go/models"
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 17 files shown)


apalsule:
@aaha97 Can you put some description of what caused the bug and what was the fix?

akshayhavile (commented):
@aaha97 structure `PodsWithTargetPort` has TargetPort field which is int32. We need to change it also.
Then code using that will also under go changes.

aaha97:
> @aaha97 structure `PodsWithTargetPort` has TargetPort field which is int32. We need to change it also. Then code using that will also under go changes.

@akshayhavile this is intentional. PodsWithTargetPort is a map between integer value associated with named port and the pod to work with NPLAnnotations.

akshayhavile (approved):
LGTM.

swathinsankaran (approved):
LGTM

PR #888, opened 2022-07-22
Fix HttpRule: Remove mandatory check for DestCA and PkiProfile

akshayhavile:
This PR fixes mandatory check "to have destinationCA or pkiProfile in tls setting section for httprule". It also adds validation to have only pkiprofile or destinationCA in rule.

Testing Done:
1. Httprule without pkiprofile or  destinationCA -->Accepted
2. Httprule with pkiprofile --> Accepted
3. httprule with destinationCA --> Accepted
4. httprule with destinationCA and pkiProfile --> rule Rejected

### DIFF
--- a/internal/k8s/crdcontroller.go
+++ b/internal/k8s/crdcontroller.go
@@ -965,6 +965,14 @@ func checkRefOnController(key, refKey, refValue string) error {
 func validateHTTPRuleObj(key string, httprule *akov1alpha1.HTTPRule) error {
 	refData := make(map[string]string)
 	for _, path := range httprule.Spec.Paths {
+		if path.TLS.PKIProfile != "" && path.TLS.DestinationCA != "" {
+			//if both pkiProfile and destCA set, reject httprule
+			status.UpdateHTTPRuleStatus(key, httprule, status.UpdateCRDStatusOptions{
+				Status: lib.StatusRejected,
+				Error:  lib.HttpRulePkiAndDestCASetErr,
+			})
+			return fmt.Errorf("key: %s, msg: %s", key, lib.HttpRulePkiAndDestCASetErr)
+		}
 		refData[path.TLS.SSLProfile] = "SslProfile"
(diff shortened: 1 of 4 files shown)


internal/k8s/crdcontroller.go
```diff
@@ -964,7 +964,12 @@ func checkRefOnController(key, refKey, refValue string) error {
 // update internal CRD caches, and push relevant ingresses to ingestion
 func validateHTTPRuleObj(key string, httprule *akov1alpha1.HTTPRule) error {
 	refData := make(map[string]string)
+	var pkiDestCASet bool
 	for _, path := range httprule.Spec.Paths {
+		if path.TLS.PKIProfile != "" && path.TLS.DestinationCA != "" {
+			pkiDestCASet = true
```
swathinsankaran, line 970: shall we return the function from here itself?

internal/lib/constants.go
```diff
@@ -172,6 +172,7 @@ const (
 	IPAMProviderInfoblox                       = "IPAMDNS_TYPE_INFOBLOX"
 	IPAMProviderCustom                         = "IPAMDNS_TYPE_CUSTOM"
 	SharedVipServiceKey                        = "SharedVipService"
+	HttpRulePkiAndDestCASetErr                 = "PKIProfile and DestinationCA has been set in a HTTPRule. Only one of the field should be set."
```
swathinsankaran, line 175: in a httprule or in the httprule? which one is correct?

akshayhavile:
build ako

apalsule (dismissed):

internal/k8s/crdcontroller.go
akshayhavile: addressed

internal/lib/constants.go
akshayhavile: addressed

apalsule (approved):

swathinsankaran (approved):

PR #906, opened 2022-08-30
Support for kubernetes 1.24

akshayhavile:
AKO supports 4 major releases of Kubernetes. So changed min and max according to it. 

@apalsule @swathinsankaran @aaha97 @abhishekbsingh : We need to check from customer perspective impact of changing min K8 version. But I think we should keep only 4 major releases into picture to reduce wide variety of testing


akshayhavile:
build AKO

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

swathinsankaran (approved):
LGTM

aaha97 (approved):
lgtm

PR #907, opened 2022-09-01
AV-147609: HTTP Ingress with a single path gives a 470 error

akshayhavile:
**Issue** is present only in Insecure Ingress for SNI deployment. DataScript is being used to redirect traffic to respective poolgroup and pool-member. As DataScript is doing case insensitive search to select pool member using `priority label`,  pool with case sensitive name was not getting selected and that was causing 470 error.

**Fix** For insecure ingress pool, priority label has been updated to all lower case characters to match appropriate pool.
For existing deployment, it will cause `PUT` request on pool and pool groups who has case sensitive host name and path.
This will impact only insecure ingress deployment on SNI.

Other modes uses, http policyset to redirect traffic and no datascript comes into picture.

**Previous to fix**
image

image

**After fix ***
image

image

### DIFF
--- a/internal/nodes/avi_model_l7_hostname_shard.go
+++ b/internal/nodes/avi_model_l7_hostname_shard.go
@@ -305,7 +305,7 @@ func buildPoolNode(key, poolName, ingName, namespace, priorityLabel, hostname st
 		IngressName:   ingName,
 		PortName:      obj.PortName,
 		Tenant:        lib.GetTenant(),
-		PriorityLabel: priorityLabel,
+		PriorityLabel: strings.ToLower(priorityLabel),
 		Port:          obj.Port,
 		TargetPort:    obj.TargetPort,
 		ServiceMetadata: lib.ServiceMetadataObj{


akshayhavile:
build ako

akshayhavile:
build ako

swathinsankaran (dismissed):
LGTM

swathinsankaran:
build ako

swathinsankaran (approved):

apalsule (approved):
What does the customer have who is already in this state have to do for this ?
He needs to update the priority labels on all his existing pools ? 
Currently the customer has removed the tolower on the datascript, right, we will have to instruct him to add it back, and update the priority labels on all pool groups ?

akshayhavile:
> What does the customer have who is already in this state have to do for this ? He needs to update the priority labels on all his existing pools ? Currently the customer has removed the tolower on the datascript, right, we will have to instruct him to add it back, and update the priority labels on all pool groups ?

@apalsule : For existing pools and pool groups in customer environment will undergo update if labels are combination of lower and upper case letters. 
Yes. Customer who updated datascript to remove `lower`. We need to ask them to add it back.

PR #908, opened 2022-09-01
IngressClass optimisations

swathinsankaran:
This PR adds the optimization to process ingresses that belong to the ingressClasses owned by AKO.
It fixes AV-148368.

Optimizations added:
1. During AKO reboot, validates every ingress object and ignore it if it belongs to a non-ako ingressClass.
2. Ignores the handling of ingresses that are not part of AKO's ingressClass in Add and Delete ingress handler functions. To capture the transition from ingressClass non-avi-lb to avi-lb and vice versa, in the Update ingress handler function the same optimisation code is not added.

Tested scenarios and their results:
1. Ingress from non-avi lb to avi-lb ingressClass -> VS got created
2. Ingress from avi lb to non-avi lb ingressClass -> VS got deleted
3. Avi-lb not default and ingress dont have ingressClass -> ignored
4. Ingress with non-default avi lb ingressClass -> VS got created
5. ingress with default avi lb ingressClass     -> VS got created

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -929,6 +929,10 @@ func (c *AviController) SetupEventHandlers(k8sinfo K8sinformers) {
 				utils.AviLog.Debugf("key : %s, msg: same resource version returning", key)
 				return
 			}
+			// optimization to check if ingress belongs to ingressClass handled by AKO.
+			if !lib.ValidateIngressForClass(key, ingress) {
+				return
+			}
 			bkt := utils.Bkt(namespace, numWorkers)
 			c.workqueue[bkt].AddRateLimited(key)
 			utils.AviLog.Debugf("key: %s, msg: ADD", key)
@@ -958,6 +962,10 @@ func (c *AviController) SetupEventHandlers(k8sinfo K8sinformers) {
 				return
 			}
 
+			// optimization to check if ingress belongs to ingressClass handled by AKO.
+			if !lib.ValidateIngressForClass(key, ingress) {
+				return
+			}
 			objects.SharedResourceVerInstanceLister().Delete(key)
 			bkt := utils.Bkt(namespace, numWorkers)
 			c.workqueue[bkt].AddRateLimited(key)
--- a/tests/ingresstests/ingressclass_test.go
+++ b/tests/ingresstests/ingressclass_test.go
@@ -53,7 +53,9 @@ func TestWrongClassMappingInIngress(t *testing.T) {
 
 	SetUpTestForIngress(t, modelName)
 	integrationtest.RemoveDefaultIngressClass()
+	integrationtest.AddIngressClassWithName("xyz")
 	defer integrationtest.AddDefaultIngressClass()
+	defer integrationtest.RemoveIngressClassWithName("xyz")
 
 	integrationtest.SetupIngressClass(t, ingClassName, lib.AviIngressController, "")
 	ingressCreate := (integrationtest.FakeIngress{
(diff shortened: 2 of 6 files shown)


swathinsankaran:
build ako

tests/ingresstests/ingressclass_test.go
```diff
@@ -191,6 +193,70 @@ func TestDefaultIngressClassChange(t *testing.T) {
 	VerifyPoolDeletionFromVsNode(g, modelName)
 }
 
+func TestIngressWithNonAVILBIngressClass(t *testing.T) {
+	// create ingress with ingressClass avi-lb
```
akshayhavile, line 197: Comments are wrong. Here we are creating first with wrong ingressclass and then update with right one,

akshayhavile (commented):
LGTM with one comment
@swathinsankaran: Is it possible to add a test case where it test whether ingress enqueued to ingestion queue or not when new ingress added with wrong ingress class

tests/ingresstests/ingressclass_test.go
swathinsankaran: I have corrected the comment. Thanks

swathinsankaran:
@akshayhavile I think we can't directly check whether the ingress is enqueued or not Since the SyncFunction of the Graph layer is inside the AKO code and overriding it is not possible. But we can check whether the models are created or not which does the same thing imo.

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

akshayhavile (dismissed):

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

internal/k8s/controller.go
```diff
@@ -905,6 +909,10 @@ func (c *AviController) SetupEventHandlers(k8sinfo K8sinformers) {
 				return
 			}
 			key := utils.Ingress + "/" + utils.ObjKey(ingress)
+			// optimization to check if ingress belongs to ingressClass handled by AKO.
+			if !lib.ValidateIngressForClass(key, ingress) {
+				return
+			}
```
apalsule, line 915: dont you need something similar for the updatefunc below ?

apalsule: Also not doing anything for delete seems like an issue, how do you handle the case where an ingress that was being managed by AKO, someone then adds an ingress class later. In that case the conrresponding items should be deleted from Avi , right ?

swathinsankaran: The ingressClass validation is not added in the update function to handle the transition of ingress from non-avi managed to avi managed and vice versa.

swathinsankaran: With this optimization, AKO ignores the ingresses that are part of non-avi managed ingressClass.

swathinsankaran:
build ako

apalsule (approved):

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

apalsule (approved):

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

akshayhavile (approved):

akshayhavile:
build ako

swathinsankaran:
build ako

PR #909, opened 2022-09-02
AV-153526: Avoid going through the system-namespaces on openshift and k8

akshayhavile:
As part of this PR
1. Introduced changes to stop processing events from namespaces mentioned in values.yaml
2. Removed logic where AKO was listening on AKO deployed namespace only for secrets.

As part of this PR, we did not handle runtime update to `blackListedNamespaces` variable as it will overload AKO with fetching updates from all un-blocked namespaces in bulk. So if customer does update to `blackListedNamespaces` list, then AKO needs to be re-started.

Testing done:
1. K8 cluster: without providing any blocked namespaces and with blocked system namespaces.
2. Openshift cluster: Without providing any blocked namespace and with blocked system namespaces.
>> Result on Openshift (with 2 master and 3 worker nodes. more than 1k secrets present)
  a. Without blackListedNamespace: Took 2 to 4 min to complete existing object processing after AKO rebooted.
  b. With blackListedNamespaces: Took 12 seconds to complete existing object processing after AKO rebooted.

3. Creating routes on un-blocked namespaces and rebooting AKO.
4.  Boot up AKO without `blackListedNamespaces` in values.yaml to check backward compatibility or where configmap or values.yaml doesn't update automatically such as wcp.

### DIFF
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -222,7 +222,7 @@ spec:
             valueFrom:
               configMapKeyRef:
                 name: avi-k8s-config
-                key: blackListedNamespaces
+                key: blockedNamespaceList
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           livenessProbe:
--- a/helm/ako/templates/statefulset.yaml
+++ b/helm/ako/templates/statefulset.yaml
@@ -218,6 +218,11 @@ spec:
               configMapKeyRef:
                 name: avi-k8s-config
                 key: enableMCI
+          - name: BLOCKED_NS_LIST
+            valueFrom:
+              configMapKeyRef:
+                name: avi-k8s-config
+                key: blackListedNamespaces
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           livenessProbe:
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -29,6 +29,11 @@ AKOSettings:
   servicesAPI: false # Flag that enables AKO in services API mode: https://kubernetes-sigs.github.io/service-apis/. Currently implemented only for L4. This flag uses the upstream GA APIs which are not backward compatible 
                      # with the advancedL4 APIs which uses a fork and a version of v1alpha1pre1 
   vipPerNamespace: "false" # Enabling this flag would tell AKO to create Parent VS per Namespace in EVH mode
+  # This is the list of system namespaces from which AKO will not listen any Kubernetes or Openshift object event.
+  blackListedNamespaces: []
+  # blackListedNamespaces:
+  #   - kube-system
+  #   - kube-public
 
 ### This section outlines the network settings for virtualservices. 
 NetworkSettings:
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -30,8 +30,8 @@ AKOSettings:
                      # with the advancedL4 APIs which uses a fork and a version of v1alpha1pre1 
   vipPerNamespace: "false" # Enabling this flag would tell AKO to create Parent VS per Namespace in EVH mode
   # This is the list of system namespaces from which AKO will not listen any Kubernetes or Openshift object event.
-  blackListedNamespaces: []
-  # blackListedNamespaces:
+  blockedNamespaceList: []
+  # blockedNamespaceList  :
   #   - kube-system
   #   - kube-public
 
--- a/internal/lib/control_config.go
+++ b/internal/lib/control_config.go
@@ -16,6 +16,8 @@ package lib
 
 import (
 	"context"
+	"sort"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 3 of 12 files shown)


pkg/utils/utils.go
```diff
@@ -183,11 +183,7 @@ func instantiateInformers(kubeClient KubeClientIntf, registeredInformers []strin
 		case EndpointInformer:
 			informers.EpInformer = kubeInformerFactory.Core().V1().Endpoints()
 		case SecretInformer:
-			if akoNSBoundInformer {
```
swathinsankaran, line 186: akoNSBoundInformer,akoInformerFactory and akoNS can be removed from this function as they are not used.

akshayhavile: @swathinsankaran I have re-introduced this block again as in `advanceL4` deployment all this will be used. So i don't want to touch that functionality. So just removed `route` related stuff.

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

internal/lib/lib.go
```diff
@@ -1397,6 +1407,14 @@ func GetAKOIDPrefix() string {
 	}
 	return akoID
 }
+
+//TODO: Optimize
+func IsNamespaceBlackListed(namespace string) bool {
+	nsBlockedList := AKOControlConfig().GetAKOBlackListedNamespaces()
```
swathinsankaran, line 1413: I think it will be better to avoid calling GetAKOBlacklistedNamespaces() every time instead call the map created out from `BLOCKED_NS_LIST` that way we can avoid unmarshaling the list every time.

internal/lib/control_config.go
```diff
@@ -121,7 +131,22 @@ func (c *akoControlConfig) SetAKOInstanceFlag(flag bool) {
 func (c *akoControlConfig) GetAKOInstanceFlag() bool {
 	return c.primaryaAKO
 }
-
+func (c *akoControlConfig) SetAKOBlackListedNamespaces(nsList []string) {
+	sort.Strings(nsList)
+	val := strings.Join(nsList, ":")
+	cksum := utils.Hash(val)
+	if c.blockedNS.nsChecksum != cksum {
```
swathinsankaran, line 138: Are we planning to listen to configmap updates for this?

akshayhavile: For current release no. For AKO, it will become overwhelming to handle bulk updates of all objects when namespaces transitioned from blocked state to non-blocked state.

internal/lib/lib.go
akshayhavile: I did not get your comment about un-marshalling. My intention was not to expose variable outside its required scope so introduced this function and reuse value in other code part if require. As part of optimization, I can introduce local variable here and check if it is set or not.

swathinsankaran: sure, we can take those changes in the next release as an optimization.

akshayhavile:
build ako

swathinsankaran:
build ako

helm/ako/values.yaml
```diff
@@ -29,6 +29,11 @@ AKOSettings:
   servicesAPI: false # Flag that enables AKO in services API mode: https://kubernetes-sigs.github.io/service-apis/. Currently implemented only for L4. This flag uses the upstream GA APIs which are not backward compatible 
                      # with the advancedL4 APIs which uses a fork and a version of v1alpha1pre1 
   vipPerNamespace: "false" # Enabling this flag would tell AKO to create Parent VS per Namespace in EVH mode
+  # This is the list of system namespaces from which AKO will not listen any Kubernetes or Openshift object event.
+  blackListedNamespaces: []
```
apalsule, line 33: can this also be some kind of filter for excluded namespaces, rather than a list, say based on some annotations ?

akshayhavile: I did not find any system specific annotation in openshift or kubernetes that can be used here. Using namespace based filter (namespace sync feature) is another approach for newly created user namespaces.

helm/ako/templates/statefulset.yaml
```diff
@@ -218,6 +218,11 @@ spec:
               configMapKeyRef:
                 name: avi-k8s-config
                 key: enableMCI
+          - name: BLOCKED_NS_LIST
+            valueFrom:
+              configMapKeyRef:
+                name: avi-k8s-config
+                key: blackListedNamespaces
```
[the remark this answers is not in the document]
akshayhavile: changed as per the comment

apalsule (approved):

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

swathinsankaran (approved):
LGTM.

One question, do we need to specify some namespaces instead of an empty list in `values.yaml`?

akshayhavile:
> LGTM.
> 
> One question, do we need to specify some namespaces instead of an empty list in `values.yaml`?

As per deployment (openshift/Kubernetes) namespaces will change. and it is optional field. As every customer will not have requirement to block namespace, we have kept it empty.

PR #933, opened 2022-09-26
[cherry-pick to release 1.8.1]fix to clear cache during partial delete

swathinsankaran:
This PR adds the code to clear the cache when a partial delete happens. During partial delete, the cache is not cleared properly and the AKO does PUT calls to the controller instead of POST calls.


swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

akshayhavile (approved):

PR #947, opened 2022-09-30
AV-156959 Nodenetwork list mandatory fix for nsx-t vlan

aaha97:


swathinsankaran (approved):
LGTM

internal/lib/lib.go
```diff
@@ -1378,7 +1378,8 @@ func IsPublicCloud() bool {
 
 func IsNodeNetworkAllowedCloud() bool {
 	cloudType := GetCloudType()
-	if cloudType == CLOUD_NSXT || cloudType == CLOUD_VCENTER {
+	if (cloudType == CLOUD_NSXT && GetNSXTTransportZone() == VLAN_TRANSPORT_ZONE) ||
```
akshayhavile, line 1381: @aaha97 can you please check at this point value of TransportZone will be set? I think setNSXTTransportZone() is getting called pretty late. Please re-confirm. I think this will create problem in `checkNodeNetwork` call only as it will not process `nodenetworkmap` related validation and return from line number 3077 of `checkNodeNetwork` function

aaha97: the function checkAndSetCloudType uses GetNSXTTransportZone, 
and IsNodeNetworkAllowedCloud is called in checkNodeNetwork which executes after checkAndSetCloudType.

so we should be good

akshayhavile (approved):

PR #954, opened 2022-10-12
AV-158136: Validate enableSSL on listener Port only when listener settings are configured

akshayhavile:
…

### DIFF
--- a/internal/k8s/crdcontroller.go
+++ b/internal/k8s/crdcontroller.go
@@ -698,7 +698,7 @@ func validateHostRuleObj(key string, hostrule *akov1alpha1.HostRule) error {
 		}
 	}
 
-	if hostrule.Spec.VirtualHost.TCPSettings != nil {
+	if hostrule.Spec.VirtualHost.TCPSettings != nil && len(hostrule.Spec.VirtualHost.TCPSettings.Listeners) > 0 {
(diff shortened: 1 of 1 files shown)


akshayhavile:
build ako

internal/k8s/crdcontroller.go
```diff
@@ -698,7 +698,7 @@ func validateHostRuleObj(key string, hostrule *akov1alpha1.HostRule) error {
 		}
 	}
 
-	if hostrule.Spec.VirtualHost.TCPSettings != nil {
+	if hostrule.Spec.VirtualHost.TCPSettings != nil && len(hostrule.Spec.VirtualHost.TCPSettings.Listeners) > 0 {
```
swathinsankaran, line 701: Do we have to inform the user in such a scenario?

akshayhavile: @swathinsankaran `SSL enabled ` error should be raised only when listeners are specified. TCPSetting can have other setting without specifying listeners e.g. laodBalancerIP.

akshayhavile: So we do not require to notify anything to customer

swathinsankaran (approved):
LGTM

aaha97 (approved):

PR #966, opened 2022-11-02
Service migration script

saurabhvaidvmware:


scripts/service_migration.py
```diff
@@ -0,0 +1,87 @@
...
+    Arguments:
+    v1 :: CoreV1Api object
+    svc :: original service as dictionary
+    --------------
+    Returns:
+    None
+    ************************************************************************
+    '''
+    
+    svc['metadata']['annotations']['load-balancer-ips'] = svc['status']['load_balancer']['ingress'][0]['ip']
```
apalsule, line 22: make the annotation a constant defined in script

apalsule:
General comment : get rid of all print statements added for debugging purposes, and only print actual errors. Also print an output of services updated and services not updated.

scripts/service_migration.py
saurabhvaidvmware: constant added to script

saurabhvaidvmware:
> General comment : get rid of all print statements added for debugging purposes, and only print actual errors. Also print an output of services updated and services not updated.

Print statements removed. Added list of Loadbalancer type services updated and not updated.

scripts/service_migration.py
```diff
@@ -0,0 +1,89 @@
+import copy
```
swathinsankaran, line 1: please add a copyright license agreement here.

scripts/service_migration.py
```diff
@@ -0,0 +1,89 @@
...
+from kubernetes import client, config
+from kubernetes.client.exceptions import ApiException
+import argparse
+import os.path
+from pprint import pprint
+
+IP_ANNOTATION = 'load-balancer-ips'
+def update_annot(v1, svc):
```
swathinsankaran, line 9: update_annotation will be more readable I think.

scripts/service_migration.py
```diff
@@ -0,0 +1,89 @@
...
+    Arguments:
+    v1 :: CoreV1Api object
+    svc :: original service as dictionary
+    --------------
+    Returns:
+    None
+    ************************************************************************
+    '''
+    
+    svc['metadata']['annotations'][IP_ANNOTATION] = svc['status']['load_balancer']['ingress'][0]['ip']
```
swathinsankaran, line 23: Better to check the presence of keys `status`, `load_balancer`, `ip` and length of svc.

scripts/service_migration.py
```diff
@@ -0,0 +1,89 @@
...
+    service = client.V1Service()
+    service.api_version = "v1"
+    service.kind = "Service"
+    service.metadata = svc['metadata']
+    service.spec = svc['spec']
+    service.status = svc['status']
+    try:
+        api_response = v1.patch_namespaced_service(name = svc["metadata"]["name"], namespace="avi-system", body=service)
+        #updated_service = v1.read_namespaced_service(name = svc["metadata"]["name"], namespace="avi-system")
```
swathinsankaran, line 32: we can remove this statement if we are not planning to add this in the future.

scripts/service_migration.py
```diff
@@ -0,0 +1,89 @@
...
+    except ApiException as e:
+        print(e)
+
+def get_svc(config_file):
```
swathinsankaran, line 36: get_svc name is misleading, it does get of services as well as update of annotation. I think a better naming is required here.

scripts/service_migration.py
```diff
@@ -0,0 +1,89 @@
...
+def main():
+    parser = argparse.ArgumentParser(description="Script to read all namespaced services and add external IP to service annotation for LoadBalancer type services")
+    parser.add_argument('kubeconfig_file', help="Path of kubeconfig file in local directory")
+    args = parser.parse_args()
+    config_file = args.kubeconfig_file
+    while not os.path.isfile(config_file):
+        config_file = input("Give valid kubeconfig filepath (Input \'End\' to exit program): ")
+        if config_file == 'End' or config_file=='end':
+            break
+    if os.path.isfile(config_file):
```
swathinsankaran, line 84: Are we doing the same condition check in L80?

swathinsankaran: We need it because there is a break statement to end the script.

scripts/service_migration.py
saurabhvaidvmware: Updated function name

scripts/service_migration.py
saurabhvaidvmware: Added copyright license agreement

scripts/service_migration.py
saurabhvaidvmware: Removed statement as per recommendation

scripts/service_migration.py
saurabhvaidvmware: Updated function name

scripts/service_migration.py
saurabhvaidvmware: Added check for existence of each key, and whether said keys hold null/empty values

scripts/service_migration.py
```diff
@@ -0,0 +1,111 @@
...
+        api_response = v1.patch_namespaced_service(name = svc["metadata"]["name"], namespace="avi-system", body=service)
+        return True
+    except ApiException as e:
+        print(e)
+        return False
+
+def fetch_lbsvc_for_update(config_file):
+    '''
+    ************************************************************************
+    Takes a kubeconfig file and retrieves all services found in all namespaces. If service if of type LoadBalancer, then calls update_annotation() to patch service
```
swathinsankaran, line 58: If service `is` of type LoadBalancer.

swathinsankaran (dismissed):
LGTM.

Please add a few scenarios that you have tested, both positive and negative.

saurabhvaidvmware:
I successfully ran the script with the following situations in the cluster:
1. No LoadBalancer type service present in the cluster.
2. LoadBalancer type service present - 
         a. and annotation missing.
         b. and annotation already populated.
         c. but status field empty.

akshayhavile (approved):
minor nit. LGTM

swathinsankaran (approved):

akshayhavile (approved):

swathinsankaran:
build ako

apalsule (approved):

PR #975, opened 2022-11-09
AV-159292: Fix for Error datascript for passthrough VS

akshayhavile:
Manual Testing done as follow:
1. create passthrough VS without infrasetting
2. Create passthrough VS with infrasetting
3. Transition passthrough VS with to without infrasetting and vice versa.
4. Traffic test

TODO: Run FT suits.

### DIFF
--- a/internal/nodes/avi_model_tls_passthrough.go
+++ b/internal/nodes/avi_model_tls_passthrough.go
@@ -97,10 +97,14 @@ func (o *AviObjectGraph) BuildGraphForPassthrough(svclist []IngressHostPathSvc, 
 	}
 	dsNode := datascriptList[0]
 	var infrasettingName string
+	var infraSettingNameWithSuffix string
 	var pgName string
 	if infraSetting != nil {
 		infrasettingName = infraSetting.Name
+		infraSettingNameWithSuffix = infrasettingName + "-"
 	}
+	//Replace AVIINFRA with infrasettingname if present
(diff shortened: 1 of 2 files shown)


internal/nodes/avi_model_tls_passthrough.go
```diff
@@ -97,10 +97,14 @@ func (o *AviObjectGraph) BuildGraphForPassthrough(svclist []IngressHostPathSvc,
...
 	dsNode := datascriptList[0]
 	var infrasettingName string
+	var infraSettingNameWithSuffix string
 	var pgName string
 	if infraSetting != nil {
 		infrasettingName = infraSetting.Name
+		infraSettingNameWithSuffix = infrasettingName + "-"
 	}
+	//Replace AVIINFRA with infrasettingname if present
+	dsNode.Script = strings.Replace(dsNode.Script, "AVIINFRA", infraSettingNameWithSuffix, 1)
```
apalsule, line 107: isnt it better to just append infrasetting name rather than replacing it ?

akshayhavile: @apalsule: `PoolGroup` name text is in the middle of datascript text (dsNode.Script). So replacing the exact value with avi-infrasetting placeholder is the way to go ahead.

apalsule (approved):

swathinsankaran (approved):

PR #976, opened 2022-11-09
[AV-150821] fix for status IPs not getting updated properly for some Ingresses/Routes

swathinsankaran:
This PR contains the code changes to copy the parent reference while adding the child VS cache object to the cache.

In the rest layer in order to populate the IP address for the status of Ingress/Routes, we either traverse through the VS VIP key collections present in the VS cache or traverse through the parent VS's VSVIP Collection and fetch the IP Addresses mapped to it.
But for a child VS the latter will be the approach taken to figure out the IP address. But when the child VS is newly added for the first time, the parent reference is not added properly, which resulted in an empty IP address getting populated for it.

TODO: FT runs

### DIFF
--- a/internal/rest/avi_obj_vs.go
+++ b/internal/rest/avi_obj_vs.go
@@ -680,6 +680,9 @@ func (rest *RestOperations) AviVsCacheAdd(rest_op *utils.RestOp, key string) err
 			if lastModifiedStr == "" {
 				vs_cache_obj.InvalidData = true
 			}
+			if vhParentKey != nil {
+				vs_cache_obj.ParentVSRef = vhParentKey.(avicache.NamespaceName)
+			}
 
 			rest.cache.VsCacheMeta.AviCacheAdd(k, vs_cache_obj)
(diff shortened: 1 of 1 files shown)


swathinsankaran:
build ako

apalsule (approved):

swathinsankaran:
build ako

internal/rest/avi_obj_vs.go
```diff
@@ -847,8 +850,12 @@ func (rest *RestOperations) isHostPresentInSharedPool(hostname string, parentVs
 
 func (rest *RestOperations) GetIPAddrsFromCache(vsCache *avicache.AviVsCache) []string {
 	var IPAddrs []string
-	if len(vsCache.VSVipKeyCollection) == 0 {
-		parentVSKey := vsCache.ParentVSRef
+	vsCacheCopy, ok := vsCache.GetVSCopy()
```
akshayhavile, line 853: I did not get purpose behind this change. is new variable required over here?

swathinsankaran: yes, since we are passing the pointer of child VS and overwriting the pointer with the parent VS object.

akshayhavile (approved):

PR #978, opened 2022-11-11
[HA] Rest layer code changes

swathinsankaran:
This PR includes the following changes in the rest layer:

1. Adds a time delay before doing any REST operation towards the controller in the passive AKO to give the active AKO enough time to create objects in the controller.

2. In the passive AKO, all the REST operations are modified into GET operations, and the path in POST operation is modified to `path + "?name=" + objname`. Sometimes the GET operation for the path `path + "?name=" + objname` returns 0, added a retry so that active AKO creates the object and passive AKO reads the object in the subsequent GET calls.

3. During failover, there is a high chance to get 409 (object already exists) errors from the controller. The rest operation code is modified to build the cache and retry on getting a 409 error.

4. In passive AKO, the GET could sometimes return non-404 errors for an Ingress/Route/L4 Svc deletion. The code is modified to retry the GET in such scenarios.

5. For a GET call, the controller returns parent_UUID in the format `http://<controller-ip>/api/virtualservice/<uuid>` but AKO expects the uuid in the format `http://<controller-ip>/api/virtualservice/<uuid>#<parent-vs-name>` while building the cache. The commit also modifies this behavior for a passive AKO to expect only the uuid in the response and get the parent vs name from the key published.
**NOTE:** This was coming because of a missing query parameter in the URI which is resolved in the recent commits.

6. Modifies the status sync function to honor the state(leader/follower) of the AKO.

### DIFF
--- a/internal/rest/avi_obj_vs.go
+++ b/internal/rest/avi_obj_vs.go
@@ -570,7 +570,7 @@ func (rest *RestOperations) AviVsCacheAdd(rest_op *utils.RestOp, key string) err
 		return errors.New("Error rest_op")
 	}
 
-	resp_elems := RestRespArrToObjByType(rest_op, "virtualservice", key)
+	resp_elems := rest.restOperator.RestRespArrToObjByType(rest_op, "virtualservice", key)
 	if resp_elems == nil {
 		utils.AviLog.Warnf("key: %s, msg: unable to find vs obj in resp %v", key, rest_op.Response)
 		return errors.New("vs not found")
@@ -607,6 +607,9 @@ func (rest *RestOperations) AviVsCacheAdd(rest_op *utils.RestOp, key string) err
 		if found_parent {
 			// the uuid is expected to be in the format: "https://IP:PORT/api/virtualservice/virtualservice-88fd9718-f4f9-4e2b-9552-d31336330e0e#mygateway"
 			vs_uuid := avicache.ExtractUuid(vh_parent_uuid.(string), "virtualservice-.*.#")
+			if !lib.AKOControlConfig().IsLeader() {
+				vs_uuid = avicache.ExtractUuidWithoutHash(vh_parent_uuid.(string), "virtualservice-.*.")
+			}
 			utils.AviLog.Debugf("key: %s, msg: extracted the vs uuid from parent ref: %s", key, vs_uuid)
 			// Now let's get the VS key from this uuid
 			var foundvscache bool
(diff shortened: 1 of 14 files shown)


internal/rest/rest_operation.go
```diff
@@ -245,7 +250,153 @@ type AviRestClientPool struct {
...
+					l.restOp.PublishKeyToSlowRetryLayer(publishKey, key)
+				}
+			}
+			return false, processNextObj
+		}
+	}
+	return true, true
+}
+
+func (l *leader) AviRestOperateWrapper(aviClient *clients.AviClient, rest_ops []*utils.RestOp, key string) error {
```
swathinsankaran, line 383: needs to change.

internal/rest/rest_operation.go
```diff
@@ -270,17 +421,210 @@ func AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp) error {
...
+		return errors.New("required to populate cache and then retry")
+	}
+	return nil
+}
+
+type follower struct {
+	restOp *RestOperations
+}
+
+func (f *follower) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp, aviObjKey avicache.NamespaceName, avimodel *nodes.AviObjectGraph, key string, isEvh bool, sslKey ...utils.NamespaceName) (bool, bool) {
```
swathinsankaran, line 468: <comment from code review> Code duplication must be minimized.

internal/rest/rest_operation.go
```diff
@@ -270,17 +421,210 @@ func AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp) error {
...
+
+func (f *follower) AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp, key string) error {
+	for i, op := range rest_ops {
+		SetTenant := session.SetTenant(op.Tenant)
+		SetTenant(c.AviSession)
+		if op.Version != "" {
+			SetVersion := session.SetVersion(op.Version)
+			SetVersion(c.AviSession)
+		}
+		op.Path += "?name=" + op.ObjName
```
swathinsankaran, line 613: <comment from code review> createdby, tenant, cloud must be appended since the name may not be unique.

internal/rest/avi_obj_vs.go
```diff
@@ -607,6 +607,9 @@ func (rest *RestOperations) AviVsCacheAdd(rest_op *utils.RestOp, key string) err
 		if found_parent {
 			// the uuid is expected to be in the format: "https://IP:PORT/api/virtualservice/virtualservice-88fd9718-f4f9-4e2b-9552-d31336330e0e#mygateway"
 			vs_uuid := avicache.ExtractUuid(vh_parent_uuid.(string), "virtualservice-.*.#")
+			if !lib.AKOControlConfig().IsLeader() {
```
swathinsankaran, line 610: comment from code review: Not required once we add the `include_name=true` to the path for GET operations.

internal/rest/rest_operation.go
swathinsankaran: Addressed

internal/rest/rest_operation.go
swathinsankaran: Addressed.

internal/rest/avi_obj_vs.go
swathinsankaran: Addressed

internal/rest/rest_operation.go
swathinsankaran: Addressed

internal/rest/rest_operation.go
```diff
@@ -270,17 +299,93 @@ func AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp) error {
...
+		utils.AviLog.Debugf("key: %s, msg: aborted the rest operation due the reason: %s", key, err.Error())
+		return true
+	}
+	return false
+}
+
+func (f *follower) AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp, key string) error {
+
+	// Delay the REST calls in the follower.
+	<-time.After(500 * time.Millisecond)
```
swathinsankaran, line 358: requires a comment here.

internal/rest/rest_operation.go
```diff
@@ -270,17 +299,93 @@ func AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp) error {
...
+
+	for i, op := range rest_ops {
+		SetTenant := session.SetTenant(op.Tenant)
+		SetTenant(c.AviSession)
+		if op.Version != "" {
+			SetVersion := session.SetVersion(op.Version)
+			SetVersion(c.AviSession)
+		}
+
+		op.Path += "?name=" + op.ObjName + "&include_name=true"
```
swathinsankaran, line 368: comment here.

PR #994, opened 2022-11-21
Support for LoadBalancerIP annotation

saurabhvaidvmware:


internal/nodes/avi_model_advl4_translator.go
```diff
@@ -447,7 +447,11 @@ func (o *AviObjectGraph) ConstructSharedVipSvcLBNode(sharedVipKey, namespace, ke
 		}
 
 		if i == 0 {
-			sharedPreferredVIP = svcObj.Spec.LoadBalancerIP
+			if svcObj.Spec.LoadBalancerIP != "" {
+				sharedPreferredVIP = svcObj.Spec.LoadBalancerIP
+			} else if svcObj.Spec.LoadBalancerIP == "" && svcObj.Annotations[lib.LoadBalancerIP] != "" {
```
akshayhavile, line 452: @saurabhvaidvmware: I think we can avoid first check `svcObj.Spec.LoadBalancerIP== ""`. This should be done everywhere.

internal/nodes/dequeue_ingestion.go
```diff
@@ -508,16 +508,28 @@ func handleL4SharedVipService(namespacedVipKey, key string, fullsync bool) {
 		// Initializing the preferred VIP from the first Service we get, so any other Service
 		// that wishes for static IP allocation differently conflicts with this.
 		if i == 0 {
-			sharedVipLBIP = svcObj.Spec.LoadBalancerIP
+			if svcObj.Spec.LoadBalancerIP != "" {
```
swathinsankaran, line 511: these conditions are repeated everywhere. It will be better if we can move this to a function.

saurabhvaidvmware: Added functions in util.go to handle condition checks

internal/nodes/avi_model_advl4_translator.go
saurabhvaidvmware: Dropped this condition check.

DixitAakash:
build ako

akshayhavile (approved):

PR #998, opened 2022-11-25
[HA] Configmap changes

swathinsankaran:
This PR contains the following changes:
1. Allows the leader AKO to delete the objects in the controller created by
the AKO when the deleteConfig flag is set to True.

2. Leader AKO updates the statefulset annotation when the deleteConfig flag is set or unset.


akshayhavile:
@swathinsankaran : Have we tested `deleteConfigmap` related changes? like 
1. changing `deleteConfigmap` variable from true to false and vice versa.
2. Reboot AKO with deleteConfigmap set to true and set to false.

swathinsankaran:
@akshayhavile, Yes I have tested the following scenarios.
1. Set deleteConfig as true without any failover(normal)
2. Set deleteConfig as false without any failover(normal)
3. Set deleteConfig as true and then failover
4. Set deleteConfig as false and then failover

PR #1013, opened 2022-12-14
AV-160076 Cleanup istio resources after istio is disabled

aaha97:
istio resources should now get cleaned up as per regular workflow if istioenabled is set to false or deleteconfig is set to true

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -355,7 +355,12 @@ func (c *AviObjCache) DeleteUnmarked(childCollection []string) {
 	for _, objkey := range c.SSLKeyCache.AviGetAllKeys() {
 		intf, _ := c.SSLKeyCache.AviCacheGet(objkey)
 		if obj, ok := intf.(*AviSSLCache); ok {
-			if obj.HasReference == false && obj.Name != lib.GetIstioWorkloadCertificateName() {
+			if obj.HasReference == false {
+				// if deleteConfig is false and istio is enabled, do not delete istio sslkeycert
+				if obj.Name == lib.GetIstioWorkloadCertificateName() &&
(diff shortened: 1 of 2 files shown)


internal/cache/controller_obj_cache.go
```diff
@@ -355,7 +355,12 @@ func (c *AviObjCache) DeleteUnmarked(childCollection []string) {
 	for _, objkey := range c.SSLKeyCache.AviGetAllKeys() {
 		intf, _ := c.SSLKeyCache.AviCacheGet(objkey)
 		if obj, ok := intf.(*AviSSLCache); ok {
-			if obj.HasReference == false && obj.Name != lib.GetIstioWorkloadCertificateName() {
+			if obj.HasReference == false {
+				// if deleteConfig is false and istio is enabled, do not delete istio sslkeycert
+				if !lib.GetDeleteConfigMap() && lib.IsIstioEnabled() &&
```
akshayhavile, line 360: Should we optimize this condition as `obj.Name == lib.GetIstioWorkloadCertificateName() && !lib.GetDeleteConfigMap() && lib.IsIstioEnabled()` 
Reason: With current condition, For most of the objects, first two conditions will be true (when deleteConfig is false and istio is disabled) and then last condition will be false. 
So we can avoid processing these two conditions for remaining objects

akshayhavile: Similar changes at other places.

aaha97: changed it to have the order of rarity of encounter, name > istioEnabled true > deleteConfig false

swathinsankaran (approved):
LGTM.

swathinsankaran:
build ako

akshayhavile (approved):

aaha97:
build ako

PR #1026, opened 2022-12-23
fix for AV-165125

swathinsankaran:
This PR fixes a crash issue seen in NPL deployments.

For NPL deployments the parameter `Key` was not filled in the `StatusOptions` from the rest layer and it was getting accessed at the follower AKO dequeue status method. Hence the crash.

With this commit, the parameter `Key` will be filled by the rest layer in all cases.


swathinsankaran:
build ako

swathinsankaran:
build ako

arihantg (approved):

akshayhavile (approved):

PR #1027, opened 2022-12-23
fix for AV-150791

swathinsankaran:
This commit fixes an issue where the VS VIP goes down with the error message `No network/subnet matched Virtual Service IP`.

On debugging the code, the AKO was not filling the discovered networks correctly in one PUT operation of VS VIP resulting in this error.

This commit ensures that the discovered network is filled correctly in VS VIP before the PUT operations.

Tested scenarios:
1. Deleted the ingress which resulted in a VS VIP PUT operation with the discovered networks in its payload.


akshayhavile (approved):

aaha97 (approved):
LGTM

PR #1038, opened 2023-01-17
[Doc] [AV-164899] Documentation for `securityContext`

swathinsankaran:
This PR adds the documentation for `securityContext`.

### DIFF
--- a/Dockerfile.ako
+++ b/Dockerfile.ako
@@ -18,7 +18,7 @@ RUN GOARCH=amd64 \
 
 FROM ${photon_src_repo}
 RUN yum install -y tar.x86_64
-COPY --from=build /go/bin/akc .
+COPY --from=build /go/bin/akc /usr/bin/
 COPY ./scripts/pre_stop_hook.sh /var/pre_stop_hook.sh
 EXPOSE 8080
-ENTRYPOINT ["./akc"]
+ENTRYPOINT ["/usr/bin/akc"]
--- a/docs/values.md
+++ b/docs/values.md
@@ -283,6 +283,10 @@ It might not be desirable to have all the nodes of a kubernetes cluster to parti
 
 By default, AKO prints all the logs to stdout. Instead, persistentVolumeClaim(PVC) can be used for publishing logs of AKO pod to a file in PVC. To use this, the user has to create a PVC (and a persistent volume, if required) and specify the name of the PVC as the value of persistentVolumeClaim.
 
+### securityContext
+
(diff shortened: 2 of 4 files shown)


swathinsankaran:
build ako

docs/values.md
```diff
@@ -283,6 +283,10 @@ It might not be desirable to have all the nodes of a kubernetes cluster to parti
 
 By default, AKO prints all the logs to stdout. Instead, persistentVolumeClaim(PVC) can be used for publishing logs of AKO pod to a file in PVC. To use this, the user has to create a PVC (and a persistent volume, if required) and specify the name of the PVC as the value of persistentVolumeClaim.
 
+### securityContext
```
akshayhavile, line 286: @swathinsankaran: `securityContext` applies to individual container.  For AKO, do we really want to set individual container specific security contexts? Can we apply same set of context at POD level?  Should we have separate field?

swathinsankaran: I agree we have only a single container. But some options are unique to`securityContext`.

swathinsankaran: I checked the options which may cause issues in AKO if we set `securityContext`.

 `runAsNonRoot: true` won't work as we are not providing any user during image creation and it is taking the default user which is `root`.

Tried the following settings and it is working.
```
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      readOnlyRootFilesystem: false
      seccompProfile:
        type: RuntimeDefault
```

swathinsankaran:
build ako

docs/values.md
swathinsankaran: As discussed, I have changed the binary location to `/usr/bin` so that setting `runAsNonRoot: true` won't return an error.

Dockerfile.ako
```diff
@@ -18,7 +18,7 @@ RUN GOARCH=amd64 \
 
 FROM ${photon_src_repo}
 RUN yum install -y tar.x86_64
-COPY --from=build /go/bin/akc .
+COPY --from=build /go/bin/akc /usr/bin/
 COPY ./scripts/pre_stop_hook.sh /var/pre_stop_hook.sh
 EXPOSE 8080
-ENTRYPOINT ["./akc"]
+ENTRYPOINT ["/usr/bin/akc"]
```
swathinsankaran, line 24: @abhishekbsingh will this cause any issues in vcf env?

akshayhavile (approved):

aaha97 (approved):
LGTM

Dockerfile.ako
abhishekbsingh: It should be okay.

abhishekbsingh (approved):

PR #1043, opened 2023-01-25
Doc: LB annotation support for static IP

akshayhavile:


akshayhavile:
@apalsule: I have changed min controller version supported to 21.1.5.

akshayhavile:
build ako

docs/objects.md
```diff
@@ -43,6 +43,28 @@ spec:
...
 ```
+As `spec.loadBalancerIP` field has been deprecated from kubernetes version 1.24 onwards, AKO has introduced a new annotation `ako.vmware.com/load-balancer-ip` to L4 service that will be used to specify preferred IP. Example usage could look something like this:
+
+```
+apiVersion: v1
+kind: Service
+metadata:
+  name: avisvc-lb
+  namespace: red
+  ako.vmware.com/load-balancer-ip: "10.20.10.100"
```
swathinsankaran, line 54: the annotation must be under the metadata.annotations.
for eg:
```yaml
annotations:
    ako.vmware.com/load-balancer-ip: "10.20.10.100"
```

docs/objects.md
```diff
@@ -43,6 +43,28 @@ spec:
...
+    name: eighty
+  selector:
+    app: avi-server
+```
+
+***Note***:
+1. If customer is upgrading to Kubernetes version 1.24+, user has to manually update L4 service definition with above annotation in case user wants to use preferred IP feature.
+2. User should specify either `ako.vmware.com/load-balancer-ip` annotation or `spec.loadBalancerIP` field in service definition if kubernetes version is less than 1.24. 
 
 Avi does not allow users to update preferred virtual IPs bound to a particular virtualservice. Therefore in order to update the user preferred IP, it is required to re-create the Service object, failing which Avi/AKO throws an error. The following transition cases should be kept in mind, and for these, an explicit Service re-create with changed configuration is required.
```
swathinsankaran, line 69: Do we have to mention the transition cases for annotations also?

docs/objects.md
```diff
@@ -43,6 +43,28 @@ spec:
...
  - updating loadBalancerIP value, from `loadBalancerIP: 10.10.10.11` to `loadBalancerIP: 10.10.10.22`.
```
saurabhvaidvmware, line 70: Should we also mention the case of transitioning a service from spec.loadBalancerIP to the LBIP annotation?

docs/objects.md
akshayhavile: addressed

swathinsankaran (approved):
LGTM

swathinsankaran:
build ako

saurabhvaidvmware (approved):
LGTM

saurabhvaidvmware (approved):

PR #1049, opened 2023-02-02
Fix Istio resource creation

aaha97:
Istio Bootstrap needs to execute after leader election.
need to use key instead of model name (name is empty in model) for certificate rotation flow


arihantg (approved):
LGTM

akshayhavile (approved):

PR #1060, opened 2023-02-16
AV-166913 synchronisation in ingresstests

arihantg:


tests/ingresstests/l7_rest_test.go
```diff
@@ -829,6 +829,7 @@ func TestMultiHostIngressStatusCheck(t *testing.T) {
 	}
 	integrationtest.AddSecret("my-secret-v2", "default", "tlsCert", "tlsKey")
 	integrationtest.AddSecret("my-secret", "default", "tlsCert", "tlsKey")
+	time.Sleep(2 * time.Second) //secrets should be created before creating ingress
```
akshayhavile, line 832: @arihantg: Instead of adding sleep, if you think, secret creation is taking time, should we add it as part of `SetUpTestForIngress` function? so that secret will be created in sufficient time.

arihantg: this secret is used only in this UT so can;t be added in SetUpTestForIngress

arihantg:
build ako

tests/ingresstests/l7_rest_test.go
swathinsankaran: instead of adding the sleep, we can check where the secret is present in the shared informer cache first and then proceed with the ingress creation.

tests/oshiftroutetests/oshift_route_passthrough_test.go
```diff
@@ -401,6 +401,8 @@ func TestPassthroughRouteWithAlternateBackends(t *testing.T) {
 	integrationtest.CreateSVC(t, "default", "absvc2", corev1.ServiceTypeClusterIP, false)
 	integrationtest.CreateEP(t, "default", "absvc2", false, false, "3.3.3")
 
+	time.Sleep(2 * time.Second) //service should be created before route
```
swathinsankaran, line 404: here also, we have to check the shared informer cache and then proceed with the route creation.

arihantg:
build ako

arihantg:
build ako

swathinsankaran (approved):
LGTM

akshayhavile (approved):

PR #1068, opened 2023-02-27
[bug-fix] Secret handling during bootup

swathinsankaran:
This PR adds the code to do bulk get and push of secret objects during bootup


swathinsankaran:
closing the PR as it is having few unwanted changes.

PR #1073, opened 2023-03-01
AV-171818: Fix vrf cache issue in update node case

akshayhavile:
We are fixing two issues as part of one bug id.
This PR fixes one issue as follow.
Issue: When AKO is running, AKO VRF cache is build. If multiple node update occurs, route-id for each route was getting populated same. Due to this PATCH to vrfcontext was failing due to duplicate route-id.

Resolution: When update to existing node come, use existing route-id present in cache.

This PR fixes this issue.


swathinsankaran (approved):
LGTM

DixitAakash (approved):

swathinsankaran (approved):
LGTM

PR #1077, opened 2023-03-02
[fix] Restrict secret handling by using a flag in `values.yaml`

swathinsankaran:
This PR adds a flag in `values.yaml` by using which the secret handling can be restricted only to the namespace where the AKO is present.


swathinsankaran:
build ako

helm/ako/values.yaml
```diff
@@ -36,7 +36,7 @@ AKOSettings:
   #   - kube-system
   #   - kube-public
   ipFamily: "" # This flag can take values V4 or V6 (default V4). This is for the backend pools to use ipv6 or ipv4. For frontside VS, use v6cidr
-
+  handleSecretsFromAKONSOnly: "false" # If the flag is switched on, handling of the secrets will be restricted only to the namespace where the AKO is installed.
```
akshayhavile, line 39: Can we re-word it as: `If this flag is set to true, AKO will handle secrets from the namespace where AKO is installed`

internal/k8s/validator.go
```diff
@@ -192,6 +194,23 @@ func (l *leader) ValidateHostRuleObj(key string, hostrule *akov1alpha1.HostRule)
...
+	// reject the hostrule if the secret handling is restricted to the namespace where
+	// AKO is installed.
+	if utils.GetInformers().RouteInformer != nil &&
+		namespace != utils.GetAKONamespace() &&
+		utils.IsSecretsHandlingRestrictedToAKONS() {
+		err := fmt.Errorf("secret handling is restricted to %s namespace only", utils.GetAKONamespace())
+		return err
+	}
+	_, err := utils.GetInformers().SecretInformer.Lister().Secrets(namespace).Get(secretName)
+	if err != nil {
```
akshayhavile, line 208: can we just return err here. `err` will be either nil or some value.

swathinsankaran:
build ako

internal/k8s/validator.go
swathinsankaran: I have addressed the comment.

helm/ako/values.yaml
swathinsankaran: I have addressed the comment.

akshayhavile (dismissed):

arihantg (dismissed):

swathinsankaran:
@akshayhavile, @arihantg Could you please review this PR again? I have added two UTs to test the TLS in hostrule.

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

PR #1082, opened 2023-03-03
AV-171818 Fixing static route population

aaha97:


internal/lib/lib.go
```diff
@@ -1921,8 +1921,11 @@ func GetIPFromNode(node *v1.Node) (string, string) {
 		if addr.Type == corev1.NodeInternalIP {
 			nodeIP := addr.Address
 			if utils.IsV4(nodeIP) {
-				nodeV4 = nodeIP
-			} else {
+				//Use first IP in the list
```
swathinsankaran, line 1924: Better to add a UT for this.

swathinsankaran (approved):

akshayhavile (approved):

PR #1085, opened 2023-03-03
[Cherry-pick] AV-171375: Restricts secret handling by using a flag in values.yaml

swathinsankaran:
This PR adds a flag in values.yaml by using which the secret handling can be restricted only to the namespace where the AKO is present.

Tested Scenario:
Added 14K secrets in namespace test and rebooted the AKO.
```
root@shalini-python3-dev:/home/aviuser/swathin# ock get secret -n test | wc -l
14206
root@shalini-python3-dev:/home/aviuser/swathin
```

1. With the flag set as true
```
CONTAINER           CPU %               MEM                 DISK                INODES
0f3c6e9f6f45b       0.01                26.72MB             32.77kB             19
```
2. With flag set as false
```
CONTAINER           CPU %               MEM                 DISK                INODES
b929eb67e0602       0.24                456.5MB             16.38kB             19
```

### DIFF
--- a/docs/install/helm.md
+++ b/docs/install/helm.md
@@ -177,6 +177,7 @@ The following table lists the configurable parameters of the AKO chart and their
 | `AKOSettings.blockedNamespaceList` | List of K8s/Openshift namespaces blocked by AKO | `Empty List` |
 | `AKOSettings.istioEnabled` | set to true if user wants to deploy AKO in istio environment (tech preview)| false |
 | `AKOSettings.ipFamily` | set to V6 if user wants to deploy AKO with V6 backend (vCenter cloud with calico CNI only) (tech preview)| V4 |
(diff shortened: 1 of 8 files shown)


swathinsankaran:
build ako

internal/lib/constants.go
```diff
@@ -36,6 +36,7 @@ const (
 	BGP_PEER_LABELS                            = "BGP_PEER_LABELS"
 	SEG_NAME                                   = "SEG_NAME"
 	BLOCKED_NS_LIST                            = "BLOCKED_NS_LIST"
+	HANDLE_SECRETS_FROM_AKO_NS_ONLY            = "HANDLE_SECRETS_FROM_AKO_NS_ONLY"
```
akshayhavile, line 39: should we take out this?

akshayhavile (approved):

DixitAakash (commented):
LGTM

PR #1089, opened 2023-03-21
AV-173914 update max version of avi controller

arihantg:


arihantg:
build ako

internal/rest/avi_evh_obj.go
```diff
@@ -519,6 +520,7 @@ func (rest *RestOperations) AviVsChildEvhBuild(vs_meta *nodes.AviEvhVsNode, rest
 			Path: &path_match,
 		}
 		vHMatchRule := &avimodels.VHMatchRule{
+			Name:    &name,
```
akshayhavile, line 523: @arihantg : Is `NAME` unique key? In case of multiple MatchRule it might create problem.

arihantg: there will only be a single matchrule for a hostname.So we can keep the value hardcoded

akshayhavile: It will create problem for hostrule with GSLB FQDN or alias feature.

swathinsankaran: @arihantg Have you tested the hostrule aliases and GSLB FQDN?

arihantg: I have tested this test  test_hostrule_fdqn_aliases in test_host_rules which is passing successfully.

akshayhavile (approved):

internal/rest/avi_evh_obj.go
swathinsankaran: Thanks @arihantg

swathinsankaran (approved):

PR #1096, opened 2023-03-28
 AV-174358  get ingress/route status from api server instead of informer

arihantg:


swathinsankaran:
@arihantg Changes look good to me. I was wondering why the number of commits is showing some  extra commits.

arihantg:
> @arihantg Changes look good to me. I was wondering why the number of commits is showing some extra commits.

These are probably due to merging done during sync done with master

akshayhavile (approved):
@arihantg: Changes looks good. Did we run UT pipeline multiple time to check that we are not hitting same issue again?

swathinsankaran (approved):
LGTM

arihantg:
> @arihantg: Changes looks good. Did we run UT pipeline multiple time to check that we are not hitting same issue again?
Have executed it multiple times on local and twice on jenkins.

PR #1105, opened 2023-04-07
[L4 CRD] Graph layer changes

swathinsankaran:
This PR contains the graph layer code changes for the L4 CRD.


akshayhavile:
build ako

internal/nodes/avi_model_l4_translator.go
```diff
@@ -136,6 +137,11 @@ func (o *AviObjectGraph) ConstructAviL4VsNode(svcObj *corev1.Service, key string
 		buildWithInfraSetting(key, avi_vs_meta, vsVipNode, infraSetting)
 	}
 
+	// Copy the VS properties from L4Rule object
+	if l4Rule, err := getL4Rule(key, svcObj); err == nil {
+		buildWithL4Rule(key, avi_vs_meta, l4Rule)
```
swathinsankaran, line 142: shared L4 VS need to take care.

swathinsankaran: Shared VIP VS will be taken care of in the next PR as it requires testing and UT changes.
cc: @akshayhavile

internal/nodes/avi_model_l7_hostname_shard.go
```diff
@@ -525,7 +526,7 @@ func (vsNode *AviVsNode) AddSSLPort(key string) {
 }
 func (vsNode *AviVsNode) DeletSSLRefInDedicatedNode(key string) {
 	vsNode.SSLKeyCertRefs = []*AviTLSKeyCertNode{}
-	vsNode.SSLProfileRef = ""
+	vsNode.SslProfileRef = proto.String("")
```
DixitAakash, line 529: This may cause problem from ConvertToRef point of view, since pointer to sslprofileref will not be nil.

swathinsankaran: I have addressed the comment. Thanks

PR #1108, opened 2023-04-09
Resolve clustername overlap issue while creating static route for cluster

akshayhavile:
Issue: If two clusters with name `orange` and `orange-new` are syncing to same Avicontroller, then rebooting AKO with clustername `orange` was deleting static routes of `orange-new`

Reason: AKO was doing prefix match for matching routes with clustername. That was causing issue.

Solution: USed regex to match routes belonging to same cluster.

### DIFF
--- a/internal/rest/avi_obj_vrf.go
+++ b/internal/rest/avi_obj_vrf.go
@@ -17,7 +17,8 @@ package rest
 import (
 	"encoding/json"
 	"errors"
-	"strings"
+	"fmt"
+	"regexp"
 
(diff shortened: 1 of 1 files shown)


swathinsankaran:
@akshayhavile Changes look good to me. Could you please add UTs also for this?

swathinsankaran (approved):

akshayhavile:
> @akshayhavile Changes look good to me. Could you please add UTs also for this?

@swathinsankaran UTs will be difficult as it will involve two clusters to test changes

akshayhavile:
build ako

arihantg (approved):

aaha97 (approved):
LGTM

PR #1119, opened 2023-04-18
AV-175659 : graph layer changes for SSORule CRD

DixitAakash:


internal/nodes/avi_model_evh_nodes.go
```diff
@@ -95,6 +96,13 @@ type AviVsEvhSniModel interface {
 
 	GetVHDomainNames() []string
 	SetVHDomainNames([]string)
+
+	GeneratedModel
+}
+
+type GeneratedModel interface {
+	GetGeneratedFields() *AviVsNodeGeneratedFields
```
swathinsankaran, line 104: Is the GeneratedModel interface used somewhere else? I think it will be better to have this added directly in the AviVsEvhSniModel instead of embedding.

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -2092,9 +2126,11 @@ func DeleteDedicatedEvhVSNode(vsNode *AviEvhVsNode, key string, hostsToRemove []
 func manipulateEvhNodeForSSL(key string, vsNode *AviEvhVsNode, evhNode *AviEvhVsNode) {
 	oldSSLProfile := vsNode.GetSSLProfileRef()
 	newSSLProfile := evhNode.GetSSLProfileRef()
-	if oldSSLProfile != "" && oldSSLProfile != newSSLProfile {
+	if oldSSLProfile != nil &&
+		*oldSSLProfile != "" &&
+		oldSSLProfile != newSSLProfile {
```
swathinsankaran, line 2131: are we missing a nil check of newSSLProfile here?

internal/nodes/avi_model_nodes_utils.go
```diff
@@ -0,0 +1,145 @@
...
+
+package nodes
+
+import (
+	"fmt"
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+
+	// avimodels "github.com/vmware/alb-sdk/go/models"
```
swathinsankaran, line 25: I think we can take out this commented import from the template present in the tool.

internal/nodes/ingress_model_rel.go
```diff
@@ -116,6 +116,11 @@ var (
 		Type:                           lib.ServiceImport,
 		GetParentMultiClusterIngresses: ServiceImportToMultiClusterIng,
 	}
+	OAuthSamlConfig = GraphSchema{
+		Type:               "OAuthSamlConfig",
```
swathinsankaran, line 120: Better to reuse `lib.OAuthSamlConfig` constant here.

internal/nodes/avi_model_evh_nodes.go
DixitAakash: Added the methods directly in AviVsEvhSniModel.

internal/nodes/avi_model_nodes_utils.go
DixitAakash: Taken out from both ako code and template.

internal/nodes/ingress_model_rel.go
DixitAakash: Done

internal/nodes/avi_model_evh_nodes.go
DixitAakash: Nil check not required as newSSLProfile will not be nil. Also, I have modified the condition from
```
if oldSSLProfile != nil &&
		*oldSSLProfile != "" &&
		oldSSLProfile != newSSLProfile {
		utils.AviLog.Warnf("key: %s msg: overwriting old ssl profile %s with new ssl profile %s", key, oldSSLProfile, newSSLProfile)
}
```
to 
```
if oldSSLProfile != nil &&
		*oldSSLProfile != "" &&
		*oldSSLProfile != *newSSLProfile {
		utils.AviLog.Warnf("key: %s msg: overwriting old ssl profile %s with new ssl profile %s", key, oldSSLProfile, newSSLProfile)
}
```
as we would end up comparing the pointers instead of values here `oldSSLProfile != newSSLProfile {`

swathinsankaran (approved):
LGTM

PR #1161, opened 2023-06-01
AV-180047: fix: AKO Bootup time is 36 min with 193 Nodes and 9k ingresses and 4k pods

akshayhavile:
Issue: AKO 1.10.1 taking 36 min to bootup for 193 nodes and 9k ingresses with 4k pods

Reason: As part of WCP merge, order of Object reading during AKO Full Sync is changed. As Ingresses were getting read first, all objects which can impact ingress object like httprule, ingressclass, aviinfrasetting etc. were causing update again on each ingress object leading to additional delay.

Resolution: Changed ordering back to older commit. It will solve delay issue.

### DIFF
--- a/internal/k8s/ako_init.go
+++ b/internal/k8s/ako_init.go
@@ -921,26 +921,6 @@ func (c *AviController) FullSyncK8s(sync bool) error {
 		}
 	}
 
-	if utils.GetInformers().IngressInformer != nil {
-		for namespace := range acceptedNamespaces {
-			ingObjs, err := utils.GetInformers().IngressInformer.Lister().Ingresses(namespace).List(labels.Set(nil).AsSelector())
-			if err != nil {
-				utils.AviLog.Errorf("Unable to retrieve the ingresses during full sync: %s", err)
-			} else {
-				for _, ingObj := range ingObjs {
-					key := utils.Ingress + "/" + utils.ObjKey(ingObj)
-					meta, err := meta.Accessor(ingObj)
-					if err == nil {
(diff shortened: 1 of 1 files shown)


internal/k8s/ako_init.go
```diff
@@ -1050,7 +1030,26 @@ func (c *AviController) FullSyncK8s(sync bool) error {
...
+						if err == nil {
+							resVer := meta.GetResourceVersion()
+							objects.SharedResourceVerInstanceLister().Save(key, resVer)
+						}
+						utils.AviLog.Debugf("Dequeue for ingress key: %v", key)
+						nodes.DequeueIngestion(key, true)
+					}
+				}
+			}
+		}
```
swathinsankaran, line 1052: Shall we add it at the end after if - else condition since this is a duplicate code?

akshayhavile: For WCP, `Gateway` being parent for `Ingress`, we might need to check impact on performance of adding that at the end. To keep that order of object creation same, I added that duplicate code.

swathinsankaran: sounds good.

@abhishekbsingh, could you please review this PR on priority?

swathinsankaran (approved):
LGTM

swathinsankaran:
build ako

abhishekbsingh (approved):

abhishekbsingh (approved):

DixitAakash (approved):
LGTM

PR #1166, opened 2023-06-06
AV-180638: Fix for the network profile for SCTP not coming as expected

swathinsankaran:
This PR fixes one regression introduced as part of the L4Rule and SSORule CRD code merge. The function `copier.Copy` that copies the value by using the field names was overwriting the application profile and network profile to an empty string. Because of this issue, the application profile and network profile were getting configured with their default values only every time.

TODO: FT


swathinsankaran:
build ako

akshayhavile (commented):
@swathinsankaran: Can we have UT/FT to cover this?

akshayhavile:
build ako

swathinsankaran:
UTs and fixes in UT is currently in-progress.

akshayhavile (approved):
@swathinsankaran: Overall changes looks good. Can error happen during copy? If yes, should we handle (atleast log that) so that in case of any wrong rest calls, we will get an idea.

akshayhavile (approved):
@swathinsankaran: Overall changes looks good. Can error happen during copy? If yes, should we handle (atleast log that) so that in case of any wrong rest calls, we will get an idea.

swathinsankaran:
build ako

swathinsankaran:
build ako

PR #1175, opened 2023-06-13
AV-179162: No Autogenerated domain for Dedicated Shard VS when AutoFQDN is `flat` or `default`

swathinsankaran:
This PR changes the number of domains added to a dedicated VS. The autogenerated FQDN will not be added to a Dedicated VS even if the value of AutoFQDN is set as **flat** or **default**.

**What's changed:** The previous behavior was to add two domains - one from the ingress/route and the other an autogenerated one based on the flag AutoFQDN. As part of this PR, the autogenerated FQDN is removed from Dedicated VS as it is not required.

**Reason:**  The autogenerated FQDN is added to retain the VIP of Shared VSes when the AVI controller is used with Infoblox IPAM. This behavior is not required for a Dedicated VS since the VIP is not shared, and the IP can be released when the VS is removed.

**Tested Scenarios:**
Manually checked the absence of generated domain in VSVIP of a dedicated VS.

**TODO:** FTs

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -187,9 +187,16 @@ func GetNSXTTransportZone() string {
 	return NsxTTzType
 }
 
-func GetFqdns(vsName, key string, subDomains []string) ([]string, string) {
+func GetFqdns(vsName, key string, subDomains []string, shardSize uint32) ([]string, string) {
 	var fqdns []string
 	var fqdn string
+
+	// Only one domain will be added for a Dedicated VS irrespective of
+	// the value given for the AutoFQDN.
+	if shardSize == 0 {
+		return fqdns, fqdn
+	}
+
 	autoFQDN := true
 	if GetL4FqdnFormat() == AutoFQDNDisabled {
 		autoFQDN = false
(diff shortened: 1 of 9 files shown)


swathinsankaran:
build ako

swathinsankaran:
build ako

internal/lib/lib.go
```diff
@@ -190,6 +190,11 @@ func GetNSXTTransportZone() string {
 func GetFqdns(vsName, key string, subDomains []string) ([]string, string) {
 	var fqdns []string
 	var fqdn string
+	// Only one domain is added for a Dedicated VS irrespective of
+	// the value given for the AutoFQDN.
+	if GetshardSize() == 0 {
```
akshayhavile, line 195: @swathinsankaran: From AviInfrasetting also we could get ShardSize. So we need to consider that also.

swathinsankaran: I have added the code to consider `AviInfraSetting`. Thanks

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

swathinsankaran:
build ako

akshayhavile (approved):

swathinsankaran:
build ako

swathinsankaran:
build ako

PR #1165, opened 2023-06-06
AV-180342: Fix: shared vip service support for mix protocols. Mix protocol support was not working when SCTP protocol is specified with either UDP/TCP

akshayhavile:
In UTs, parameter protocol is added while creating service. That resulted in changes in multiple files.

PR #1181, opened 2023-06-19
AV-180342: Fix: shared vip service support for mix protocols. Mix protocol support was not working when SCTP protocol is specified with either UDP/TCP (#1165)

akshayhavile:
…


akshayhavile:
build ako

DixitAakash (approved):

PR #1182, opened 2023-06-21
fix: add missing keys to Helm values file

alex1989hu:
These keys were removed in 2d8141a8a6be but the templating options were partially kept.

Having these keys will give option to users to override fields of the StatefulSet

The default value of affinity is copied from 9fef18449c26

### DIFF
--- a/helm/ako/values.yaml
+++ b/helm/ako/values.yaml
@@ -95,6 +95,21 @@ resources:
     cpu: 200m
     memory: 300Mi
 
+nodeSelector: {}
+
+tolerations: []
+
+affinity:
+  podAntiAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+    - labelSelector:
+        matchExpressions:
+        - key: app.kubernetes.io/name
+          operator: In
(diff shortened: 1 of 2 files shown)


Dhivyaaj:

akshayhavile:

helm/ako/values.yaml
```diff
@@ -95,6 +95,21 @@ resources:
     cpu: 200m
     memory: 300Mi
 
+nodeSelector: {}
+
+tolerations: []
+
+affinity:
+  podAntiAffinity:
```
swathinsankaran, line 103: @akshayhavile should we give the user the ability to change the affinity rules?

alex1989hu: hey @swathinsankaran in my use case I need to schedule AKO to specific worker nodes by setting conditions in affinity rule (e.g. egress firewall policy to the Avi controller)

swathinsankaran: Thanks @alex1989hu. We will plan to include this in the upcoming release.

PR #1222, opened 2023-07-29
AV-184860 : VS and Pools not deleted when multiple ingress use same infrasetting for shard size

DixitAakash:
This PR fixes the issue of VS and Pools not getting deleted when multiple ingress use same aviinfrasetting for shard size.
This issue is see when two ingress use the same ingress class and the ingress class refers to an aviinfrasetting object which sets the shard size. The vs and pool creation works fine. But on deleting the first ingress, the shard size to avinfrasetting mapping was also getting deleted even though it should still be in use with the second ingress. So, when the second ingress is deleted, the shard size from avinfrasetting object is not considered and hence wrong model name was calculated leading to failure in deletion.

Fix:
Check the ingress to aviinfasetting mapping in **IngRouteInfraSettingStore**. If no mapping exists, then only delete the shard size to aviinfrasetting mapping in **InfraSettingShardSizeStore**.

### DIFF
--- a/internal/objects/infrasetting.go
+++ b/internal/objects/infrasetting.go
@@ -59,10 +59,16 @@ func (v *AviInfraSettingL7Lister) UpdateIngRouteInfraSettingMappings(ingrouteNsN
 func (v *AviInfraSettingL7Lister) RemoveIngRouteInfraSettingMappings(ingrouteNsName string) bool {
 	v.InfraSettingIngRouteLock.Lock()
 	defer v.InfraSettingIngRouteLock.Unlock()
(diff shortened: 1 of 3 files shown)


DixitAakash:
build ako

akshayhavile (commented):
Can we please add some UT to test this?

DixitAakash:
> Can we please add some UT to test this?

@akshayhavile I have added the required UT's.

DixitAakash:
build ako

akshayhavile (approved):

swathinsankaran (approved):

PR #1248, opened 2023-08-15
AV-186801 fix retry with tenant switching

arihantg:


internal/lib/avi_api.go
```diff
@@ -66,10 +66,10 @@ func AviGet(client *clients.AviClient, uri string, response interface{}, retryNu
 		if aviError, ok := err.(session.AviError); ok && aviError.HttpStatusCode == 403 {
 			utils.AviLog.Debugf("Switching to admin context from %s", GetTenant())
 			SetAdminTenant := session.SetTenant(GetAdminTenant())
-			SetTenant := session.SetTenant(GetTenant())
 			SetAdminTenant(client.AviSession)
+			SetTenant := session.SetTenant(GetTenant())
```
akshayhavile, line 70: @arihantg What is the advantage of moving line from 69 to 70?

arihantg: just moved for ease in readability to group it with the defer statement on line 71

internal/lib/avi_api.go
```diff
@@ -66,10 +66,10 @@ func AviGet(client *clients.AviClient, uri string, response interface{}, retryNu
...
 			defer SetTenant(client.AviSession)
-			if err := AviGet(client, uri, response, retry+1); err != nil {
+			if err = AviGet(client, uri, response, retry+1); err != nil {
```
swathinsankaran, line 72: not sure how this solves the issue. Can this be simulated using our UTs or FTs?

arihantg: err variable has to be of a bigger scope. 
Issue was once this block is finished this local `err` variable values get destroyed and `err` now contains values obtained from line 63 (rest call with non-admin tenant).  
IT can be simulated with FTs for  non-admin tenant and we can check if ingress has cluster-uuid  populated,

swathinsankaran: I got it. Thanks. In that case, if the call with the admin tenant will return err as nil and should we proceed with the other checks in L77, L79, and all?

arihantg: added an else block to return cleanly.

arihantg:
build ako

arihantg:
build ako

swathinsankaran (approved):
LGTM

pkoshtavmware (approved):
LGTM

akshayhavile (approved):

PR #1251, opened 2023-08-21
[GatewayAPI] Graph layer changes httproute

aaha97:


ako-gateway-api/lib/lib.go
```diff
@@ -63,3 +63,18 @@ func FindListenerStatusByName(name string, status []gatewayv1beta1.ListenerStatu
 	}
 	return -1
 }
+
+func GetChildName(parentNs, parentName, routeNs, routeName, matchName string) string {
+	name := lib.GetNamePrefix() + parentNs + "-" + parentName + "-" + routeNs + "-" + routeName + "-" + utils.Stringify(utils.Hash(matchName)) + "-EVH"
```
swathinsankaran, line 68: Do we need the `-EVH` at the end?

ako-gateway-api/lib/lib.go
```diff
@@ -63,3 +63,18 @@ func FindListenerStatusByName(name string, status []gatewayv1beta1.ListenerStatu
...
+	return lib.Encode(name, lib.EVHVS)
+}
+
+func GetPoolName(parentNs, parentName, routeNs, routeName, matchName, backendNs, backendName, backendPort string) string {
+	name := lib.GetNamePrefix() + parentNs + "-" + parentName + "-" + routeNs + "-" + routeName + "-" + utils.Stringify(utils.Hash(matchName)) + "-" + backendNs + "-" + backendName + "-" + backendPort
+	return lib.Encode(name, lib.Pool)
+}
+
+func GetPoolGroupName(parentNs, parentName, routeNs, routeName, matchName string) string {
+	name := lib.GetNamePrefix() + parentNs + "-" + parentName + "-" + routeNs + "-" + routeName + "-" + utils.Stringify(utils.Hash(matchName)) + "-PG"
```
swathinsankaran, line 78: same here.

ako-gateway-api/nodes/avi_model_l7_translator.go
```diff
@@ -16,30 +16,39 @@ package nodes
...
+	avimodels "github.com/vmware/alb-sdk/go/models"
+	akogatewayapilib "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/ako-gateway-api/lib"
+	akogatewayapiobjects "github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/ako-gateway-api/objects"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/lib"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/internal/nodes"
 	"github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pkg/utils"
 )
 
-func (o *AviObjectGraph) ProcessL7Routes(key string, routeModel RouteModel, parentNsName string) {
+func (o *AviObjectGraph) ProcessL7Routes(key string, routeModel RouteModel, parentNsName string, listeners []string) {
```
swathinsankaran, line 33: are the listeners required here? can we take care of the listener deletion in DeleteStaleChild function instead of passing it to the buildChild?

ako-gateway-api/nodes/avi_model_l7_translator.go
```diff
@@ -16,30 +16,39 @@ package nodes
...
 	for _, rule := range routeModel.ParseRouteRules().Rules {
 
 		// TODO: add the scenarios where we will not create child VS here.
```
swathinsankaran, line 36: We may have to do some experiments for HTTPRoute with filter and without backend or match.

ako-gateway-api/nodes/avi_model_l7_translator.go
```diff
@@ -50,32 +59,89 @@ func (o *AviObjectGraph) BuildChildVS(key string, routeModel RouteModel, parentN
...
+			return
+		}
+		poolName := akogatewayapilib.GetPoolName(parentNs, parentName,
+			routeModel.GetNamespace(), routeModel.GetName(),
+			utils.Stringify(rule.Matches),
+			backend.Namespace, backend.Name, strconv.Itoa(int(backend.Port)))
+		poolNode := &nodes.AviPoolNode{
+			Name:     poolName,
+			Tenant:   lib.GetTenant(),
+			Protocol: listenerProtocol,
```
swathinsankaran, line 122: This protocol must match the protocol in the service. Seems to me that something is wrong here.
@akshayhavile could you share your views?

swathinsankaran: I had a discussion with Akshay. We have to set the health monitors according to the listener protocols. For the time being, the system-TCP is enough and in future, we may have to change the behavior in AKO to set the HM based on the application.

swathinsankaran:
closing this PR as the comments are already taken care of and merged in a separate PR.

PR #1256, opened 2023-08-25
[GatewayAPI]Code to process objects during fullsync and bug fixes

swathinsankaran:
This PR adds the following changes:

  1. Code to process HTTPRoute objects during bootup.
  2. Optimization to process the GatewayAPI objects once during bootup.
  3. Get the object from DeltaFIFO if the delete event is missed.
  4. Bug fixes.


swathinsankaran:
Merging this code as it is required for end-to-end testing.

PR #1262, opened 2023-09-01
AV-188019 : Adding ImagePullSecret support for AKO in Helm Chart

DixitAakash:
This PR adds imagePullSecrets support for AKO in Helm chart. This will enable the user to store the AKO image in their own private registries and use the imagePullSecrets to pull the AKO image on their cluster from their own private registry.

TO DO : Documentation

### DIFF
--- a/docs/values.md
+++ b/docs/values.md
@@ -288,7 +288,7 @@ with the private registry name.
 
 ### image.pullSecrets
 
-If you are setting the [image.repository](#imagerepository) field to use a private container image registry for ako image, then you must specify the pull secrets in this field. The pull secrets are a list of Kubernetes Secret objects that are created from the login credentials of the image registry. The container runtime uses the pull secrets to authenticate with the registry in order to pull the ako image. The image pull secrets must be created in the `avi-system` namespace before deploying AKO.
+If you are setting the [image.repository](#imagerepository) field to use a secure private container image registry for ako image, then you must specify the pull secrets in this field. The pull secrets are a list of Kubernetes Secret objects that are created from the login credentials of the scure private image registry. The container runtime uses the pull secrets to authenticate with the registry in order to pull the ako image. The image pull secrets must be created in the `avi-system` namespace before deploying AKO.
 
     pullSecrets:
     - name: regcred
--- a/docs/values.md
+++ b/docs/values.md
@@ -286,6 +286,13 @@ One AKO runs in active mode, and the second in passive mode. The AKO, which is r
 If you are using a private container registry and you'd like to override the default dockerhub settings, then this field can be edited
 with the private registry name.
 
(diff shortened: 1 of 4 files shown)


DixitAakash:
build ako

helm/ako/templates/statefulset.yaml
```diff
@@ -42,6 +42,12 @@ spec:
         persistentVolumeClaim:
           claimName: {{ .Values.persistentVolumeClaim }}
       {{ end }}
+      {{ if gt (len .Values.image.pullSecrets) 0 }}
+      {{- with .Values.image.pullSecrets }}
```
swathinsankaran, line 46: Is length check enough here?

DixitAakash: @swathinsankaran This condition also works if pullSecrets field is not specified at all. Do we need anything else?

swathinsankaran: I was under the impression that `len(param) > 0` would be enough.

DixitAakash:
build ako

helm/ako/templates/statefulset.yaml
DixitAakash: I have removed both the if condition and the with block. None of those were actually required.

swathinsankaran (dismissed):
LGTM

DixitAakash:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

docs/values.md
```diff
@@ -286,6 +286,13 @@ One AKO runs in active mode, and the second in passive mode. The AKO, which is r
 If you are using a private container registry and you'd like to override the default dockerhub settings, then this field can be edited
 with the private registry name.
 
+### image.pullSecrets
+
+If you are setting the [image.repository](#imagerepository) field to use a private container image registry for ako image, then you must specify the pull secrets in this field. The pull secrets are a list of Kubernetes Secret objects that are created from the login credentials of the image registry. The container runtime uses the pull secrets to authenticate with the registry in order to pull the ako image. The image pull secrets must be created in the `avi-system` namespace before deploying AKO.
```
akshayhavile, line 291: Is this applicable for `insecure-registries` also?

DixitAakash:
build ako

DixitAakash:
build ako

akshayhavile:
AKO PR Sanity Check succeeded

docs/values.md
DixitAakash: It should not be. I will update the doc for only secure private registry.

DixitAakash:
build ako

swathinsankaran (approved):
LGTM

PR #1294, opened 2023-09-18
AV-189284: Fix httprule issue overwriting other httprule setting applied to ingress with same fqdn with different path.

akshayhavile:
In AKO, HTTPrule is namespace specific.  So if customer creates two ingresses say ingress1 and ingress2 with same fqdn.

ingress1 is in namespace default with path /path1 and ingress2 is namespace red with path /path2 . Then customer creates 2 httprule, rule1 and rule2.

rule1 is applied to ingress1/path1 and rule2 is applied to ingress2/path2 . So when rule2 is applied, configuration done by rule1 on pool is getting removed.


akshayhavile:
build ako

arihantg (approved):
LGTM

DixitAakash (approved):
LGTM

PR #1328, opened 2023-11-20
AV-193232 : Fix for regression introduced in 1.11.1 for SNI Dedicated L7 VS while adding SSL Key and Cert Ref using hostrule

DixitAakash:
This PR fixes the regression introduced in AKO 1.11.1 for SNI Dedicated L7 VS while adding SSL Key Cert Ref using hostrule. This issue has been detected as part of https://avinetworks.atlassian.net/browse/AV-193232. This issue was because the same SSL Key Cert Ref was appended twice to the virtual service SSL Key and Cert Refs List in case of SNI Dedicated L7 VS.

PR #1334, opened 2023-11-30
AV-193232 : Fix for regression introduced in 1.11.1 for SNI Dedicated…

DixitAakash:
… L7 VS while adding SSL Key and Cert Ref using hostrule (#1328)

* AV-193232 : Fix for regression introduced in 1.11.1 for SNI Dedicated L7 VS while adding SSL Key and Cert Ref using hostrule

* AV-193232 : Adding UT for fixing issue for SNI Dedicated L7 VS while adding SSL Key and Cert Ref using hostrule

* AV-193232 : Adding SslKeyAndCertificateRefs to checksum calculation for EVH dedicated mode

* AV-193232 : Modified imgae registry for helm-unittest image to private registry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -323,7 +323,7 @@ helmtests:
 	-u root:root \
 	-v $(PWD)/helm/ako:/apps \
 	-v $(PWD)/tests/helmtests:/apps/tests \
-	helmunittest/helm-unittest:3.11.1-0.3.0 .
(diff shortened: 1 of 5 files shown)


pkoshtavmware (approved):
LGTM as cherry-pick

akshayhavile (approved):

akshayhavile:
build ako

PR #1335, opened 2023-12-01
AV-193293: Fix: VSes are getting deleted when there is an issue with an access to the kube_api server resulting in setting DeleteConfig to true

akshayhavile:
Issue details:
When there is an issue with kube_api access, deleteConfig was getting set. This was resulting in deleting all existing VS


akshayhavile:
build ako

akshayhavile:
build ako

arihantg (approved):

PR #1338, opened 2023-12-07
AV-194094 : Cover profile generation for AKO UT's from Makefile for coverage report in Sonarqube

DixitAakash:
This PR adds **--privileged** flag to docker run test commands due to some permission issues seen with the latest docker version installed in the VM where sonarqube jenkins job runs.


DixitAakash:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

Makefile
```diff
@@ -172,162 +172,162 @@ ako-gateway-api-docker: glob-vars
...
-	sudo docker run \
+	sudo docker run --privileged \
 	-w=/go/src/$(PACKAGE_PATH_AKO) \
 	-v $(PWD):/go/src/$(PACKAGE_PATH_AKO) $(BUILD_GO_IMG) \
 	$(GOTEST) -v -mod=vendor $(PACKAGE_PATH_AKO)/tests/k8stest -failfast -coverprofile cover-1.out -coverpkg=./...
 
 .PHONY: integrationtest
 integrationtest:
-	sudo docker run \
+	sudo docker run --privileged \
```
akshayhavile, line 182: @DixitAakash: As per my understanding with --privilage option, AKO container will run with same authority as that of host OS process having access to all underlying resources. Do we want to provide such access level to container? Do we have anyother alternative to it?

DixitAakash: @akshayhavile I had to add this flag due to `Operation not permitted` issue on VM where Sonarqube Jenkins job is run. I added it because I don't see any such concern here as these containers are created with **golang:latest** image just to run the go test command and then stopped.

akshayhavile: At what step we are getting issue? While creating files `cover.out`? or while running the container itself? If its permission issue for copying file, i think it will be good to give permission to those folders.

DixitAakash:
Closed this PR as it no longer needed. Instead of adding --privileged flag to docker run test commands, updated runc and libseccomp in the VM running Sonarscan Jenkins job to solving compatibility issue with latest version of docker.

PR #1342, opened 2023-12-11
AV-194491 : Setting local_as field type in bgp profile of vrf context to uint32 to avoid parsing error

DixitAakash:
This PR is to fix the issue of AKO not creating static routes when a value greater than **2147483647** is specified for **LocalAs** (Local Autonomous System ID) field in Bgp profile. This scenario is applicable only when Bgp profile is specified for VRF Context. The type for **LocalAs** field was **int32** and hence a value greater than **2147483647** was not accepted. Now, the type has been changed to **unit32**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,3 +292,8 @@ All notable changes to this project will be documented in this file. The format 
 
 ### Changed
  - Annotation `external-dns.alpha.kubernetes.io/hostname` on the Service of type LoadBalancer overrides the `autoFQDN` feature for it.
+
+## AKO-1.7.6
+
+### Fixed
+ - AKO does not create static routes when a value greater than **2147483647** is specified for **LocalAs** (Local Autonomous System ID) field in Bgp profile. This scenario is applicable only when Bgp profile is specified for VRF Context.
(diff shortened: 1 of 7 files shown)


akshayhavile (commented):
Should we implement `null` check to avoid crash and just log the message of field being null?

DixitAakash:
> Should we implement `null` check to avoid crash and just log the message of field being null?

@akshayhavile can be done but is it required for this patch release?

DixitAakash:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

akshayhavile:
> > Should we implement `null` check to avoid crash and just log the message of field being null?
> 
> @akshayhavile I believe crash due to nil check is a different issue. Here, it is vrfcontext parsing issue due to incorrect data type generated as part of SDK for local_as field in bgp_profile under vrfcontext. AKO is not crashing but just not creating static routes as it is unable to parse vrfcontext and create cache.

Yup. My bad. I got confused with other issue.

akshayhavile (approved):

DixitAakash:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

PR #1366, opened 2024-01-23
AV-197474: VRF support in vcenter cloud

akshayhavile:
Following is manual testing done:
1. NodePort mode:
   a. If VRF is set in configmap, use that else use Global VRF

2. ClusterIP mode for Vcenter cloud only
   a. Test VRF is valid or not.
   b. Test VIP network, mentioned in Configmap, is under the VRF, mentioned in configmap, or not.
   c. Test Node network list (used for pool placement network) is under the VRF or not.
   d. Create a new VRF in admin tenant and check all objects are getting created under given vrf including static routes.
   e. Use `test` tenant and VRF `global` present under `admin tenant` to check all objects are getting created under given tenant and global vrf or not.

UT Pending: Need to check how to do UT for this scenario

### DIFF
--- a/cmd/ako-main/main.go
+++ b/cmd/ako-main/main.go
@@ -108,6 +108,7 @@ func InitializeAKC() {
 	}
 	akoControlConfig.SetAKOInstanceFlag(isPrimaryAKO)
 	akoControlConfig.SetAKOBlockedNSList(lib.GetGlobalBlockedNSList())
+	akoControlConfig.SetControllerVRFContext(lib.GetControllerVRFContext())
 	var crdClient *crd.Clientset
 	var advl4Client *advl4.Clientset
 	var svcAPIClient *svcapi.Clientset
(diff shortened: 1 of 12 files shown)


apalsule:
The VS, VSVIP, Pool objects are associated with the VRF, reference did not see where you are doing that.

Where are you choosing the right VRF for static routes ?

akshayhavile:
> The VS, VSVIP, Pool objects are associated with the VRF, reference did not see where you are doing that.
> 
> Where are you choosing the right VRF for static routes ?

@apalsule: For each of these objects, VRF is being set from `lib.GetVRF()` function. Example of set is as shown here: https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/internal/nodes/avi_model_l7_translator.go#L103
This `GetVRF()` returns vrfContext variable which is set during AKO bootup after validation in `checkAndSetVRFFromNetwork()` function

akshayhavile:
build ako

akshayhavile:
build ako

DixitAakash (approved):
LGTM

abhishekbsingh (approved):

akshayhavile:
build ako

DixitAakash (approved):

akshayhavile:
build ako

PR #1368, opened 2024-01-30
AV-196436 & AV-196437 : Removing model saving after child deletion to allow model saving and publishing in DequeueIngestion for new child creation

DixitAakash:
This PR is for removing model saving after each child deletion to allow model saving and publishing once in DequeueIngestion. This will delete the old child VS and create the new child VS. When the matches section is updated in HTTPRoute, a new child vs should be created.


DixitAakash:
build ako

pkoshtavmware (approved):
LGTM

akshayhavile (approved):

PR #1384, opened 2024-03-07
AV-201231 : Allow AKO to use existing avi-secret in avi-system namespace

DixitAakash:
This PR is to allow AKO to use existing avi-secret in avi-system namespace. If username and either of password or authtoken are not specified in Helm chart values.yaml, avi-secret will not be created as part of Helm installation. AKO will assume that avi-secret already exists in avi-system namespace and will reference it.

### DIFF
--- a/docs/values.md
+++ b/docs/values.md
@@ -281,6 +281,8 @@ authtoken = "<authtoken>"
 print(base64.b64encode(authtoken.encode("ascii")))
 ```
 
+**Note:** From release v1.12.1 onwards, AKO supports reading Avi controller credentials from existing `avi-secret` in `avi-system` namespace. If username and either of password or authtoken are not specified, avi-secret will not be created as part of Helm installation. AKO will assume that avi-secret already exists in avi-system namespace and will reference it.
+
 ### avicredentials.certificateAuthorityData
 
 This field allows setting the rootCA of the Avi controller, that AKO uses to verify the server certificate provided by the Avi Controller during the TLS handshake. This also enables AKO to connect securely over SSL with the Avi Controller, which is not possible in case the field is not provided.
--- a/docs/values.md
+++ b/docs/values.md
@@ -281,7 +281,7 @@ authtoken = "<authtoken>"
 print(base64.b64encode(authtoken.encode("ascii")))
 ```
 
-**Note:** From release v1.12.1 onwards, AKO supports reading Avi Controller credentials from existing `avi-secret` from the namespace in which AKO is installed. If username and either password or authtoken are not specified, avi-secret will not be created as part of Helm installation. AKO will assume that avi-secret already exists in the namespace in which the AKO Helm release is installed and will reference it.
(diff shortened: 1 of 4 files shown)


DixitAakash:
build ako

saurabhvaidvmware:
@DixitAakash the IF condition looks good. What happens if avi-secret already exists and someone also specifies username and password/authtoken? In k8s it gives `AlreadyExists` error, will that effect anything during the AKO installation?

DixitAakash:
> @DixitAakash the IF condition looks good. What happens if avi-secret already exists and someone also specifies username and password/authtoken? In k8s creating already existing secret again gives `AlreadyExists` error, will that effect anything during the AKO installation?

@saurabhvaidvmware the helm installation will not go through and result in an error.

saurabhvaidvmware:
build ako

saurabhvaidvmware:
build ako

saurabhvaidvmware (approved):
LGTM

docs/values.md
```diff
@@ -281,6 +281,8 @@ authtoken = "<authtoken>"
 print(base64.b64encode(authtoken.encode("ascii")))
 ```
 
+**Note:** From release v1.12.1 onwards, AKO supports reading Avi controller credentials from existing `avi-secret` in `avi-system` namespace. If username and either of password or authtoken are not specified, avi-secret will not be created as part of Helm installation. AKO will assume that avi-secret already exists in avi-system namespace and will reference it.
```
akshayhavile, line 284: We should say `ako installed namespace`. Customer might install AKO in different namespace.

akshayhavile: Did we test HA deployment?

akshayhavile: Do we know it can be used as part of AKO operator installed AKO?

DixitAakash: @akshayhavile I have updated it to `ako installed namespace`. HA deployment should not be an issue since it is just one more replica which will use the same secret. Also, for AKO Operator, we are required to create avi-secret before applying akoconfig object to create ako deployment. Currently, we only support avi-system namespace with ako-operator.

DixitAakash:
build ako

helm/ako/values.yaml
```diff
@@ -127,7 +127,7 @@ rbac:
   # Creates the pod security policy if set to true
   pspEnable: false
 
-
+# If username and either password or authtoken are not specified, avi-secret will not be created. AKO will assume that the avi-secret already exists and will reference it.
```
akshayhavile, line 130: I think it will be good to mention this: `In such scenario(where username and pass/authtoken empty),  value of `certificateAuthorityData` will be taken from existing `avi-secret`.
Please re-word as per your convenience.

akshayhavile (approved):
Small comment. Else looks good.

DixitAakash:
build ako

DixitAakash:
build ako

helm/ako/values.yaml
DixitAakash: I have updated the PR with these changes.

akshayhavile (approved):

arihantg (approved):
LGTM

PR #1390, opened 2024-03-24
AV-202495: Fix Panic when avi controller returns error during object creation.

akshayhavile:
…


akshayhavile:
build ako

DixitAakash (approved):
LGTM

arihantg (approved):
Looks good from AKO perspective. For gateway we need to check the object type before typecasting the interface. I will add that

PR #1391, opened 2024-03-26
AV-202495 check object type before the typecast

arihantg:
Manually tested with a expired certificate in secure ingress .

### DIFF
--- a/internal/rest/dequeue_nodes.go
+++ b/internal/rest/dequeue_nodes.go
@@ -699,7 +699,15 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
 
 func updateGatewayStatusWithVsError(key string, rest_op *utils.RestOp) {
 	if rest_op.Model == "VirtualService" {
-		vs := rest_op.Obj.(avimodels.VirtualService)
+		var vs avimodels.VirtualService
+		switch rest_op.Obj.(type) {
+		case avimodels.VirtualService:
+			vs = rest_op.Obj.(avimodels.VirtualService)
+		case *avimodels.VirtualService:
+			vs = *(rest_op.Obj.(*avimodels.VirtualService))
+		default:
+			return
+		}
 		if vs.Type != nil && *vs.Type == utils.VS_TYPE_VH_PARENT {
 			if vs.ServiceMetadata != nil {
 				var svc_mdata_obj lib.ServiceMetadataObj


arihantg:
build ako

arihantg:
build ako

internal/rest/dequeue_nodes.go
```diff
@@ -699,7 +699,15 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
 
 func updateGatewayStatusWithVsError(key string, rest_op *utils.RestOp) {
 	if rest_op.Model == "VirtualService" {
-		vs := rest_op.Obj.(avimodels.VirtualService)
+		var vs avimodels.VirtualService
+		switch rest_op.Obj.(type) {
+		case avimodels.VirtualService:
+			vs = rest_op.Obj.(avimodels.VirtualService)
```
pkoshtavmware, line 705: Please describe in which case object type will be *avimodels.VirtualService and avimodels.VirtualService ?

pkoshtavmware:
> Manually tested with a expired certificate in secure ingress .

Isn't this function used for gateway only ?

internal/rest/dequeue_nodes.go
arihantg: in case of child vs it will be *avimodels.VirtualService while in parent vs it is avimodels.VirtualService

arihantg:
> > Manually tested with a expired certificate in secure ingress .
> 
> Isn't this function used for gateway only ?

It is the case after the merge of https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/1390. I just tested without pulling this changeset.

pkoshtavmware (approved):
LGTM

DixitAakash (approved):
LGTM

internal/rest/dequeue_nodes.go
```diff
@@ -699,7 +699,15 @@ func (rest *RestOperations) ExecuteRestAndPopulateCache(rest_ops []*utils.RestOp
...
+		case *avimodels.VirtualService:
```
akshayhavile, line 706: As logic at line number 711 is for Parent VS only, should we return from line number 707? no need to check type for child vs.. right?

arihantg: let's keep it as defensive check in case there is unexplored scenario for parentvs  where type can be pointer.

pkoshtavmware (approved):

PR #1412, opened 2024-04-23
AV-205437 Static Route fixes

pkoshtavmware:
AV-205437 Static Route fixes
This PR fixes index outofbound error causing panic if indexes are not updated properly.
Adds a unit testcase to simulate the panic
Upgrades AKO version from 1.10.3 to 1.10.4

### DIFF
--- a/internal/nodes/avi_vrf_translator.go
+++ b/internal/nodes/avi_vrf_translator.go
@@ -92,42 +92,60 @@ func (o *AviObjectGraph) BuildVRFGraph(key, vrfName, nodeName string, deleteFlag
 			// update case
 			// Assumption: updated routes (values) for given node will not overlap with other nodes.
 			// So only updating existing routes of that node.
+			utils.AviLog.Debugf("key: %s, StaticRoutes before updation/deletion: [%v]", key, utils.Stringify(aviVrfNode.StaticRoutes))
 			startIndex := nodeStaticRouteDetails.StartIndex
 			lenNewNodeRoutes := len(nodeRoutes)
 			diff := lenNewNodeRoutes - nodeStaticRouteDetails.Count
 
 			var staticRouteCopy []*models.StaticRoute
-			staticRouteCopy = append(staticRouteCopy, aviVrfNode.StaticRoutes[:startIndex]...)
+			copyTill := int(math.Min(float64(startIndex), float64(len(aviVrfNode.StaticRoutes))))
+			staticRouteCopy = append(staticRouteCopy, aviVrfNode.StaticRoutes[:copyTill]...)
+
 			staticRouteCopy = append(staticRouteCopy, nodeRoutes...)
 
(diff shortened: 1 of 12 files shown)


internal/nodes/avi_vrf_translator.go
```diff
@@ -97,10 +97,13 @@ func (o *AviObjectGraph) BuildVRFGraph(key, vrfName, nodeName string, deleteFlag
 			diff := lenNewNodeRoutes - nodeStaticRouteDetails.Count
 
 			var staticRouteCopy []*models.StaticRoute
-			staticRouteCopy = append(staticRouteCopy, aviVrfNode.StaticRoutes[:startIndex]...)
+			if startIndex <= len(aviVrfNode.StaticRoutes) {
```
akshayhavile, line 100: Can `startIndex` be negative?

akshayhavile: Should we have some logs to indicate record is not added?

akshayhavile (commented):
@pkoshtavmware Can we please add unit test case which was failing in AKO 1.10.3 but was successful with your changes?

pkoshtavmware:
> @pkoshtavmware Can we please add unit test case which was failing in AKO 1.10.3 but was successful with your changes?

Done

internal/nodes/avi_vrf_translator.go
pkoshtavmware: We can add a message showing index is either -ve or out of length but not sure if we would be able to say specifically this record is not added, since if we would try to obtaibe value at that index, it will again get into panic state

pkoshtavmware: Done

akshayhavile (approved):

akshayhavile (approved):

akshayhavile:
build ako

tests/cnitests/static_route_test.go
```diff
@@ -48,7 +48,7 @@ var v1alpha2CRDClient *v1alpha2crdfake.Clientset
 var ctrl *k8s.AviController
 var DynamicClient *dynamicfake.FakeDynamicClient
 
-var cniPlugin = flag.String("cniPlugin", "", "cni plugin for the setup")
+var cniPlugin = flag.String("cniPlugin", "calico", "cni plugin for the setup")
```
akshayhavile, line 51: This will fail other CNI

DixitAakash (approved):
LGTM

DixitAakash:
build ako

DixitAakash (dismissed):

akshayhavile (dismissed):

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile (approved):

PR #1419, opened 2024-05-02
AV-205437 Static Route fix

pkoshtavmware:
AV-205437 Static Route fix
This PR fixes following issue with Static routes:
If we have multiple nodes with multiple PODCidrs, if we delete all block affinities on any of intermediate node, the NodeId and NodeStaticRoute arrays and nodes gets deleted which results in a continuous route_ids(since we have handled it in delete and update case). However, if we delete all static routes from controller and reboot ako, the Node, NodeIds and NodeStaticRoute gets added even if number of block affinities are 0 for that node(Since we had not handled addition case earlier). As a part of the fix, if number of static route/PODCidrs attached to the Node are zero, we will not add Node, NodeIds and NodeStaticRoutes to the VRF node

### DIFF
--- a/internal/nodes/avi_vrf_translator.go
+++ b/internal/nodes/avi_vrf_translator.go
@@ -77,7 +77,7 @@ func (o *AviObjectGraph) BuildVRFGraph(key, vrfName, nodeName string, deleteFlag
 		}
 		if !ok {
 			//node not found, check overlapping and then add case
-			if !findRoutePrefix(nodeRoutes, aviVrfNode.StaticRoutes, key) {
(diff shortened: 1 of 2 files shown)


DixitAakash (approved):
LGTM

akshayhavile (approved):
@pkoshtavmware changes looks good. Can we add UT to check sequence of route ids?

akshayhavile (approved):

akshayhavile:
build ako

DixitAakash (approved):

pkoshtavmware:
build ako

akshayhavile:
build ako

PR #1424, opened 2024-05-07
AV-205976 : disabling status update for vs and pool from follower

DixitAakash:
This PR addresses the issue of Follower AKO keeps on restarting after setup creation. The follower was getting OOMKilled due to status update in happening in bulk. The status update has been disabled for follower as it was only processing the status update operations but ultimately not doing anything.

### DIFF
--- a/internal/rest/avi_obj_vs.go
+++ b/internal/rest/avi_obj_vs.go
@@ -486,6 +486,10 @@ func AviVsHttpPSAdd(vs_meta interface{}, isEVH bool) []*avimodels.HTTPPolicies {
 }
 
 func (rest *RestOperations) StatusUpdateForPool(restMethod utils.RestMethod, vs_cache_obj *avicache.AviVsCache, key string) {
+	if !lib.AKOControlConfig().IsLeader() {
+		utils.AviLog.Debugf("key: %s, AKO is not running as a leader, will not publish the status", key)
+		return
+	}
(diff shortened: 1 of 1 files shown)


akshayhavile (approved):

akshayhavile:
build ako

arihantg (approved):

pkoshtavmware:
Should we remove code to process this from Status layer(Consumer) as well since that is anyways a dead/non reachable code ?

DixitAakash:
> Should we remove code to process this from Status layer(Consumer) as well since that is anyways a dead/non reachable code in case of follower ?

@pkoshtavmware can be done but I think we can do it later as it is mostly one method that is unused. Also, we need to run the FT's and other tests against this so have not removed any code now.

pkoshtavmware:
> > Should we remove code to process this from Status layer(Consumer) as well since that is anyways a dead/non reachable code in case of follower ?
> 
> @pkoshtavmware can be done but I think we can do it later as it is mostly one method that is unused. Also, we need to run the FT's and other tests against this so have not removed any code now.

There are multiple methods(Update/Delete) for every K8s object type. However, its okay to remove them after release.

akshayhavile:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

PR #1434, opened 2024-05-16
AV-206565 Fix for AKO GW crash when HttpRoute attaching to one existing and one non existing gateway

pkoshtavmware:
AV-206565 Fix for AKO GW crash when HttpRoute attaching to one existing and one non existing gateway.

This PR fixes ako gw crashes we were seeing due to index mismatch in httpRoute->Status->Parent validation.

**Reason for crash**: In case of non-existent gateway or gateway having gateway class with a different controller, we are not setting anything in httpRoute status corresponding to this parent which is correct as per Gateway API Spec, however when we were trying to set something at next index in httpRoute->Status->Parent for the next valid parent i.e. at 1 index a panic was seen leading to ako gw container crash since we were adding entry at 0th index and trying to update it at index 1.

**Testing Done:**
Added a Unit Test Case simulating the same scenario.

### DIFF
--- a/ako-gateway-api/k8s/validator.go
+++ b/ako-gateway-api/k8s/validator.go
@@ -276,11 +276,12 @@ func IsHTTPRouteValid(key string, obj *gatewayv1.HTTPRoute) bool {
 	httpRouteStatus := obj.Status.DeepCopy()
 	httpRouteStatus.Parents = make([]gatewayv1.RouteParentStatus, 0, len(httpRoute.Spec.ParentRefs))
 	var invalidParentRefCount int
-	for index := range httpRoute.Spec.ParentRefs {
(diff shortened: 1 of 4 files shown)


pkoshtavmware:
build ako

DixitAakash (dismissed):
LGTM

pkoshtavmware:
build ako

pkoshtavmware:
build ako

DixitAakash (approved):

arihantg (approved):
LGTM

akshayhavile:
build ako

akshayhavile:
build ako

pkoshtavmware:
build ako

PR #1435, opened 2024-05-17
AV-205821 add external httppolicyset in cache

arihantg:
At bootup AKO was making calls to populate httppolicy set in cache for each VS if it was finding missing match in cache for any external policyset.  With this fix we will fetch the missing policyset by uuid and add it to cache.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -1624,6 +1624,62 @@ func (c *AviObjCache) AviPopulateAllHttpPolicySets(client *clients.AviClient, cl
 	}
 	return httpPolicyData, result.Count, nil
 }
+func (c *AviObjCache) AviPopulateHttpPolicySetbyUUID(client *clients.AviClient, uuid string) error {
+
+	uri := "/api/httppolicyset/" + uuid
+	rawData, err := lib.AviGetRaw(client, uri)
+	if err != nil {
+		utils.AviLog.Warnf("Get uri %v returned err for httppolicyset %v", uri, err)
+		return err
+	}
+	httppol := models.HTTPPolicySet{}
+	err = json.Unmarshal(rawData, &httppol)
+	if err != nil {
+		utils.AviLog.Warnf("Failed to unmarshal httppolicyset data, err: %v", err)
+		return err
+	}
+
+	if httppol.Name == nil || httppol.UUID == nil {
+		utils.AviLog.Warnf("Incomplete http policy data unmarshalled, %s", utils.Stringify(httppol))
+		return errors.New("incomplete http policy data unmarshalled")
+	}
+
+	// Fetch the pgs associated with the http policyset object
+	var poolGroups []string
+	var pools []string
+	if httppol.HTTPRequestPolicy != nil {
+		for _, rule := range httppol.HTTPRequestPolicy.Rules {
+			if rule.SwitchingAction != nil {
+				val := reflect.ValueOf(rule.SwitchingAction)
+				if !val.Elem().FieldByName("PoolGroupRef").IsNil() {
+					pgUuid := ExtractUuid(*rule.SwitchingAction.PoolGroupRef, "poolgroup-.*.#")
+					pgName, found := c.PgCache.AviCacheGetNameByUuid(pgUuid)
+					if found {
+						poolGroups = append(poolGroups, pgName.(string))
+					}
+				} else if !val.Elem().FieldByName("PoolRef").IsNil() {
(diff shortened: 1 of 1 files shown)


arihantg:
build ako

arihantg:
build ako

internal/cache/controller_obj_cache.go
```diff
@@ -2006,8 +2062,11 @@ func (c *AviObjCache) AviObjVSCachePopulate(client *clients.AviClient, cloud str
 								c.PopulateHttpPolicySetToCache(client, cloud)
```
akshayhavile, line 2062: @arihantg: If we look at `PopulateHttpPolicySetToCache` function, it fetches all httpps which is created by AKO user and populate the cache. And then we retrieve out httpps by uuid at line number 2063. I think fetching all httpps each time is not required here (at line number 2062) as we are going through httpps attached to each VS. 

Instead of this call, we can call your function `AviPopulateHttpPolicySetbyUUID` here. So that it will fetch it from uuid.

All HTTPPS cache update is already happened as part of `AviRefreshObjectCache`

arihantg: I have replaced the 2nd call to PopulateHttpPolicySetToCache with PopulateHttpPolicySetToCache.

internal/cache/controller_obj_cache.go
```diff
@@ -1624,6 +1624,62 @@ func (c *AviObjCache) AviPopulateAllHttpPolicySets(client *clients.AviClient, cl
...
+		for _, rule := range httppol.HTTPRequestPolicy.Rules {
+			if rule.SwitchingAction != nil {
+				val := reflect.ValueOf(rule.SwitchingAction)
+				if !val.Elem().FieldByName("PoolGroupRef").IsNil() {
+					pgUuid := ExtractUuid(*rule.SwitchingAction.PoolGroupRef, "poolgroup-.*.#")
+					pgName, found := c.PgCache.AviCacheGetNameByUuid(pgUuid)
+					if found {
+						poolGroups = append(poolGroups, pgName.(string))
+					}
+				} else if !val.Elem().FieldByName("PoolRef").IsNil() {
```
pkoshtavmware, line 1660: Why are we populating either pools or poolgroup, not both ?

internal/cache/controller_obj_cache.go
```diff
@@ -2000,14 +2055,13 @@ func (c *AviObjCache) AviObjVSCachePopulate(client *clients.AviClient, cloud str
...
-								// We do a full refresh of the httpcache once per page, if we detect a data discrepancy
-								httpCacheRefreshCount = httpCacheRefreshCount - 1
-								c.PopulateHttpPolicySetToCache(client, cloud)
-								httpName, foundhttp = c.HTTPPolicyCache.AviCacheGetNameByUuid(httpUuid)
-								if !foundhttp {
-									// If still the httpName is not found. Log an error saying, this VS may not behave appropriately.
+							if !foundhttp && !sharedVsOrL4 {
+								err = c.AviPopulateHttpPolicySetbyUUID(client, httpUuid)
+								// If still the httpName is not found. Log an error saying, this VS may not behave appropriately.
+								if err != nil {
```
pkoshtavmware, line 2061: Should we not recalculate values of `httpName` and `foundhttp` after cache population here ?

arihantg: values won't change in error condition. In else condition it is handled.

pkoshtavmware: but after error we are populating httppolicyset by calling this function , `AviPopulateHttpPolicySetbyUUID`, will it not change values for httpName, foundhttp

internal/cache/controller_obj_cache.go
akshayhavile: @pkoshtavmware:In Avi controller, switching action can point to  either pool or PG.

arihantg: It is because SwitchingAction can be either to select HTTP_SWITCHING_SELECT_POOLGROUP or HTTP_SWITCHING_SELECT_POOL not both

pkoshtavmware (approved):
LGTM

arihantg:
build ako

arihantg:
build ako

arihantg:
build ako

arihantg:
build ako

akshayhavile (approved):

arihantg:
build ako

arihantg:
build ako

PR #1461, opened 2024-06-25
AV-209807 : Adding UT to test AKO now honours readines probe for pods when Antrea with NPL is used

DixitAakash:
Thsi PR adds a UT to test that AKO now honours readines probe for pods when Antrea with NPL is used for LoadBalancer service. Previously UT was only added for ingress. Also, small changes have been made in the previously added UT to fix some issues.


DixitAakash:
build ako

akshayhavile (approved):

DixitAakash:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

PR #1474, opened 2024-07-03
AV-210878: Increase API timeout to 200 seconds

akshayhavile:


pkg/utils/avi_rest_utils.go
```diff
@@ -45,6 +46,7 @@ func NewAviRestClientPool(num uint32, api_ep, username,
 	options := []func(*session.AviSession) error{
 		session.SetNoControllerStatusCheck,
 		session.SetTransport(transport),
+		session.SetTimeout(200 * time.Second),
```
abhishekbsingh, line 49: nit: increasing the timeout here should be good enough as other clients do not get used in SSL cert creation in Avi.

akshayhavile: Currently, we only know sslcert is taking time. To secure other API calls and paths which is not covered by this flow, we are fending it with this timeout. There are bunch of places, we directly create session object and connect to controller.

akshayhavile:
build ako

abhishekbsingh (approved):

akshayhavile:
build ako

DixitAakash (approved):

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

PR #1486, opened 2024-08-01
AV-213944 Status update concurrency fix

aaha97:
fix incorrect status update due to concurrency issue.

the issue was found when running the UT ingresstests::TestMultiHostIngressStatusCheck

concurrent threads were overwriting valid IP in status because updates were not atomic in nature and both referred to the same old ingress objects.

this is fixed in this PR by adding locks before status updates when same object is being updated by multiple threads and fetching the latest status before proceeding with the update

### DIFF
--- a/internal/lib/utils.go
+++ b/internal/lib/utils.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/cache"
@@ -545,4 +546,27 @@ func IsInfraSettingNSScoped(infraSetting, namespace string) bool {
 		}
 	}
 	return false
+}
+
+type LockSet struct {
+	locks sync.Map
+}
+
+var lockSet LockSet
+
+func (s *LockSet) Lock(lockName string) {
+	lock, _ := s.locks.LoadOrStore(lockName, &sync.Mutex{})
+	lock.(*sync.Mutex).Lock()
+}
+
+func (s *LockSet) Unlock(lockName string) {
+	if lock, ok := s.locks.Load(lockName); !ok {
+		panic("unlocked an unlock mutex")
+	} else {
+		lock.(*sync.Mutex).Unlock()
+	}
+}
+
+func GetLockSet() *LockSet {
+	return &lockSet
 }
--- a/internal/status/ing_status.go
+++ b/internal/status/ing_status.go
@@ -149,6 +149,19 @@ func updateObject(mIngress *networkingv1.Ingress, updateOption UpdateOptions, re
 	var updatedIng *networkingv1.Ingress
 	var err error
 	if !sameStatus {
+		ingressNsName := mIngress.Namespace + "/" + mIngress.Name
+		//lock here to avoid concurrent updates to same status
+		lib.GetLockSet().Lock(ingressNsName)
+		latestIngress := getIngresses([]string{ingressNsName}, false)
+		if latestIngress[ingressNsName] != nil {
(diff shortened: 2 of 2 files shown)


Dhivyaaj:

aaha97:
build ako

aaha97:
build ako

internal/status/ing_status.go
```diff
@@ -149,6 +149,19 @@ func updateObject(mIngress *networkingv1.Ingress, updateOption UpdateOptions, re
 	var updatedIng *networkingv1.Ingress
 	var err error
 	if !sameStatus {
+		ingressNsName := mIngress.Namespace + "/" + mIngress.Name
+		//lock here to avoid concurrent updates to same status
+		lib.GetLockSet().Lock(ingressNsName)
```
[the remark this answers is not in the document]
akshayhavile: ` getIngresses` is being called twice at 155 and 173. Is it possible to make it one call?

internal/lib/utils.go
```diff
@@ -546,3 +547,31 @@ func IsInfraSettingNSScoped(infraSetting, namespace string) bool {
 	}
 	return false
 }
+
+type LockSet struct {
+	locks map[string]*sync.Mutex
```
abhishekbsingh, line 552: we can use sync.map here and do away with explicit lock/unlock logic.

aaha97: all threads are required to perform atomic updates. sync.map will not work for this use case

abhishekbsingh: ok, in that case let's use sync.map instead of regular map which would not be safe for concurrent write/read

aaha97: replaced regular golang map with sync.Map, thanks

aaha97:
build ako

internal/lib/utils.go
```diff
@@ -546,3 +547,26 @@ func IsInfraSettingNSScoped(infraSetting, namespace string) bool {
...
+var lockSet LockSet
+
+func (s *LockSet) Lock(lockName string) {
+	lock, _ := s.locks.LoadOrStore(lockName, &sync.Mutex{})
+	lock.(*sync.Mutex).Lock()
+}
+
+func (s *LockSet) Unlock(lockName string) {
+	if lock, ok := s.locks.Load(lockName); !ok {
+		panic("unlocked an unlock mutex")
```
abhishekbsingh, line 564: If the lockName(ingress) is not there is the map - shouldn't the unlock operation be assumed successful? 
Why panic in that case?

aaha97: the map will only get populated if the lock for ingress has been called atleast once. this panic should ideally never occur unless there is a code bug. 

we are never deleting the keys in the map, so the keys remain until AKO reboots.

aaha97:
build ako

aaha97:
build ako

aaha97:
build ako

aaha97:
build ako

ashishnayak96:
build ako

ashishnayak96:
build ako

akshayhavile:
build ako

akshayhavile (approved):

aaha97:
build ako

abhishekbsingh (approved):

aaha97:
build ako

aaha97:
build ako

aaha97:
build ako

akshayhavile:
@aaha97 Can you please paste local UT run for this?

aaha97:
build ako

aaha97:
build ako

aaha97:
UT results for ingresstests

ok      github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/tests/ingresstests  1102.513s

PR #1499, opened 2024-08-14
AV-212214 Add URI encoding to rest calls (cherry-pick)

aaha97:


ashishnayak96 (approved):
LGTM

arihantg (approved):

akshayhavile:
build ako

pkg/utils/utils.go
```diff
@@ -622,3 +622,10 @@ func String(s *string) string {
 	}
 	return ""
 }
+
+func GetUriEncoded(uri string) string {
+	if uriSplit := strings.SplitN(uri, "?", 2); len(uriSplit) == 2 {
+		return uriSplit[0] + "?" + url.QueryEscape(uriSplit[1])
```
sunjayBhatia, line 628: this change is not quite right, [url.QueryEscape](https://pkg.go.dev/net/url#QueryEscape) should not be used to escape the whole query string instead it should be used to just escape the value part of each individual query parameter set

in a deployment of ako with this code, we see the following log:

```
lib/avi_api.go:43       msg: Unable to fetch collection data from uri /api/tenant/?name=admin Encountered an error on GET request to URL https://10.83.113.185//api/tenant/?name%3Dadmin: HTTP code: 400; error from Controller: map[error:Invalid Query]
```
the `=` in the `?name=admin` query is getting escaped and should not be

sunjayBhatia: cc @akshayhavile

PR #1500, opened 2024-08-15
AV-214519: AKO GW pod crashed with panic : interface {} is nil, not [……]objects.GatewayListenerStore

akshayhavile:
TODO: UT Pending

### DIFF
--- a/ako-gateway-api/objects/gateway_store.go
+++ b/ako-gateway-api/objects/gateway_store.go
@@ -206,8 +206,10 @@ func (g *GWLister) GetGatewayToListeners(gwNsName string) []GatewayListenerStore
 	defer g.gwLock.RUnlock()
 
 	_, listenerList := g.gatewayToListenerStore.Get(gwNsName)
-	return listenerList.([]GatewayListenerStore)
-
+	if listenerList != nil {
+		return listenerList.([]GatewayListenerStore)
+	}
+	return nil
 }
 
 //=====All route <-> gateway mappings go here.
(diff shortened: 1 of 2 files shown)


akshayhavile:
build ako

ako-gateway-api/objects/gateway_store.go
```diff
@@ -206,8 +206,10 @@ func (g *GWLister) GetGatewayToListeners(gwNsName string) []GatewayListenerStore
 	defer g.gwLock.RUnlock()
 
 	_, listenerList := g.gatewayToListenerStore.Get(gwNsName)
-	return listenerList.([]GatewayListenerStore)
-
+	if listenerList != nil {
```
pkoshtavmware, line 209: In master branch this code block is replaced by
```
`found, listenerList := g.gatewayToListenerStore.Get(gwNsName)
	if found {
		return listenerList.([]GatewayListenerStore)
	} else {
		return nil
	}`,
```
, to avoid having different codes across versions doing same thing, please change it

akshayhavile: In above code block, there is no requirement of introducing `else` block as logic is returning from if block itself.  So I will recommend to remove unnecessary `else` block from code base in master.

pkoshtavmware: Should we make if block same atleast ?

akshayhavile: `found` variable is of no use for us as we are returning `nil` or `non nil` value based on value present in `listenerlist`. Why to allocate one more variable if it is not being used.

DixitAakash (approved):
LGTM

PR #1513, opened 2024-08-27
AV-217108 Fix for Hostmap cache not cleared in delete case for Insecure EVH

aaha97:


internal/nodes/avi_model_evh_nodes.go
```diff
@@ -2099,6 +2099,12 @@ func RouteIngrDeletePoolsByHostnameForEvh(routeIgrObj RouteIngressModel, namespa
 		}
 
 		SharedHostNameLister().DeleteNamespace(host)
+		if found, ingressHostMap := SharedHostNameLister().Get(host); found {
```
akshayhavile, line 2102: @aaha97 Similar problem should exists for SNI insecure ingress deletion. Did we face issue over there?

internal/nodes/avi_model_evh_nodes.go
```diff
@@ -2099,6 +2099,12 @@ func RouteIngrDeletePoolsByHostnameForEvh(routeIgrObj RouteIngressModel, namespa
...
+			mapkey := namespace + "/" + objname
+			if utils.HasElem(ingressHostMap.GetIngressesForHostName(), mapkey) {
+				delete(ingressHostMap.HostNameMap, mapkey)
+			}
```
akshayhavile, line 2106: Once value is delete from map. We should save the updated map for that host.

internal/nodes/avi_model_evh_nodes.go
aaha97: this was found when running evhtests. might find the issue in SNI UTs.

internal/nodes/avi_model_evh_nodes.go
aaha97: this update to the map should be in-place as maps are passed by reference

aaha97:
build ako

aaha97:
build ako

internal/nodes/avi_model_evh_nodes.go
aaha97: added the fix for SNI case

aaha97:
build ako

internal/nodes/avi_model_evh_nodes.go
akshayhavile: Do we need to guard delete of map with if condition?

aaha97:
adding relevant UT results

ok      github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/tests/ingresstests  1014.488s

ok      github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/tests/evhtests  1589.397s

ok      github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/tests/evhtests      1593.173s (vipPerNs)

internal/nodes/avi_model_evh_nodes.go
aaha97: not required, removed the if condition

aaha97:
build ako

akshayhavile (approved):
LGTM

akshayhavile:
build ako

DixitAakash (approved):
LGTM

PR #1518, opened 2024-08-30
AV-216533 : Revert encoding uri changes

DixitAakash:


DixitAakash:
build ako

akshayhavile:
@DixitAakash Can we please include any UTs or FTs from our end to check that we are getting expected behaviour? Can we please paste the UT run results?

PR #1536, opened 2024-09-11
Add dual-stack support for NPL and NodePort

### LINKED ISSUE #1535: NodePort mode does not fully support dual-stack networking
### Describe the bug

Currently there is no full support for full dual-stack in AKO. For ingress-services running in ClusterIP mode it is fully working. Running AKO in NodePort-mode does not currently support dual-stack for backend. But it does create a vip with both IPv4 and IPv6. But the IPv6 address is unusable without any IPv6 backend servers.

I believe this is due to the logic found in this file: [internal/nodes/avi_model_l4_translator.go](https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/master/internal/nodes/avi_model_l4_translator.go) and functions PopulateServersForNPL, PopulateServersForNodePort and PopulateServers.

In the following code-snippet for PopulateServersForNodePort it will always populate the poolMeta with either an IPv4
or IPv6 address due to using if/else if. It will never return both an IPv4 and IPv6 address that is intended when running in dual-stack mode.

I don't see any reason why it could not be 2 if-statements where both addresses gets appended to the poolMeta. I will try to make a fix myself and try it in a test-cluster if there are any issues with this approach.

[avi_model_l4_translator.go #L457](https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/blob/ba9256b668268e5bc342c521f95d6668bd34ce51/internal/nodes/avi_model_l4_translator.go#L457)
```golang
if v4enabled && v4Family && nodeIP != "" {
	v4ServerCount++
	atype = "V4"
	serverIP = avimodels.IPAddr{Type: &atype, Addr: &nodeIP}
} else if v6enabled && v6Family && nodeIP6 != "" {
	v6ServerCount++
	atype = "V6"
	serverIP = avimodels.IPAddr{Type: &atype, Addr: &nodeIP6}
} else {
	continue
}
server := AviPoolMetaServer{Ip: serverIP}
poolMeta = append(poolMeta, server)
```

Proposed fix:
```golang
if v4enabled && v4Family && nodeIP != "" {
	v4ServerCount++
	atype = "V4"
	serverIPV4 = avimodels.IPAddr{Type: &atypeV4, Addr: &nodeIP}
	serverV4 := AviPoolMetaServer{Ip: serverIP}
	poolMeta = append(poolMeta, serverV4)
}
if v6enabled && v6Family && nodeIP6 != "" {
	v6ServerCount++
	atypeV6 = "V6"
	serverIPV6 = avimodels.IPAddr{Type: &atypeV6, Addr: &nodeIP6}
	serverV6 := AviPoolMetaServer{Ip: serverIPV6}
	poolMeta = append(poolMeta, serverV6)
}
```

Symptoms is that the error: "expected IPv6 servers but found none" will always display even in the cases where there is a valid IPv4 and IPv6 address is present on the node.

### Reproduction steps

1. Run AKO(1.12 or later) in a dual-stack cluster with NodePort mode and IP_FAMILY = V4_V6
2. Deploy any dual-stack LoadBalancer service.
3. IPv4 and IPv6 addresses get allocated to the VIP but only IPv4 addresses get added to the pool in AVI.

### Expected behavior

The expected behavior is that both vip and backend servers are populated with IPv4 and IPv6 addresses from InternalIP of the node.

### Additional context

_No response_

Qwiko:
My attempt to resolve #1535.
Would like some input from the maintainers if this approach is a good one.
In PopulateServersForNPL there is some duplicate code now that could be tweaked further.
Also: I have not looked over PopulateServers in this PR but would like some input if this also needs to be tweaked to fully support dual-stack support.

### DIFF
--- a/internal/nodes/avi_model_l4_translator.go
+++ b/internal/nodes/avi_model_l4_translator.go
@@ -345,32 +345,48 @@ func PopulateServersForNPL(poolNode *AviPoolNode, ns string, serviceName string,
 		}
 		annotations = obj.([]lib.NPLAnnotation)
 		for _, a := range annotations {
-			var atype string
+
 			if v4enabled && v4Family && utils.IsV4(a.NodeIP) {
 				v4ServerCount++
-				atype = "V4"
-			} else if v6enabled && v6Family && k8net.IsIPv6String(a.NodeIP) {
-				v6ServerCount++
-				atype = "V6"
-			} else {
-				continue
-			}
-			if (poolNode.TargetPort.Type == intstr.Int && a.PodPort == poolNode.TargetPort.IntValue()) ||
-				a.PodPort == int(targetPort) {
-				enabled, ok := conditionMap[pod.Name]
-				if !ok {
-					// not disabling just because the pod was not found in condition map. i.e. ignoring false negatives
-					utils.AviLog.Debugf("key: %s, msg: pod %s not found in condition map. enabling by default", key, pod.Name)
-					enabled = proto.Bool(true)
+				atype := "V4"
+				if (poolNode.TargetPort.Type == intstr.Int && a.PodPort == poolNode.TargetPort.IntValue()) ||
+					a.PodPort == int(targetPort) {
+					enabled, ok := conditionMap[pod.Name]
+					if !ok {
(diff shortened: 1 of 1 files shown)


Dhivyaaj:

PR #1551, opened 2024-10-08
AV-219313: Fix ako with non admin tenant fails to add static routes to global vrfcontext of the admin cloud

akshayhavile:
VRF can be tenant specific or in admin tenant. So this PR takes care of handling both the scenarios.

### DIFF
--- a/internal/rest/rest_operation.go
+++ b/internal/rest/rest_operation.go
@@ -317,6 +317,24 @@ func (l *leader) AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp, 
 				continue
 			} else if op.Model == "VrfContext" && aviErr.HttpStatusCode == 412 {
 				utils.AviLog.Debugf("key: %s, msg: Error in rest operation for VrfContext Put request.", key)
+			} else if op.Model == "VrfContext" && aviErr.HttpStatusCode == 400 && strings.Contains(*aviErr.Message, lib.VrfContextNoPermission) {
+				// retry operation  by switching to admin tenant
+				err = nil
+				SetAdminTenant := session.SetTenant(lib.GetAdminTenant())
+				SetTenant := session.SetTenant(lib.GetTenant())
+				utils.AviLog.Warnf("key:%s, msg: Switching to Admin tenant from %s for %s method", key, lib.GetTenant(), op.Method)
+				SetAdminTenant(c.AviSession)
+				op.Err = c.AviSession.Put(utils.GetUriEncoded(op.Path), op.Obj, &op.Response)
+				if op.Err == nil {
+					utils.AviLog.Infof("Switching back to %s tenant", lib.GetTenant())
+					SetTenant(c.AviSession)
+					continue
+				}
+				utils.AviLog.Warnf("key: %s, msg: RestOp method %v path %v tenant %v Obj %s returned err %s with response %s",
+					key, op.Method, op.Path, lib.GetAdminTenant(), utils.Stringify(op.Obj), utils.Stringify(op.Err), utils.Stringify(op.Response))
+				err = &utils.WebSyncError{Err: op.Err, Operation: string(op.Method)}
+				SetTenant(c.AviSession)
+
 			} else if !isErrorRetryable(aviErr.HttpStatusCode, *aviErr.Message) {
 				if op.Method != utils.RestPost {
 					continue
--- a/internal/lib/constants.go
+++ b/internal/lib/constants.go
@@ -202,6 +202,7 @@ const (
 	FullSyncInterval                           = 300
 	Namespace                                  = "Namespace"
 	VrfContextNotFoundError                    = "VrfContext not found"
+	VrfContextNoPermission                     = "Cannot modify VrfContext"
 	HTTPMethodGet                              = "GET"
 	HTTPMethodPut                              = "PUT"
 	VrfContextObjectNotFoundError              = "VrfContext object not found"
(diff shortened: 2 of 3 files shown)


internal/rest/rest_operation.go
```diff
@@ -317,6 +317,24 @@ func (l *leader) AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp,
...
+				op.Err = c.AviSession.Put(utils.GetUriEncoded(op.Path), op.Obj, &op.Response)
+				if op.Err == nil {
+					utils.AviLog.Infof("Switching back to %s tenant", lib.GetTenant())
+					SetTenant(c.AviSession)
+					continue
+				}
+				utils.AviLog.Warnf("key: %s, msg: RestOp method %v path %v tenant %v Obj %s returned err %s with response %s",
+					key, op.Method, op.Path, lib.GetAdminTenant(), utils.Stringify(op.Obj), utils.Stringify(op.Err), utils.Stringify(op.Response))
+				err = &utils.WebSyncError{Err: op.Err, Operation: string(op.Method)}
+				SetTenant(c.AviSession)
```
arihantg, line 336: Logic seems correct. Please check if you can use defer setTenant(c.Avisession) instead of calling 					SetTenant(c.AviSession) in two places by adding this block in a method.

akshayhavile: @arihantg: defer setTenant will be called once function exit. So at line number 330 there is continue statement, due to which  it will continue remaining rest operations with tenant set at line number 326 .

internal/lib/lib.go
```diff
@@ -995,10 +995,6 @@ func FetchSEGroupWithMarkerSet(client *clients.AviClient, overrideUri ...NextPag
 		if aviError, ok := err.(session.AviError); ok && aviError.HttpStatusCode == 403 {
 			//SE in provider context no read access
 			utils.AviLog.Debugf("Switching to admin context from  %s", GetTenant())
-			SetAdminTenant := session.SetTenant(GetAdminTenant())
-			SetTenant := session.SetTenant(GetTenant())
-			SetAdminTenant(client.AviSession)
-			defer SetTenant(client.AviSession)
```
arihantg, line 1001: code to fallback client to admin tenant needs to be added here?

abhishekbsingh: this function is called when Advanced_L4 is set (wcp mode), there no tenant is defined in the config map which means the admin tenant client will be used by default

arihantg (approved):
LGTM

internal/cache/controller_obj_cache.go
```diff
@@ -3320,8 +3318,8 @@ func validateAndConfigureSeGroup(client *clients.AviClient, returnErr *error) bo
 		if aviError, ok := err.(session.AviError); ok && aviError.HttpStatusCode == 403 {
 			//SE in provider context no read access
 			utils.AviLog.Debugf("Switching to admin context from  %s", lib.GetTenant())
-			SetAdminTenant(client.AviSession)
-			defer SetTenant(client.AviSession)
+			//fallback to Admin Tenant
+			client = SharedAVIClients(lib.GetAdminTenant()).AviClient[0]
```
akshayhavile, line 3322: @abhishekbsingh: Here we should create new variable like `client := Shared..`  else original variable will be replaced.

internal/cache/controller_obj_cache.go
```diff
@@ -3375,8 +3371,7 @@ func ConfigureSeGroupLabels(client *clients.AviClient, seGroup *models.ServiceEn
 			if aviError, ok := err.(session.AviError); ok && aviError.HttpStatusCode == 400 {
 				//SE in provider context
 				utils.AviLog.Debugf("Switching to admin context from  %s", lib.GetTenant())
-				SetAdminTenant(client.AviSession)
-				defer SetTenant(client.AviSession)
+				client = SharedAVIClients(lib.GetAdminTenant()).AviClient[0]
```
akshayhavile, line 3374: same here

internal/cache/controller_obj_cache.go
akshayhavile: With this change, we are creating additional 8 client (more memory consumed) than just switching tenant for one operation without creating any new variables.

abhishekbsingh: 8 clients for admin tenant will anyway be created during PopulateCache method. Also, client replacement should be ok as the client is being passed to ConfigureSeGroupLabels as well.

akshayhavile: @abhishekbsingh If Tenant is non-admin, then populateCache will have 8 clients for that tenant. So here 8 more clients will be created.

akshayhavile: @abhishekbsingh: Also `validateAndConfigureSeGroup` is being called from `ValidateUserInput`. So with this change, client will be replaced for all subsequent function calls (instead of using actual tenant, admin tenant will be used.)

internal/cache/controller_obj_cache.go
```diff
@@ -1786,14 +1786,15 @@ func (c *AviObjCache) AviPopulateHttpPolicySetbyUUID(client *clients.AviClient,
 
 		}
 	}
+	tenant := getTenantFromTenantRef(*httppol.TenantRef)
```
ashishnayak96, line 1789: is there any default tenant value that needs to be set incase *httppol.TenantRef is empty? Are we sure that *httppol.TenantRef will always be non-empty?

internal/rest/rest_operation.go
```diff
@@ -317,6 +318,26 @@ func (l *leader) AviRestOperate(c *clients.AviClient, rest_ops []*utils.RestOp,
 				continue
 			} else if op.Model == "VrfContext" && aviErr.HttpStatusCode == 412 {
 				utils.AviLog.Debugf("key: %s, msg: Error in rest operation for VrfContext Put request.", key)
+			} else if op.Model == "VrfContext" && aviErr.HttpStatusCode == 400 && strings.Contains(*aviErr.Message, lib.VrfContextNoPermission) {
```
ashishnayak96, line 321: possible nil pointer error if aviErr.Message is nil?

ashishnayak96 (approved):
LGTM otherwise

PR #1561, opened 2024-10-23
AV-220244: Check for Namespace annotation and VS lb svc for determining the correct cloud

Rishabh-Baranwal:
Tests Done -
1)Restarted AKO in SV cluster for already working setup, cloud gets picked using old value set in namespace annotation.
2)Removed cloud from namespace annotation and restart AKO, cloud associated with VS "'cluster name'--kube-system-kube-apiserver-lb-svc" gets picked.
3)Failed/Bypassed the above 2 check by introducing some code and cloud gets picked through old method.

### DIFF
--- a/ako-infra/ingestion/avi_infra_controller.go
+++ b/ako-infra/ingestion/avi_infra_controller.go
@@ -90,43 +90,74 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
 	return fmt.Errorf("Avi Controller license is not in accepted list %s. License tier is: %s", acceptedLicensesInAvi, *response.DefaultLicenseTier)
 }
 
+func (a *AviControllerInfra) checkNSAnnotations() (error, string) {
+	cloudName := ""
+	nsName := utils.GetAKONamespace()
+	nsObj, err := a.cs.CoreV1().Namespaces().Get(context.TODO(), nsName, metav1.GetOptions{})
+	if err != nil {
+		utils.AviLog.Errorf("Failed to GET the %s namespace details due to the following error :%v", nsName, err.Error())
+		return err, cloudName
+	}
+	if nsObj.Annotations != nil {
+		cloudName = nsObj.Annotations["ako.vmware.com/wcp-cloud-name"]
+		if cloudName != "" {
+			utils.AviLog.Infof("Found cloud %s in ns annotation", cloudName)
+		}
+	}
+	return nil, cloudName
+}
+
 func (a *AviControllerInfra) checkVirtualService() (error, string) {
-	createdBy := "ako-" + lib.GetClusterID()
-	uri := "/api/virtualservice/?include_name&created_by=" + createdBy
-	result, err := lib.AviGetCollectionRaw(a.AviRestClient, uri)
+	cloudName := ""
+	tenants := make(map[string]struct{})
+	err := lib.GetAllTenants(a.AviRestClient, tenants)
 	if err != nil {
-		utils.AviLog.Errorf("Get uri %v returned err %v", uri, err)
-		return err, ""
+		utils.AviLog.Errorf("Error during GET tenants %v", err)
+		return err, cloudName
 	}
-	elems := make([]json.RawMessage, result.Count)
-	err = json.Unmarshal(result.Results, &elems)
-	if err != nil {
-		utils.AviLog.Errorf("Failed to unmarshal data, err: %v", err)
-		return err, ""
-	}
-	expName := lib.GetClusterID() + "--kube-system-kube-apiserver-lb-svc"
-	for i := 0; i < len(elems); i++ {
-		vs := models.VirtualService{}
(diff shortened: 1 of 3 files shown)


gurunath-avi:

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -84,23 +84,92 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+		utils.AviLog.Errorf("Error during GET tenants %v", err)
+		return err, cloudName
+	}
+	createdBy := lib.AKOPrefix + lib.GetClusterName()
+	vsName := lib.GetClusterName() + "--kube-system-kube-apiserver-lb-svc"
+	SetAdminTenant := session.SetTenant(lib.GetAdminTenant())
+	defer SetAdminTenant(a.AviRestClient.AviSession)
+	for tenantName := range tenants {
+		SetTenant := session.SetTenant(tenantName)
+		SetTenant(a.AviRestClient.AviSession)
```
abhishekbsingh, line 118: we can list VS in all tenants by setting tenantName as *. No need for listing all the tenants and VS in each tenant.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -123,42 +192,45 @@ func (a *AviControllerInfra) DeriveCloudNameAndSEGroupTmpl(tz string) (error, st
...
+		matchCloud = cloud
+		break
+	}
+	if matchCloud == nil {
+		return errors.New("cloud not found matching transport zone " + tz), "", ""
+	}
+	if matchCloud.SeGroupTemplateRef != nil && *matchCloud.SeGroupTemplateRef != "" {
+		tokenized := strings.Split(*matchCloud.SeGroupTemplateRef, "/api/serviceenginegroup/")
+		if len(tokenized) == 2 {
+			return nil, *matchCloud.Name, tokenized[1]
```
abhishekbsingh, line 204: Do we need to bring this logic out of the loop? don't think we need matchCloud flag either.

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -268,6 +340,10 @@ func fetchSEGroup(client *clients.AviClient, overrideUri ...lib.NextPage) (error
 
 func fetchVcenterServer(client *clients.AviClient) (string, error) {
 	uri := "/api/vcenterserver"
+	if utils.CloudName != "" {
```
abhishekbsingh, line 343: don't need this check as utils.CloudName will never be empty

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -84,23 +84,92 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
 
+func (a *AviControllerInfra) checkNSAnnotations() (error, string) {
+	cloudName := ""
+	nsName := utils.GetAKONamespace()
+	nsObj, err := a.cs.CoreV1().Namespaces().Get(context.TODO(), nsName, metav1.GetOptions{})
+	if err != nil {
+		utils.AviLog.Errorf("Failed to GET the %s namespace details due to the following error :%v", nsName, err.Error())
+		return err, cloudName
+	}
+	if nsObj.Annotations != nil {
```
abhishekbsingh, line 95: nit: no need for a nil check. Also, let's create a variable for `ako.vmware.com/wcp-cloud-name` string

ako-infra/ingestion/avi_infra_controller.go
```diff
@@ -84,23 +84,92 @@ func (a *AviControllerInfra) VerifyAviControllerLicense() error {
...
+
+func (a *AviControllerInfra) checkVirtualService() (error, string) {
+	cloudName := ""
+	tenants := make(map[string]struct{})
+	err := lib.GetAllTenants(a.AviRestClient, tenants)
+	if err != nil {
+		utils.AviLog.Errorf("Error during GET tenants %v", err)
+		return err, cloudName
+	}
+	createdBy := lib.AKOPrefix + lib.GetClusterName()
```
abhishekbsingh, line 112: nit: would be better if we set AKO user during boot up, and then later reuse it as and when required.

ako-infra/ingestion/avi_infra_controller.go
Rishabh-Baranwal: For code readability, I moved all the other things out of the loop. The logic is same, just the loop gets to exit or break early.

ako-infra/ingestion/avi_infra_controller.go
Rishabh-Baranwal: I had seen somewhere probably in main, where we initialise AKO user but this flow hits first so AKO user is not set at this time. I will let that part of code to set AKO user, not changing here.

abhishekbsingh (approved):

PR #1565, opened 2024-11-04
AKO doesn't support Route object with subdomain specified

DixitAakash:


ashishnayak96:
build ako

DixitAakash:
build ako

helm/ako/values.yaml
```diff
@@ -93,6 +93,7 @@ L7Settings:
   passthroughShardSize: "SMALL" # Control the passthrough virtualservice numbers using this ENUM. ENUMs: LARGE, MEDIUM, SMALL
   enableMCI: "false" # Enabling this flag would tell AKO to start processing multi-cluster ingress objects.
   fqdnReusePolicy: "InterNamespaceAllowed" # Use this to control whether AKO allows cross-namespace usage of FQDNs. enum Strict|InterNamespaceAllowed
+  defaultRouteDomain: "" # If spec.subdomain field is specified instead of spec.host field for an OpenShift route, then defaultRouteDomain is appended to the subdomain to form the FQDN.
```
akshayhavile, line 96: It will be good if we provide an example of subdomain to clarify the format. Example, AKO expect `avi.internal` as subdomain and not `.avi.internal`.

akshayhavile: Is it possible to use `L4Settings.defaultDomain` field? It serves the same purpose. We can move this field to `Networksetting` or another section of values.yaml so that it will represent both context L4 and `L7 in openshift`?

helm/ako/templates/statefulset.yaml
```diff
@@ -293,6 +293,11 @@ spec:
               configMapKeyRef:
                 name: avi-k8s-config
                 key: fqdnReusePolicy
+          - name: DEFAULT_ROUTE_DOMAIN
```
akshayhavile, line 296: This has to be added in Gateway container also.

DixitAakash:
Closing this PR in favour of https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/1575 due to major changes.

PR #1584, opened 2024-11-21
Overwrite incorrect segment in cloud with Avi SE segment And Handle Default NSX project in VPC

abhishekbsingh:


ako-infra/avirest/netinfo_handler.go
```diff
@@ -113,10 +113,10 @@ func (t *T1LRNetworking) SyncLSLRNetwork() {
 		return
 	}
 
-	if len(lslrmap) > 0 {
-		dataNetworkTier1Lrs, updateRequired := t.removeStaleLRLSEntries(client, cloudModel, lslrmap)
-		if !t.matchSegmentInCloud(dataNetworkTier1Lrs, lslrmap) || updateRequired {
-			cloudModel.NsxtConfiguration.DataNetworkConfig.Tier1SegmentConfig.Manual.Tier1Lrs = t.constructLsLrInCloud(dataNetworkTier1Lrs, lslrmap)
+	if len(lrlsMap) > 0 {
+		dataNetworkTier1Lrs, updateRequired := t.removeStaleLRLSEntries(client, cloudModel, lrlsMap)
+		if !t.matchSegmentInCloud(dataNetworkTier1Lrs, lrlsMap) || updateRequired {
```
akshayhavile, line 118: @abhishekbsingh as `updateRequired` already evaluated, can we make it as first condition in `if condition` so that evaluation is faster?

akshayhavile (commented):
@abhishekbsingh: One minor comment. Do we have any UTs or FTs to cover this use case?

abhishekbsingh:
> @abhishekbsingh: One minor comment. Do we have any UTs or FTs to cover this use case?

We do not have cloud side UTs, but we are planning to add them. These cases would be covered as part of them.

ako-infra/avirest/netinfo_handler.go
abhishekbsingh: done

akshayhavile (approved):

PR #1585, opened 2024-11-21
AV-222796: Avoid panic on service type of ExternalName

ashishnayak96:


ashishnayak96:
build ako

internal/nodes/avi_model_l4_translator.go
```diff
@@ -519,7 +519,7 @@ func PopulateServers(poolNode *AviPoolNode, ns string, serviceName string, ingre
 	if len(svcObj.Spec.IPFamilies) == 2 {
 		v4Family = true
 		v6Family = true
-	} else if svcObj.Spec.IPFamilies[0] == "IPv6" {
+	} else if len(svcObj.Spec.IPFamilies) > 0 && svcObj.Spec.IPFamilies[0] == "IPv6" {
```
akshayhavile, line 522: During ingestion of service, we check type of service and log a message saying `ExternalName service is not supported` and return from there, will help in avoid going through all remaining logic.

ashishnayak96: fixed

ashishnayak96:
build ako

ashishnayak96:
build ako

ashishnayak96:
build ako

ashishnayak96:
build ako

DixitAakash (approved):

internal/k8s/controller.go
```diff
@@ -915,6 +915,11 @@ func (c *AviController) SetupEventHandlers(k8sinfo K8sinformers) {
 			namespace, _, _ := cache.SplitMetaNamespaceKey(utils.ObjKey(svc))
 			isSvcLb := isServiceLBType(svc)
 			var key string
+			if !lib.ValidServiceType(svc) {
```
akshayhavile, line 918: @ashishnayak96 We can move this logic above line number 916 (isServiceLBType) so that extra processing will be avoided else we can change if condition as `!isSvcLb && !lib.ValidServiceType`

ashishnayak96:
> @ashishnayak96: changes looks good with one minor comment. Also is it possible to raise event against ingress to notify user that servicetype is not supported? This will help user in debugging the issue from k8s without looking into AKO logs.

The event handler is of service so it has no context of ingress.

internal/k8s/controller.go
ashishnayak96: Will take in next iteration

PR #1630, opened 2025-01-23
AV-22790: Delete VS on same fqdn different namespace in Dedicated mode

ashishnayak96:


ashishnayak96:
build ako

ashishnayak96:
build ako

DixitAakash (approved):
LGTM

internal/nodes/validator.go
```diff
@@ -213,6 +213,9 @@ func (v *Validator) ParseHostPathForIngress(ns string, ingName string, ingSpec n
 
 	var tlsConfigs []TlsSettings
 	for _, rule := range ingSpec.Rules {
+		if !EnqueueIng(key, ns, rule.Host, ingName) {
```
akshayhavile, line 216: why are we introducing this logic here?

internal/nodes/hostname_cache.go
```diff
@@ -191,18 +191,22 @@ func EnqueueIng(key, namespace, hostName, ingName string) bool {
...
+
+func hostnameExistInDifferentNamespace(key, hostName, namespace string) bool {
 	found, ingressHostMap := SharedHostNameLister().Get(hostName)
 	if found {
 		ingresses := ingressHostMap.GetIngressesForHostName()
-		if len(ingresses) > 0 {
-			nsIngressname := strings.Split(ingresses[0], "/")
-			if len(nsIngressname) > 0 {
-				if nsIngressname[0] != namespace {
+		for _, ingress := range ingresses {
```
akshayhavile, line 201: why are we introducing loop here? for all ingresses present in `ingressHostMap.GetIngressesForHostName()` should be from same namespace. So comparing first entry should serve our purpose.

internal/nodes/validator.go
ashishnayak96: we were adding the new host(already claimed) in the model here itself which shouldnt be added. So apparently the ingress status was being set to nil but during the rest layer it was again being set with IP of already claimed host.

internal/nodes/hostname_cache.go
ashishnayak96: The ingresshostmap contains ingresses from other namespace as well

internal/nodes/validator.go
akshayhavile: Is this happening in all mode `shared` and `dedicated`?

akshayhavile: Same logic has to be introduced in Route path also.

ashishnayak96: Removed the logic as deleting the VS was fixing this as well

ashishnayak96:
build ako

ashishnayak96:
build ako

akshayhavile (approved):

akshayhavile (approved):

DixitAakash (approved):

PR #1646, opened 2025-02-17
Update in ClusterName and fixed cleanup script

abhishekbsingh:


internal/lib/lib.go
```diff
@@ -1079,7 +1079,8 @@ func GetClusterIDSplit() string {
 	if clusterID != "" {
 		clusterName := strings.Split(clusterID, ":")
 		if len(clusterName) > 1 {
-			return clusterName[0]
+			// Include first 5 characters to add more uniqueness to cluster name
+			return clusterName[0] + "-" + clusterName[1][:5]
```
akshayhavile, line 1083: What is impact of this on upgrade scenario?

abhishekbsingh: Addressed the upgrade impact by keeping the clustername consistent for non VPC based deployments. VPC deployments would only be Greenfield.

ako-clean/cleanup.go
```diff
@@ -125,19 +132,62 @@ func (cfg *AKOCleanupConfig) Cleanup(ctx context.Context) error {
 	}
 	return nil
 }
+func getCloudInVPCMode() (string, error) {
+	uri := "/api/cloud/"
+	aviRestClientPool := avicache.SharedAVIClients(lib.GetTenant())
```
akshayhavile, line 137: Is there any restriction to have only one VPC enabled cloud?

internal/lib/lib.go
```diff
@@ -1079,6 +1079,10 @@ func GetClusterIDSplit() string {
 	if clusterID != "" {
 		clusterName := strings.Split(clusterID, ":")
 		if len(clusterName) > 1 {
+			if GetVPCMode() {
+				// Include first 5 characters to add more uniqueness to cluster name
+				return clusterName[0] + "-" + clusterName[1][:5]
```
akshayhavile, line 1084: Will this logic not impact VPC based deployments on upgrade ?

ako-clean/cleanup.go
```diff
@@ -104,8 +101,18 @@ func (cfg *AKOCleanupConfig) Cleanup(ctx context.Context) error {
 		cleanupL4PolicySets,
 		cleanupPoolGroups,
 		cleanupPools,
-		func() error { return avirest.DeleteServiceEngines() },
-		avirest.DeleteServiceEngineGroup,
+		func() error {
+			if VPCMode {
```
akshayhavile, line 105: why are we not deleting ServiceEngine or SEG?

internal/lib/lib.go
abhishekbsingh: No, VPC based deployment will start from VC 9.0 where this logic is aimed to go

ako-clean/cleanup.go
abhishekbsingh: In VPC based deployments, we have moved away from SE group per Supervisor cluster. Instead, we will be using the Default SE Group per Tenant, and this SE group can be shared multiple Supervisors mapped to the same Avi controller, so not deleting the SEs and SE Group.

akshayhavile (approved):

PR #1651, opened 2025-02-20
AV-226901 : Setting custom values for qps and burst for k8s client to speed up status update process

DixitAakash:


DixitAakash:
build ako

akshayhavile (approved):

PR #1654, opened 2025-02-21
AV-228668: Fix: tenant name with special characters

akshayhavile:
TODO:
1. UT / FT for tenant name with special characters
2.  Run existing UTs
3.  As AviController now allows special character in each object name so we need to check whether to handle that or not.

### DIFF
--- a/internal/lib/lib.go
+++ b/internal/lib/lib.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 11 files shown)


arihantg (approved):
LGTM

internal/lib/lib.go
```diff
@@ -274,6 +274,8 @@ func GetNSXTTransportZone() string {
 	return NsxTTzType
 }
 
+var nonDnsLabelRegex = regexp.MustCompile(`[^a-zA-Z0-9_-]+`)
```
abhishekbsingh, line 277: Please check and confirm if _ is allowed in FQDN or not. Don't think it's DNS compliant.

abhishekbsingh:
Nit: Instead of calling url.QueryEscape in all the files, we can also write a function to get tenant ref, and queryescape the tenant there.

internal/lib/lib.go
akshayhavile: Error message from AviController `labels must start with a letter (a-zA-Z) or a digit and end with a letter or a digit, and label must only have letters, digits and hyphen or underscore as composing characters`

akshayhavile:
build ako

akshayhavile:
build ako

PR #1660, opened 2025-02-25
AV-231510 Troubleshoot doc update for AV-230178

pkoshtavmware:
AV-231510 Troubleshoot doc update for AV-230178

### DIFF
--- a/docs/troubleshooting/troubleshooting.md
+++ b/docs/troubleshooting/troubleshooting.md
@@ -89,6 +89,24 @@ Check connectivity between AKO Pod and Avi controller.
 
   Refer the [document](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#securitycontext-v1-core) to select the best suitable configuration.
 
+#### Static routes not getting populated in vrfcontext
+
+#### Possible Reasons/Solutions
+
+  Reboot all ako pods by setting replica count to 0 and then reverting replica count back in ako statefuleset using following steps:
+
+    1. Edit `ako` sts in `avi-system` namespace using following command and set `replicas: 0` :
+    
+        kubectl edit sts ako -n avi-system
+  
+        Save and exit.
+
+    2. Edit `ako` sts in `avi-system` namespace using following command and set `replicas: replica-count-before-edit` :
+    
+        kubectl edit sts ako -n avi-system
+  
+        Save and exit.
+
 ## Log Collection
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting/troubleshooting.md
```diff
@@ -89,6 +89,22 @@ Check connectivity between AKO Pod and Avi controller.
 
   Refer the [document](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#securitycontext-v1-core) to select the best suitable configuration.
 
+#### Static routes not getting populated in vrfcontext
+
+#### Possible Reasons/Solutions
```
arihantg, line 94: we should explain the reason also with solution . "AKO might have failed to add the static routes after leader election and as a workaround please reboot both leader and follower ako pods  by setting replica count to 0 and then reverting replica count back  in ako statefuleset."

arihantg (approved):

docs/troubleshooting/troubleshooting.md
akshayhavile: Have we taken care of this comment?

docs/troubleshooting/troubleshooting.md
```diff
@@ -89,6 +89,24 @@ Check connectivity between AKO Pod and Avi controller.
 
   Refer the [document](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#securitycontext-v1-core) to select the best suitable configuration.
 
+#### Static routes not getting populated in vrfcontext
```
akshayhavile, line 92: We are mentioning this for HA use case so we should add it `When AKO is running in HA, ....` to clarify user when to use this.

pkoshtavmware: As per the discussion with Anurag in the meeting, he has asked not to specifically say that this should be done in this case. Instead, just say something like whenever static routes are not getting populated reboot all the ako pods both in HA and no HA mode

docs/troubleshooting/troubleshooting.md
pkoshtavmware: Taking care of this comment is not required since, As per the discussion with Anurag in the meeting, he has asked not to mention that this should be done in this case specifically. Instead, just say something like whenever static routes are not getting populated reboot all the ako pods both in HA and non HA mode

akshayhavile (approved):

PR #1673, opened 2025-03-08
AV-229617 : app-root not configured in httppolicyset when openshift routes do not have ports specified (#1664)

DixitAakash:


DixitAakash:
build ako

PR #1677, opened 2025-03-10
AV-228152 Check if Vs is present in cache before bulk delete of route status

arihantg:
…e status

### DIFF
--- a/internal/status/route_status.go
+++ b/internal/status/route_status.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


arihantg:
build ako

akshayhavile (approved):
@arihantg How are we testing it? Do we have FT/UT for it

arihantg:
> @arihantg How are we testing it? Do we have FT/UT for it

Tested with scale on LBP-suite by system-test. I am checking if we can simulate in our HA FT suite

PR #1688, opened 2025-03-18
AV-232211 : Stale string groups not getting cleaned on controller inspite of multiple AKO reboots

DixitAakash:
This PR fixes issue of stale StringGroup deletion. There were two causes of failure that are described below.

1) The **cloud_ref.name** field used in URI for getting all string groups while cache population was not a valid field for stringgroup and hence no stringgroups were fetched at bootup leading to failure in deletion of stale ones. The cloud_ref field was removed from the URI and the fix was tested by first creating ingress, hostrule with useRegex enabled, and other k8s resources. AKO creates the VS, stringgroup, and other avi objects. Later, AKO is scaled to zero pods and all k8s resources are deleted. AKO is scaled up to one pod again and AKO deletes all the stale objects including stringgroups.

2) There was another scenario which could arise even after fixing the URI in step 1. If, at the time of AKO bootup, the stale virtual service corresponding to stale stringgroup also existed, then the stale stringgroup deletion would be handled as part of stale virtual service deletion. However, if the stale VS does not exist or was never created due to some error, the stale stringgroup would not be deleted when AKO reboots. To fix this, stringgroup deletion is also added as part of dummy VS deletion. This was also tested in a similar fashion as step 1 only the VS was manually deleted from controller before AKO was scaled up to one pod.

### DIFF
--- a/internal/cache/controller_obj_cache.go
+++ b/internal/cache/controller_obj_cache.go
@@ -423,18 +423,34 @@ func (c *AviObjCache) DeleteUnmarked(childCollection map[string][]string) {
 		}
 	}
 
+	sgKeys := make(map[string][]NamespaceName)
+	for _, objkey := range c.StringGroupCache.AviGetAllKeys() {
+		if _, ok := allTenants[objkey.Namespace]; !ok {
+			allTenants[objkey.Namespace] = struct{}{}
+		}
+		intf, _ := c.StringGroupCache.AviCacheGet(objkey)
+		if obj, ok := intf.(*AviStringGroupCache); ok {
+			if !obj.HasReference {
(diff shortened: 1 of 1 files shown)


DixitAakash:
build ako

arihantg (approved):
LGTM

akshayhavile (approved):
@DixitAakash changes looks good to me. Did we test deleteConfig flag and checked SG getting deleted?

DixitAakash:
@akshayhavile **deleteConfig** was already tested when we first added regex and app-root functionality, and to be sure I test it again today. StringGroups are deleted when deleteConfig is set and again re-created along with other resources when unset.

PR #1689, opened 2025-03-19
AV-232413 : Encode pg and pool name in SNI mode for ingress/route host with no path defined

DixitAakash:
This PR fixes the underlying issue seen in the following bugs https://vmw-jira.broadcom.net/browse/AV-232211 and https://vmw-jira.broadcom.net/browse/AV-232413.
The primary issue is that If an ingress/route does not specify a path in spec, and the corresponding hostrule has useRegex field set to true, the PG name in httppolicyset is not being encoded for the empty path in SNI mode. This leads to httppolicyset creation failure as the actual pg and pools are re-created with encoded names, but httppolicyset tries to use the old not encoded pg name.


DixitAakash:
build ako

arihantg (approved):
Changes look good to me.

akshayhavile (commented):
One nit. OtherWise LGTM

akshayhavile (approved):

PR #1695, opened 2025-03-20
Added support for Proxy Enabled Application Profile in WCP env

Rishabh-Baranwal:
WCP Doc : https://vmw-confluence.broadcom.net/pages/viewpage.action?spaceKey=WCP&title=Log+Client+IP+-+Proxy+Protocol+-+Notes

Change: 
When AKO comes up in WCP environment, it will create proxy enabled application profile in AVI. This app profile will get used when any advL4LB service gets created with certain annotation as a hint to AKO to add it as ref in VS. This application profile will be deleted using cleanup when supervisor/wcp gets disabled.


jeyapradeen-avi:

internal/k8s/advl4_controller.go
```diff
@@ -80,8 +80,10 @@ func (c *AviController) SetupAdvL4EventHandlers(numWorkers uint32) {
 			}
 			oldObj := old.(*advl4v1alpha1pre1.Gateway)
 			gw := new.(*advl4v1alpha1pre1.Gateway)
+			oldAnnotVal := oldObj.Annotations[lib.GwProxyProtocolEnableAnnotation]
+			newAnnotVal := gw.Annotations[lib.GwProxyProtocolEnableAnnotation]
 
-			if !reflect.DeepEqual(oldObj.Spec, gw.Spec) || gw.GetDeletionTimestamp() != nil {
+			if !reflect.DeepEqual(oldObj.Spec, gw.Spec) || gw.GetDeletionTimestamp() != nil || oldAnnotVal != newAnnotVal {
```
abhishekbsingh, line 86: We do not need this check

internal/lib/utils.go
```diff
@@ -539,3 +540,65 @@ func IsInfraSettingNSScoped(infraSetting, namespace string) bool {
...
+	}
+	resp := models.ApplicationProfileAPIResponse{}
+	err, appProfs := ProxyEnabledAppProfileGet(client)
+	if err == nil {
+		if len(appProfs) > 1 {
+			utils.AviLog.Warnf("More than one proxy enabled app profile found")
+			return fmt.Errorf("More than one proxy enabled app profile found")
+		}
+		uri := fmt.Sprintf("/api/applicationprofile/%s", *appProfs[0].UUID)
+		err = AviPut(client, uri, appProfile, resp)
```
abhishekbsingh, line 593: we should compare the required fields and only make an API call when needed?

internal/lib/lib.go
```diff
@@ -689,6 +689,11 @@ func GetVrf() string {
 	return VRFContext
 }
 
+// One proxy enabled app profile per cluster
+func GetProxyEnabledApplicationProfileName() string {
+	return Encode(GetClusterName()+"-applicationprofile", ApplicationProfile)
```
abhishekbsingh, line 694: include `Proxy` in the name

ako-clean/cleanup.go
```diff
@@ -406,6 +408,21 @@ func cleanupPools() error {
 	return deleteAviResource("/api/pool", pools)
 }
 
+func cleanupAppProfiles() error {
+	tenant := lib.GetAdminTenant()
+	aviClient := avicache.SharedAVIClients(tenant).AviClient[0]
+	err, appProfs := lib.ProxyEnabledAppProfileGet(aviClient)
+	if err != nil {
+		utils.AviLog.Warnf("Application profile Get returned err %v", err)
+		return err
```
abhishekbsingh, line 417: are we checking response status code, we should not return error in case of 404

Rishabh-Baranwal: this is api query "?name=<value>" , will not return 404 but blank result list.

internal/lib/utils.go
```diff
@@ -539,3 +539,72 @@ func IsInfraSettingNSScoped(infraSetting, namespace string) bool {
...
+		}
+		if len(appProfs) == 0 {
+			uri := "/api/applicationprofile"
+			err = AviPost(client, uri, appProfile, resp)
+		}
+	}
+	if err != nil {
+		return err
+	}
+	utils.AviLog.Debugf("Proxy enabled application profile %s created/updated", name)
```
abhishekbsingh, line 608: This should be Info

abhishekbsingh (approved):

internal/k8s/ako_init.go
```diff
@@ -1226,6 +1226,17 @@ func (c *AviController) FullSyncK8s(sync bool) error {
 			}
 		}
 
+		// Proxy Enabled Application Profile GET/CREATE/UPDATE
+		aviClientPool := avicache.SharedAVIClients(lib.GetAdminTenant())
+		if aviClientPool == nil || len(aviClientPool.AviClient) == 0 {
+			return fmt.Errorf("avi Rest client initialization failed")
+		}
+		err = lib.ProxyEnabledAppProfileCU(aviClientPool.AviClient[0])
```
akshayhavile, line 1234: @Rishabh-Baranwal @abhishekbsingh Should we include this AppProfile creation flow as part of infra container? We already creating SEG, network in that. So resource creation for wcp will remain in one container only.

internal/lib/utils.go
```diff
@@ -539,3 +539,72 @@ func IsInfraSettingNSScoped(infraSetting, namespace string) bool {
...
+}
+
+func ProxyEnabledAppProfileCU(client *clients.AviClient) error {
+	name := GetProxyEnabledApplicationProfileName()
+	tenant := fmt.Sprintf("/api/tenant/?name=%s", GetAdminTenant())
+	appType := AllowedL4ApplicationProfile
+	createBy := GetAKOUser()
+	ppe := true
+	tcpAppProfile := models.TCPApplicationProfile{
+		ProxyProtocolEnabled: &ppe,
```
akshayhavile, line 575: Should we use proto.Bool(true) instead of new variable?

internal/lib/utils.go
```diff
@@ -539,3 +539,72 @@ func IsInfraSettingNSScoped(infraSetting, namespace string) bool {
...
+	}
+	appProfile := models.ApplicationProfile{
+		Name:          &name,
```
akshayhavile, line 578: we can use proto. functions here to convert into pointer.

internal/k8s/ako_init.go
Rishabh-Baranwal: We need this in case of VDS as well.

akshayhavile (approved):

Rishabh-Baranwal:
Closing as the change got merged to master

PR #1716, opened 2025-04-07
AV-235192: Import alb-sdk package into AKO

akshayhavile:


akshayhavile:
build ako

akshayhavile:
build ako

arihantg (approved):

abhishekbsingh:
Do we need some FT testing here as UT won't cover the SDK changes

akshayhavile:
> Do we need some FT testing here as UT won't cover the SDK changes

@abhishekbsingh: As per my opinion, we should have some Unit test cases in alb-sdk package to simulate this error condition by injecting error. From FT perspective, we need to check if Avi Controller has any method exposed to track request from AKO, return 419 for first request and then validate functionality.

abhishekbsingh (approved):

PR #1754, opened 2025-04-28
AV-237335 retry for vsvip 412 failure

arihantg:


arihantg:
build ako

internal/rest/dequeue_nodes.go
```diff
@@ -288,6 +288,12 @@ func (rest *RestOperations) CheckAndPublishForRetry(err error, publishKey avicac
 					rest.PublishKeyToSlowRetryLayer(publishKey, key)
 					return true
 				}
+			case 412:
```
akshayhavile, line 291: should we make it more generic, for all objects?

arihantg: Don't think it is needed for other objects. From rest layer 412 is seen only for VSVIP (Had done a search on Concurrent update in FT logs)

arihantg:
build ako

internal/rest/dequeue_nodes.go
```diff
@@ -288,6 +288,12 @@ func (rest *RestOperations) CheckAndPublishForRetry(err error, publishKey avicac
...
+				if strings.Contains(*aviError.Message, lib.NeedToReloadObjectDataVsVip) {
```
ashishnayak96, line 292: NIT: the check may not be required and we can retry on any 412 failure

ashishnayak96 (approved):
LGTM

PR #1770, opened 2025-05-09
Use chart appVersion instead of tag in image

ashishnayak96:


akshayhavile (commented):
@ashishnayak96 changes looks good to me. One comment... we should replace public facing url with internal url for repo .. like avi-buildops-docker-registry-02-lv.avilb.broadcom.net:8080/avi-buildops/ako/master/ako-crd-operator. so that it will be easier to test internally. Existing mentioned url will be used for public facing purpose.

PR #1783, opened 2025-05-21
AV-239699: Avoid reconcile on restart

ashishnayak96:


ako-crd-operator/internal/cache/cache.go
```diff
@@ -0,0 +1,86 @@
...
+			if err != nil {
+				return err
+			}
+			if err := json.Unmarshal(result.Results, &dataList); err != nil {
+				return err
+			}
+			for _, data := range dataList {
+				UUID, ok := data["uuid"].(string)
+				if !ok {
+					utils.AviLog.Warnf("unable to find uuid in object :[%v]", data)
```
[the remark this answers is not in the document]
ashishnayak96: added

ako-crd-operator/internal/controller/healthmonitor_controller.go
```diff
@@ -137,19 +145,38 @@ func (r *HealthMonitorReconciler) ReconcileIfRequired(ctx context.Context, hm *a
...
+		}
 		resp := map[string]interface{}{}
 		if err := r.AviClient.AviSession.Put(utils.GetUriEncoded(fmt.Sprintf("%s/%s", constants.HealthMonitorURL, hm.Status.UUID)), hmReq, &resp); err != nil {
 			utils.AviLog.Errorf("error updating healthmonitor [%s/%s]: %s", hmReq.namespace, hmReq.Name, err.Error())
 			return err
 		}
 		utils.AviLog.Infof("succesfully updated healthmonitor:[%s/%s]", hmReq.namespace, hmReq.Name)
 	}
+
+	hm.Status.LastUpdated = &metav1.Time{Time: time.Now()}
```
arihantg, line 173: Should we use the Avi response to calculate time here?

ako-crd-operator/internal/cache/cache.go
```diff
@@ -0,0 +1,86 @@
...
+				}
+				c.dataStore.Store(UUID, data)
```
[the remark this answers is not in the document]
ashishnayak96: This will be taken in separate ticket when we handle objects based on clustername. https://vmw-jira.broadcom.net/browse/AV-240348

ako-crd-operator/internal/controller/healthmonitor_controller.go
ashishnayak96: This can be done but will require extra processing of extracting last_updated from resp and converting it into human readable format everytime.
cc: @akshayhavile

akshayhavile: Good point. 
1. In kubernetes, timestamps are stored in UTC format. So we can have that instead of local.
2. As object is being updated from K8s side, we should use lastUpdated from K8s side side, instead of using from AviController side as source of truth for us is K8s side object and lastUpdated on AviController might be due to customer out of bound operation.

arihantg (approved):
LGTM

PR #1785, opened 2025-05-22
AV-240009 : Implementation to support externalTrafficPolicy as Local for NodePort services in AKO

DixitAakash:


DixitAakash:
build ako

DixitAakash:
build ako

DixitAakash:
build ako

internal/nodes/avi_model_l4_translator.go
```diff
@@ -451,14 +451,113 @@ func PopulateServersForNodePort(poolNode *AviPoolNode, ns string, serviceName st
 				}
 
 			}
+			node4Found := true
+			node6Found := true
```
akshayhavile, line 455: One Generic comment: Can we make this logic as a separate function? Reason: It will reduce overall length of parent function and easier to go through it without scrolling up and down.

internal/nodes/avi_model_l4_translator.go
```diff
@@ -451,14 +451,113 @@ func PopulateServersForNodePort(poolNode *AviPoolNode, ns string, serviceName st
...
+			if svcObj.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyLocal {
```
akshayhavile, line 456: Will this logic kicked in if POD goes down on specific node and we need to remove endpoint and add different one?

akshayhavile (commented):
Initial comments: Going through UTs.

DixitAakash:
build ako

ashishnayak96 (approved):
LGTM

akshayhavile (approved):

PR #1789, opened 2025-05-27
AV-239368 Add 'revokeVipRoute' support to L4Rule crd

n1lesh:
1. Adds `revokeVipRoute` field to L4Rule CRD.
2. Adds validation for usage only in NSX-T cloud. 
3. Adds Event of type `InvalidConfiguration` when `revokeVipRoute` is used in combination with `ako.vmware.com/enable-shared-vip` annotation.
4. Adds `TestSharedVIPSvcWithRevokeVipRoute` integration test case.


akshayhavile:
build ako

PR #1797, opened 2025-06-04
AV-239773: Add validation for PING and TCP HMs

ashishnayak96:
Added validations for PING and TCP HMs.
Manual tests done: 
1. Created both PING and TCP HMs
2. Updated both PING and TCP HMs.
3. Validate TCP HM creation/updation fails if tcp_half_open is set to true and other tcp fields are set.

Also fixed an issue where auth_type validation was failing when complete http object was missing in case of TCP or PING HM

### DIFF
--- a/ako-crd-operator/docs/samples/healthmonitor/hm-ping.yaml
+++ b/ako-crd-operator/docs/samples/healthmonitor/hm-ping.yaml
@@ -1,0 +1,13 @@
+apiVersion: ako.vmware.com/v1alpha1
+kind: HealthMonitor
+metadata:
+  name: example-ping-healthmonitor
+  namespace: default
+spec:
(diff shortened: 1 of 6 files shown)


ako-crd-operator/docs/samples/healthmonitor/hm-ping.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: example-ping-healthmonitor
+  namespace: default
+spec:
+  type: HEALTH_MONITOR_PING
+  failed_checks: 3
+  successful_checks: 2
+  receive_timeout: 5
+  send_interval: 10
+  monitor_port: 80
+  is_federated: false
```
pkoshtavmware, line 13: Is this sample complete ?

PR #1804, opened 2025-06-10
AV-240348: Make objects clustername aware

ashishnayak96:


ako-crd-operator/internal/controller/healthmonitor_controller_test.go
```diff
@@ -81,4 +83,67 @@ var _ = Describe("HealthMonitor Controller", func() {
...
+
+		It("should handle empty namespace", func() {
+			// Set environment variable for cluster name
+			originalClusterName := os.Getenv("CLUSTER_NAME")
+			defer os.Setenv("CLUSTER_NAME", originalClusterName)
+			os.Setenv("CLUSTER_NAME", "test-cluster")
+
+			markers := ctrlutils.CreateMarkers("", "")
+
+			// Should only have cluster name marker since namespace is empty
```
pkoshtavmware, line 143: Is this test working since CLUSTER_NAME is not set here ?

pkoshtavmware (changes requested):

ako-crd-operator/internal/controller/healthmonitor_controller_test.go
ashishnayak96: I will remove the unit tests and take it separately

PR #1816, opened 2025-06-20
AV-239701: Add secret to pass authentication details to health monitor

ashishnayak96:
Adding a field DependencySum to manage the changes with external dependencies. Changing a secret may result in no op since the generation of the healthmonitor does not change. So we will be using combination of generation and DependencySum to evaluate if it is a no op.
We are also adding index on hm and predicate on secret to optimize the number of objects we are watching.


ako-crd-operator/internal/constants/constants.go
```diff
@@ -6,4 +6,5 @@ const (
 	Sensitive                   = "<sensitive>"
 	ApplicationProfileFinalizer = "applicationprofile.ako.vmware.com/finalizer"
 	ApplicationProfileURL       = "/api/applicationprofile"
+	HealthMonitorSecretType     = "ako.vmware.com/healthmonitor"
```
arihantg, line 9: this type needs to be generic which can be used by various akocrds .  something like "ako.vmware.com/tls" or "ako.vmware.com/basic-auth"

ashishnayak96: Done

akshayhavile (commented):
Rest of the changes looks good to me.

ako-crd-operator/api/v1alpha1/healthmonitor_types.go
```diff
@@ -102,14 +102,8 @@ type HealthMonitorSpec struct {
...
-	// Username for server authentication.
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=128
-	Username string `json:"username"`
-	// Password for server authentication.
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=128
-	Password string `json:"password"`
+	// SecretRef is the reference to the secret containing the username and password.
+	SecretRef string `json:"secret_ref,omitempty"`
```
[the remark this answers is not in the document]
ashishnayak96: secretRef might not be required if authentication is not set in health monitor. I have separately added a validation that if authentication is set then secretRef is required

PR #1838, opened 2025-07-11
AV-244081 Add known issue disclaimer about gateway api

n1lesh:


docs/vpc_mode.md
```diff
@@ -32,4 +32,6 @@ NetworkSettings:
 ```yaml
 NetworkSettings:
   vipNetworkList: []
-```
\ No newline at end of file
+```
+
+**KNOWN ISSUE**: The Gateway API doesn't support VPC mode. Gateway API will require both IPAM and `vipNetworkList` configured for VIP allocation, even when AKO is operating with `vpcMode` enabled.
```
akshayhavile, line 37: Can you please add it to CHANGELOG?

PR #1879, opened 2025-08-25
AV-247848: Add TrafficEnabled flag for ParentVS in GatewayAPI and L4 VS

akshayhavile:
This is for VCD to VCF migration use case.

VCF handling of L4 annotation will be taken separately.

### DIFF
--- a/ako-gateway-api/nodes/gateway_model.go
+++ b/ako-gateway-api/nodes/gateway_model.go
@@ -60,10 +60,9 @@ func (o *AviObjectGraph) BuildGatewayParent(gateway *gatewayv1.Gateway, key stri
 		oldTenant = lib.GetTenant()
 	}
 	// Default: Traffic is enabled
-	// Not using it in Shared VIP as we don't have use case for it.
 	trafficEnabled := proto.Bool(true)
 
-	if traffic_enabled, ok := gateway.Annotations[lib.VSTrafficEnabled]; ok && strings.ToLower(traffic_enabled) == "false" {
+	if traffic_disabled, ok := gateway.Annotations[lib.VSTrafficDisabled]; ok && strings.ToLower(traffic_disabled) == "true" {
 		trafficEnabled = proto.Bool(false)
 	}
 	// Cleanup resources in the old tenant in case of tenant change
--- a/ako-gateway-api/nodes/gateway_model.go
+++ b/ako-gateway-api/nodes/gateway_model.go
@@ -16,7 +16,9 @@ package nodes
 
 import (
 	"context"
+	"strings"
 
+	"google.golang.org/protobuf/proto"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/net"
@@ -57,6 +59,13 @@ func (o *AviObjectGraph) BuildGatewayParent(gateway *gatewayv1.Gateway, key stri
 	if oldTenant == "" {
 		oldTenant = lib.GetTenant()
 	}
+	// Default: Traffic is enabled
+	// Not using it in Shared VIP as we don't have use case for it.
+	trafficEnabled := proto.Bool(true)
+
+	if traffic_enabled, ok := gateway.Annotations[lib.VSTrafficEnabled]; ok && strings.ToLower(traffic_enabled) == "false" {
+		trafficEnabled = proto.Bool(false)
+	}
 	// Cleanup resources in the old tenant in case of tenant change
 	if tenant != oldTenant {
 		oldModelName := lib.GetModelName(oldTenant, vsName)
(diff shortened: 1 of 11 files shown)


internal/lib/constants.go
```diff
@@ -256,6 +256,7 @@ const (
 	AviIngressController             = "ako.vmware.com/avi-lb"
 	AKOConditionType                 = "ako.vmware.com/ObjectDeletionInProgress"
 	DefaultSecretEnabled             = "ako.vmware.com/enable-tls"
+	VSTrafficEnabled                 = "ako.vmware.com/enable-traffic"
```
abhishekbsingh, line 259: I think we should call it "ako.vmware.com/disable-traffic" and set that to True to disable traffic

ako-gateway-api/nodes/gateway_model.go
```diff
@@ -57,6 +59,13 @@ func (o *AviObjectGraph) BuildGatewayParent(gateway *gatewayv1.Gateway, key stri
 	if oldTenant == "" {
 		oldTenant = lib.GetTenant()
 	}
+	// Default: Traffic is enabled
+	// Not using it in Shared VIP as we don't have use case for it.
+	trafficEnabled := proto.Bool(true)
+
+	if traffic_enabled, ok := gateway.Annotations[lib.VSTrafficEnabled]; ok && strings.ToLower(traffic_enabled) == "false" {
```
abhishekbsingh, line 66: Is there any specific reason why we are not using AviInfraSetting to define this behaviour?

ako-gateway-api/nodes/gateway_model.go
```diff
@@ -57,6 +59,13 @@ func (o *AviObjectGraph) BuildGatewayParent(gateway *gatewayv1.Gateway, key stri
 	if oldTenant == "" {
 		oldTenant = lib.GetTenant()
 	}
+	// Default: Traffic is enabled
+	// Not using it in Shared VIP as we don't have use case for it.
```
abhishekbsingh, line 63: misplaced comment?

internal/nodes/avi_model_l4_translator.go
```diff
@@ -60,7 +60,13 @@ func (o *AviObjectGraph) ConstructAviL4VsNode(svcObj *corev1.Service, key string
 			fqdns = append(fqdns, fqdn)
 		}
 	}
+	// Default: Traffic is enabled
+	// Not using it in Shared VIP as we don't have use case for it.
+	trafficEnabled := proto.Bool(true)
 
+	if traffic_enabled, ok := svcObj.Annotations[lib.VSTrafficEnabled]; ok && strings.ToLower(traffic_enabled) == "false" {
```
abhishekbsingh, line 67: This logic should also go to avi_model_advl4_translator.go file

ako-gateway-api/nodes/gateway_model.go
akshayhavile: This flag is most relevant in VCD to VCF migration use case. In this use case , L7 will be handled by Gateway. L4 will be handled by LB type service.
AviInfrasetting has broader object mapping which include ingresses also. 
So we don't want to expose settings to the users which indicates this flag will be applicable only for GW and LB K8s side objects.
Also the way our AviInfrasetting CRD is designed, it is used to provide properties to L4 and L7 VS based on underlying infrastructure components like vip network, SEG, T1LR etc.
This indicates individual VS specific properties which should have gone to VS specific CRD. 
But we don't have CRD for GW.
So we planned for Annotation.

internal/nodes/avi_model_l4_translator.go
abhishekbsingh: I think there's a little confusion here, we need this logic to be present in advl4_translator as well for this to work in VCF deployments.

internal/nodes/avi_model_l4_translator.go
```diff
@@ -60,7 +60,13 @@ func (o *AviObjectGraph) ConstructAviL4VsNode(svcObj *corev1.Service, key string
 			fqdns = append(fqdns, fqdn)
 		}
 	}
+	// Default: Traffic is enabled
+	// Not using it in Shared VIP as we don't have use case for it.
+	trafficEnabled := proto.Bool(true)
 
+	if traffic_disabled, ok := svcObj.Annotations[lib.VSTrafficDisabled]; ok && strings.ToLower(traffic_disabled) == "true" {
```
akshayhavile, line 67: Not required as Use case VCF and in VCF, it is always frontended by Gateway so workflow is different.

akshayhavile:
build ako

abhishekbsingh (approved):
lgtm

akshayhavile:
build ako

akshayhavile:
build ako

akshayhavile:
build ako

arihantg (approved):
LGTM

PR #1891, opened 2025-09-03
AV-248349: Implement backend extension ref for dedicated mode

ashishnayak96:
This PR also fixes: https://vmw-jira.broadcom.net/browse/AV-249266


ashishnayak96:
build ako

PR #1901, opened 2025-09-09
AV-250125 check tenant in crd status of HM and appProfile

arihantg:


arihantg:
build ako

ashishnayak96 (approved):
LGTM

ako-gateway-api/lib/dynamic_client.go
```diff
@@ -421,6 +421,17 @@ func IsApplicationProfileProcessed(obj interface{}, namespace, name string) bool
 		return false
 	}
 
+	tenant, ok := statusJSON["tenant"]
+	if !ok || tenant == "" {
+		utils.AviLog.Warnf("key:%s/%s, msg: ApplicationProfile CRD tenant not found", namespace, name)
+		return false
+	}
+	namespaceTenant := lib.GetTenantInNamespace(namespace)
+	if tenant != namespaceTenant {
```
akshayhavile, line 430: @arihantg Will this create a deadlock?
1. In AKO crd operator, we will try to delete the object due to tenant change. But Object will not be deleted as it will be referred by VirtualService.
2. We will return false from here. So event handler will not proceed and Ref will not be removed from VS.

arihantg: It should not create a deadlock since for 2nd point vs will be removed from old tenant of avi . It won't wait for hm or appprofile ref to be removed. 
issue was in vs/pool creation in new tenant since it might fail if it tries to create with refs which are not existing in new tenant.
Ft results on tenancy with this PR https://github-vcf.devops.broadcom.net/ANS/avi-test/pull/5633
http://10.80.34.230:8080/mb3_verbose.html

PR #1910, opened 2025-09-18
AV-249574: Updating FQDN for Ingress and HostRule getting rejected

akshayhavile:


tests/dedicatedvstests/l7_dedicated_crd_test.go
```diff
@@ -1333,3 +1333,100 @@ func TestHTTPRuleCreateDeleteEnableHTTP2(t *testing.T) {
 
 	TearDownIngressForCacheSyncCheck(t, secretName, ingName, svcName, modelName)
 }
+
+/*
+This test case tests following scenario
+1. Create an ingress with foo.com
+2. Create Hostrule with Fqdn foo.com
+3. Delete the ingress.
```
DixitAakash, line 1341: Although it should not make a difference, I think we delete the hostrule first.

DixitAakash (approved):
Overall changes look good to me.
Nit : If we could re-create the ingress as well with new hostname, in one of the test cases and check if hostrule config is applied. Although, I do understand it should be applied since hostrule is getting accepted.

arihantg (approved):
LGTM

PR #1911, opened 2025-09-19
AV-251077 : AKO CRD Opertaor installation fails when Helm version 3.2.1 is used

DixitAakash:
This PR removes certain jinja2 templates defined in the chart, that try to use fields, that are not defined in values.yaml. Using undefined fields results in nil pointer exception in older helm versions while same is ignored in newer helm versions.


DixitAakash:
build ako

arihantg (approved):
LGTM

akshayhavile (approved):

PR #1921, opened 2025-09-24
AV-251027: Do not reboot AKO if configmap present in ako installed namespace is not associated with AKO POD

akshayhavile:
…


akshayhavile:
build ako

akshayhavile:
```After Adding informer for ako specific configmap```


Screenshot 2025-09-29 at 3 05 10 PM

// No message
Screenshot 2025-09-29 at 3 21 03 PM

